PR #54, opened 2023-06-02
feat(ui): create <ToggleSwitch /> components

renanvalentin:
# Checklist

- [x] JIRA -LW-6666
- [x] Screenshots added.

---

## Proposed solution

It uses [RadixUI's Switch](https://www.radix-ui.com/docs/primitives/components/switch) as an engine to handle keyboard navigation, etc.

```
 <ToggleSwitch
          label="Label"
          id="rest"
          defaultChecked
          icon={<InfoIcon />}
        />
```

## Screenshots

capture-b0996ce8

---
# Allure report

**smokeTests**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/293/5202558128/index.html) for [e4dcb674](https://github.com/input-output-hk/lace/pull/54/commits/e4dcb6740226480ddb21c1ad667c57769481e67e)
| Total | 19     | 0      | 0       | 0     | 19    | ✅     |


packages/ui/src/design-system/toggle-switch/toggle-switch.css.ts
```diff
@@ -0,0 +1,90 @@
+import { sx, vars, style, createVar } from '../../design-tokens';
+
+const boxShadow = createVar();
```
VanessaPC, line 3: is there a possibility to have a file with common variables that we could import into components, like this one? os does it seem like too much overhead for something small?

packages/ui/src/design-system/toggle-switch/toggle-switch.stories.tsx
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
+
+import type { Meta } from '@storybook/react';
+
+import { ReactComponent as InfoIcon } from '../../assets/icons/info.component.svg';
+import { ThemeColorScheme, ThemeProvider } from '../../design-tokens';
+import { Page, Variants, Section } from '../decorators';
+import { Divider } from '../divider';
+import { Grid } from '../grid';
+import { Cell } from '../grid/cell.component';
```
VanessaPC, line 10: why from the `cell.component`? do we not want to add `index`  files?

lgobbi-atix (dismissed):

packages/ui/src/design-system/toggle-switch/toggle-switch.component.tsx
```diff
@@ -0,0 +1,52 @@
...
+  'children' | 'className'
+> & {
+  icon?: ReactNode;
+  label?: string;
+  onCheckedChange?: (checked: boolean) => void;
+  disabled?: boolean;
+  checked?: boolean;
+  defaultChecked?: boolean;
+  required?: boolean;
+  id?: string;
```
VanessaPC, line 23: if the `htmlFor` property is empty, will it not complain? maybe the id shouldn't be optional?

packages/ui/src/design-system/toggle-switch/toggle-switch.component.tsx
```diff
@@ -0,0 +1,52 @@
...
+  HTMLAttributes<HTMLButtonElement>,
+  'children' | 'className'
+> & {
+  icon?: ReactNode;
+  label?: string;
+  onCheckedChange?: (checked: boolean) => void;
+  disabled?: boolean;
+  checked?: boolean;
+  defaultChecked?: boolean;
+  required?: boolean;
```
VanessaPC, line 22: The `required` property almost seems like we should require it... and set it as off by default? I have a question, if we don't send a string on a `label`, what's the default?

packages/ui/src/design-system/toggle-switch/toggle-switch.stories.tsx
renanvalentin: Good catch!

packages/ui/src/design-system/toggle-switch/toggle-switch.css.ts
VanessaPC: thank you!

packages/ui/src/design-system/toggle-switch/toggle-switch.component.tsx
renanvalentin: The `required` will add the  `aria-required="true"` attribute. You will see the "Simple base control" variant without the label.

VanessaPC (approved):

VanessaPC:
Great work!

pczeglik-iohk (approved):

PR #66, opened 2023-06-05
fix(extension): lint warnings

lgobbi-atix:
## Proposed solution

Fixes some lint warnings


mchappell (changes requested):
In `asset-transformers.test.ts` please add <typeof import('@lace/cardano')> to remove the any, and remove the eslint disable statement for this rule

mchappell (approved):

PR #78, opened 2023-06-06
feat(ui): add PieChart component

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-6673)
- [x] Screenshots added.

---

## Proposed solution

Basic PieChart component with simple API, using Recharts.

## Screenshots

image

image

---
# Allure report

**smokeTests**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/418/5279709464/index.html) for [c8283ffd](https://github.com/input-output-hk/lace/pull/78/commits/c8283ffd7eb4f3d5046d6e74854bb10c298021eb)
| Total | 19     | 0      | 0       | 0     | 19    | ✅     |


mrcnk (approved):

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+import type { PieChartProps } from './pie-chart.component';
+
+const mockDataSet = (amount: number): { name: string; value: number }[] =>
+  Array.from({ length: amount }).map((_, index) => ({
+    name: `Pool ${index + 1}`,
+    value: 100 / amount,
+  }));
+
+const meta: Meta<typeof PieChart> = {
+  title: 'PieChart',
```
renanvalentin, line 27: Can you move it to `Elements/PieChart` so it matches Figma?

packages/ui/src/design-system/pie-chart/constants.ts
```diff
@@ -0,0 +1,18 @@
+import type { ColorValueHex } from '../../types';
```
renanvalentin, line 1: Consider renaming the file to `pie-chart.data.ts`

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+      <Page title="Pie Chart">
+        <Story />
+      </Page>
+    ),
+  ],
+};
+
+export default meta;
+
+export const Overview = (): JSX.Element => (
```
renanvalentin, line 33: I noticed that the storybook is lacking a dark color scheme. Would it be possible to discuss with the UX team about adding one?

packages/ui/src/design-system/pie-chart/index.ts
```diff
@@ -0,0 +1,6 @@
+export type { PieChartProps } from './pie-chart.component';
+export { PieChart } from './pie-chart.component';
+export {
+  PIE_CHART_DEFAULT_COLOR_SET,
```
renanvalentin, line 4: Would it be appropriate to export this constant to the public API? Perhaps it would be better to make these colors the default and proceed from there.

packages/ui/src/design-system/pie-chart/pie-chart.component.tsx
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable functional/prefer-immutable-types, @typescript-eslint/strict-boolean-expressions, @typescript-eslint/no-unnecessary-condition */
```
renanvalentin, line 1: Why were those rules disabled?

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
```diff
@@ -0,0 +1,232 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/naming-convention, functional/immutable-data */
```
renanvalentin, line 1: Avoid disabling eslint rules

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+import {
+  PIE_CHART_DEFAULT_COLOR_SET,
+  PieChartGradientColor,
+} from './constants';
+import { PieChart } from './pie-chart.component';
+
+import type { PieChartProps } from './pie-chart.component';
+
+const notInChromatic = !isChromatic();
```
renanvalentin, line 18: Consider using `process.env. CI` instead.

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+    </Section>
+  </>
+);
+
+type ConfigurableStoryProps = Pick<
+  PieChartProps<{ name: string; value: number }>,
+  'colors' | 'data' | 'direction' | 'tooltip'
+>;
+
+export const Configurable = ({
```
renanvalentin, line 180: It may be worthwhile to rename the feature to "Interactions" in order to maintain consistency with the naming convention used for interactive stories.

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: I can't see any common pattern according to Figma. For example, we use `Buttons/Main` where in Figma it's `Basic input/Buttons`. Can we document this?

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: This is intended, there is no difference in dark color scheme for a PieChart and I confirmed this with Allan and Elena.

packages/ui/src/types.ts
```diff
@@ -4,3 +4,5 @@ export type OmitClassName<T extends HTMLElement = HTMLDivElement> = Omit<
   HTMLAttributes<T>,
   'className'
 >;
+
+export type ColorValueHex = `#${string}`;
```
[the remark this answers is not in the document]
przemyslaw-wlodek: Apologies but I felt that this type is unnecessary and I recommend using [this pattern](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#wrappingmirroring-a-html-element), also I think that this structure should be flat for now, unless we have at least couple of types in this `types.ts` file and we feel the need for splitting.

packages/ui/src/design-system/pie-chart/index.ts
przemyslaw-wlodek: 1. Could you please elaborate on the first point? It's exported in the public API of this component.
2. We discussed that in the staking squad and we can get back to this offline.

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: Thank you for this suggestion but I feel that `isChromatic` is more clear and more appropriate in this case (also follows their documentation). @mrcnk has really great idea for RFC to introduce [global preference](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) in our theme provider to remove the animations. This would enable an easy way to disable animations globally, not only per component and could be easily used in Chromatic.

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: It's not really interactive as in the other examples (user interaction). Here the only thing we could present as interaction is showing a tooltip on the mouseover but we don't have a real tooltip for now.

Can we emerge a new pattern of having 3 types of stories?
- Overview (all variants - used for Chromatic snapshots)
- Configurable (with controls to play with)
- Interactions (user interactions with "play" like e2e test)

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: Let's jump on a call to discuss these. I think we need to unify our linter rules as this is way different than it was in the other projects. Some rules are great but some of them are point to discussion.

1. `functional/prefer-immutable-types`
   - wraps `PieChartProps<T>` in `Readonly` which breaks its intended API on a type level
2. `@typescript-eslint/strict-boolean-expressions`
   - fails in perfectly fine examples of code e.g. simple checks `object[key] ? 0 : 1` -> `Boolean(object[key])` has to be used
   - makes it much more verbose
   - it should probably be relaxed e.g. enabled for `number`
3. `@typescript-eslint/no-unnecessary-condition`
   - disallows some common enum checks e.g. `SomeEnum[value as SomeEnum]` (example `formatPieColor` fn in this PR)
   - it's also a warning when most of our rules should throw an error (so it may pollute our lint logs)

packages/ui/src/design-system/pie-chart/pie-chart.component.tsx
przemyslaw-wlodek: Like in the other comment:

1. `@typescript-eslint/naming-convention`
   - I can revert this one until further consensus. It came out during squad-level code review that many of us consider prefixing booleans as an anti-pattern (it's well described [here](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig#comment-j4fl))
2. `functional/immutable-data`
   - produces an error with storybook controls e.g. `Configurable.argTypes`
   `ESLint: Modifying an existing object/array is not allowed.(functional/immutable-data)`

I think we should discuss which rules we find useful in `eslint-plugin-functional` as we are not really an FP codebase and some rules may bring worse code readability and performance issues.

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
szymonmaslowski: I agree with @renanvalentin here. If the PieCart is identical for the dark theme then let's show it.

packages/ui/src/types.ts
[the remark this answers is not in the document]
szymonmaslowski: I like an approach of having a parent directory and mutiple files inside. Regarding those files names I hasitete between having the same name as exported entity (`types/omit-class-name.ts`) vs grouping files (`types/props.ts`). However I think the first approach is better even that we can end up with very small files.

packages/ui/src/design-system/pie-chart/index.ts
szymonmaslowski: @renanvalentin In staking we need to match colors of the chart to the corresponding staking info cards which display color they refer to. We thought it would be best to keep colorset with the PieChart implementation and expose it for our (and maybe not only our) use case.

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
renanvalentin: Good catch on the `Buttons/Main`! I would say it should also live in `Basic input.`

packages/ui/src/design-system/pie-chart/constants.ts
renanvalentin: My bad for not providing proper documentation on the file name structure. I'll create a PR explaining it. You can search for `*.data.ts` to locate existing files.

`
*.data.ts       |    Defines the data representation of the UI component or application logic   
`

packages/ui/src/design-system/pie-chart/pie-chart.component.tsx
renanvalentin: Let's follow the eslint contract in this PR, and feel free to open a new PR with a broader change applying the new eslint rules. +1 if the new rules can be fixable automatically.

For `item 1` it removes the friction during code review as the eslint contract takes precedence. e.g: "Please, rename this variable to `isActive` or `active`."

For `item 2` you can follow this approach:
https://github.com/input-output-hk/lace/blob/main/packages/ui/.eslintrc.js#L130
https://github.com/eslint-functional/eslint-plugin-functional/blob/main/docs/rules/immutable-data.md#ignoreaccessorpattern

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
renanvalentin: Can you add a new section to the Overview story to have controls?

przemyslaw-wlodek: As discussed, I've added a `Controls` story. 👍🏻

packages/ui/src/design-system/pie-chart/pie-chart.component.tsx
przemyslaw-wlodek: Thank you, I will create a separate discussion regarding 1st one. 👍🏻

packages/ui/src/design-system/pie-chart/index.ts
przemyslaw-wlodek: I left the export but assigned this as the default value for pie chart colors. Thanks 👍🏻

packages/ui/src/types.ts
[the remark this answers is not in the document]
przemyslaw-wlodek: Done 👍🏻

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: Done 👍🏻

packages/ui/src/design-system/pie-chart/constants.ts
przemyslaw-wlodek: Done 👍🏻

packages/ui/src/design-system/pie-chart/pie-chart.stories.tsx
przemyslaw-wlodek: Done 👍🏻

szymonmaslowski (approved):
Great work!

packages/ui/src/design-system/pie-chart/pie-chart.component.tsx
```diff
@@ -0,0 +1,124 @@
...
+  data: inputData,
+  direction = 'clockwise',
+  nameKey = 'name',
+  tooltip,
+  valueKey = 'value',
+}: PieChartProps<T>): JSX.Element => {
+  const data = inputData.slice(0, colors.length);
+
+  if (data.length < inputData.length) {
+    console.error(
```
renanvalentin, line 81: One issue with the current method is that it may cause the dev tools to be spammed as the message will appear in every re-render. Additionally, it is important to ensure that the message is only displayed in development by using a guard such as `process.env.NODE_ENV === 'development',` to prevent logs from being shipped to production.

Could you take a look at how radix-ui shows any warning?

[DescriptionWarning](https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/alert-dialog/src/AlertDialog.tsx#L241)
[DescriptionWarning Usage](https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/alert-dialog/src/AlertDialog.tsx#LL145C2-L145C2)

mrcnk: Logs won't be shipped to production as they are being muted by extension's Webpack

przemyslaw-wlodek: As discussed, it will be solved by LW-7034

renanvalentin (approved):
Great job @przemyslaw-wlodek , thanks for the effort!

PR #90, opened 2023-06-07
use sdk-wallet v0.15.0

shawnbusuttil:
# Checklist

- [ ] JIRA - \<link>
- [ ] Screenshots added.

---

## Proposed solution

Updates the ADA handle receive flow to use the handles$ observable from https://github.com/input-output-hk/cardano-js-sdk/commit/1c3b532c9b9f4fe48ba1555749b21faa27648c1a.

## Testing

## Screenshots


apps/browser-extension-wallet/src/hooks/useGetHandles.ts
```diff
@@ -0,0 +1,12 @@
...
+import { useWalletStore } from '@src/stores';
+import { isAdaHandleEnabled } from '@src/features/ada-handle/config';
+import { useObservable } from './useObservable';
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const useGetHandles = () => {
+  const { inMemoryWallet } = useWalletStore();
+  const handles = useObservable(inMemoryWallet.handles$);
+
+  return useMemo(() => isAdaHandleEnabled === 'true' && handles, [handles]);
```
VanessaPC, line 11: They way `KoraLabs` works (or will work with `allSettled` is that you will get an `[handle, ... ]` fully resolved, and you won't be looking for more handles later, because you would have gotten all the found ones at once. 
In this case we could drop the `useMemo` for something like: 
`isAdaHandleEnabled ?? handles` or `isAdaHandleEnabled ? handles :  null/false` 

Also, I probably would use `useMemo` for something more resource intensive, in this case you are just returning the value from the observable, so useMemo doesn't entirely optimize the operation of the handle, because if we receive a new value, then the new value will be sent.

shawnbusuttil: Fair point. Thanks @VanessaPC .

VanessaPC (approved):

PR #95, opened 2023-06-07
feat(ui): create <SideDrawer /> component

renanvalentin:
# Checklist

- [x] JIRA - LW-6617
- [x] Screenshots added.

---

## Proposed solution

The side drawer is rendered using [Radix-UI's Dialog](https://www.radix-ui.com/docs/primitives/components/dialog) engine.

```jsx
import { SideDrawer } from '@lace/ui'



<SideDrawer.Root>
  <SideDrawer.Overlay zIndex={100} />

  <SideDrawer.Content data-testid="side-drawer-content" zIndex={101}>
    <SideDrawer.Header
      text="Label"
      onBackClick={onBackClick}
      onCloseClick={onCloseClick}
    />

    <SideDrawer.Body>
      <SideDrawer.Headline
        title="Section title"
        description="Lorem ipsum dolor sit amet quare id faciam."
      />
    </SideDrawer.Body>

    <SideDrawer.Footer>
      <Buttons.CallToAction label="Label" />

      <SideDrawer.Close>
        <Buttons.Secondary label="Close" />
      </SideDrawer.Close>
    </SideDrawer.Footer>
  </SideDrawer.Content>

  <SideDrawer.Trigger>
    <Buttons.Primary label="Open side drawer" />
  </SideDrawer.Trigger>
</SideDrawer.Root>;


```

## Testing

This test aims to ensure that the side drawer is displayed correctly. Please refer to https://github.com/input-output-hk/lace/pull/95/files#diff-cbe4990914b060277fd8a16e4417da6ea2323ffa94c6ba42ad14451c6693db36R367

## Screenshots

capture-936f94

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/352/5244226814/index.html) for [193c66f9](https://github.com/input-output-hk/lace/pull/95/commits/193c66f99ce4ba11c903614e84aec1fc89346e8b)
| Total | 18     | 1      | 0       | 0     | 19    | ❌     |


packages/ui/src/design-system/side-drawer/side-drawer-content.css.ts
```diff
@@ -0,0 +1,73 @@
...
+  to: {
+    transform: 'translateX(105%)',
+  },
+});
+
+export const container = style([
+  sx({
+    width: '$dialog',
+    height: '$auto',
+    zIndex: '$dialog',
```
szymonmaslowski, line 25: I am an opponent of zIndexes 😛 In most cases we are able to develop components with no zIndexes usage, but instead using portals + stacking (later in the tree gets rendered above).
Usefull articles:
 - https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_without_z-index
 - https://www.codegram.com/blog/stop-ab-using-z-index/
 - https://danielrotter.at/2020/04/08/avoid-z-index-whenever-possible.html

Can we by default go with not setting the zIndex and give a possibility through the API to enable it in case it would be necessary? Probably in the extension which is already using zIndexes broadly it will be ncecessary. But we are building here something fresh which should be flexible but not influenced by any bad design which zIndexes often tend to become. Of course there are technics to manage zIndexes, but if we can avoid it (and we can in majority of cases) why son't we do that?

packages/ui/src/design-system/side-drawer/side-drawer-content-body.component.tsx
```diff
@@ -0,0 +1,17 @@
+import type { ReactNode } from 'react';
```
szymonmaslowski, line 1: Do you have a strong opion on including the `content` word in the file names? I am not convinced TBH... Maybe we could avoid this word, what do you think?

packages/ui/src/design-system/side-drawer/side-drawer-content-headline.css.ts
```diff
@@ -0,0 +1,25 @@
...
+});
+
+export const separator = style([
+  sx({
+    background: '$side_drawer_separator_bgColor',
+    width: '$fill',
+    mb: '$32',
+  }),
+  {
+    height: '1.5px',
```
szymonmaslowski, line 15: Could you make sure this exceptional value is for sure required? There is no that much difference bewtween 1.5 and 2. So maybe we could stay with 2 which is defined in the spacing.

In case of the card component it turned out all the 1.5s could be changed to 2 and we achieved a standarisation 😉

packages/ui/src/design-system/side-drawer/side-drawer-content.css.ts
```diff
@@ -0,0 +1,73 @@
...
+    "footer"`,
+    gridTemplateRows: 'auto 1fr auto',
+    overflow: 'hidden',
+  },
+]);
+
+export const storybook = style({});
+
+// storybook only
+globalStyle(`#storybook ${storybook}`, {
```
szymonmaslowski, line 66: Please move it to a separate storybook related helper file and import in the stories that require it. Let's not pollute the drawer with unnecesary styles.

packages/ui/src/design-system/side-drawer/side-drawer-overlay.css.ts
```diff
@@ -0,0 +1,24 @@
...
+  sx({
+    py: '$24',
+    px: '$40',
+    zIndex: '$overlay',
+  }),
+  {
+    backgroundColor: 'rgba(0, 0, 0, 0.45)',
+    position: 'fixed',
+    inset: 0,
+    animation: `${overlayShow} 150ms cubic-bezier(0.16, 1, 0.3, 1)`,
```
szymonmaslowski, line 22: Should we have animations defined at the theme level? IMO we could have a comon set of animations to aply to different components in order to achieve a consistency.

szymonmaslowski (changes requested):

packages/ui/src/design-system/side-drawer/side-drawer-content-headline.css.ts
renanvalentin: I spoke with Elena and the UX team is implementing a 1.5px for border and separator.

packages/ui/src/design-system/side-drawer/side-drawer-content.css.ts
renanvalentin: To prevent the position fixed and enable display on storybook, I've created a new content-card component and extracted the styling.

packages/ui/src/design-system/side-drawer/side-drawer-overlay.css.ts
renanvalentin: If the animation pops up again, I might think about doing something about it. But for now, it's okay to keep it within the overlay range.

pczeglik-iohk (approved):
To me it looks really good. I have nothing to point out. Thank you

packages/ui/src/design-system/side-drawer/side-drawer-content.component.tsx
```diff
@@ -0,0 +1,23 @@
...
+import classNames from 'classnames';
+
+import { ContentCard } from './side-drawer-content-card.component';
+import * as cx from './side-drawer-content.css';
+
+type Props = Omit<Dialog.DialogContentProps, 'className'> & { zIndex?: number };
+
+export const Content = ({
+  children,
+  zIndex,
```
szymonmaslowski, line 13: Please confirm: The purpose of the `zIndex` prop here is to allow to render the `Content` component without the `Overlay` component, right?

packages/ui/src/design-system/side-drawer/side-drawer-content-card.component.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import * as cx from './side-drawer-content-card.css';
+
+interface Props {
+  children: ReactNode;
+}
+
+export const ContentCard = ({
+  children,
+  ...props
```
szymonmaslowski, line 16: There are no more props than the `children`

packages/ui/src/design-system/side-drawer/side-drawer-content-card.component.tsx
```diff
@@ -0,0 +1,21 @@
...
+}: Readonly<Props>): JSX.Element => (
+  <Box {...props} className={classNames(cx.container)}>
```
szymonmaslowski, line 18: `classNames` is redundant here.

szymonmaslowski (approved):
Looks good! I left a few minor comments.

packages/ui/src/design-system/side-drawer/side-drawer-content-body.component.tsx
przemyslaw-wlodek: For me it's better to be explicit, but I have another concern about why are we having a different **name** for the file and the main export of that file. It disables some nice refactoring features (poor DX)

packages/ui/src/design-system/side-drawer/side-drawer-close.component.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import * as Dialog from '@radix-ui/react-dialog';
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export const Close = ({ children }: Readonly<Props>): JSX.Element => (
```
przemyslaw-wlodek, line 9: Do we need this component? What's the benefit of wrapping it in our naming if it has just the `asChild` property? 🤔 

Or should we introduce our `NavigationButtons.Close` here?

```
      <Dialog.Close asChild>
        <NavigationButtons.Close onClick={onCloseClick} />
      </Dialog.Close>
```

packages/ui/src/design-system/side-drawer/side-drawer-content-card.component.tsx
```diff
@@ -0,0 +1,21 @@
+import type { ReactNode } from 'react';
+import React from 'react';
+
+import classNames from 'classnames';
+
+import { Box } from '../box';
+
+import * as cx from './side-drawer-content-card.css';
+
+interface Props {
```
przemyslaw-wlodek, line 10: This may be a point to RFC but can we please name props according to their component names (`ContentCardProps`)? When we export components we should also export their props interfaces and then you need to rename them during re-exports which would be ugly. 😅

packages/ui/src/design-system/side-drawer/side-drawer-content-header.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+import * as Typography from '../typography';
+
+import { Close } from './side-drawer-close.component';
+import * as cx from './side-drawer-content-header.css';
+import { Separator } from './side-drawer-separator.component';
+
+interface Props {
+  text: string;
+  onBackClick?: () => void;
+  onCloseClick?: () => void;
```
przemyslaw-wlodek, line 15: Shouldn't this prop be required? We are always rendering the Close button.

packages/ui/src/design-system/side-drawer/side-drawer-content-header.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Typography.Body.Large weight="$bold" className={cx.text}>
+        {text}
+      </Typography.Body.Large>
+      <Close>
+        <NavigationButtons.Close onClick={onCloseClick} />
+      </Close>
+    </Flex>
+    <Box mb="$32">
+      <Separator />
+    </Box>
```
przemyslaw-wlodek, line 41: I think this should be part of the parent component which renders sub-components.

packages/ui/src/design-system/side-drawer/side-drawer-content-headline.css.ts
przemyslaw-wlodek: @mrcnk had the same question and Elena agreed to round these values to 2px. I also asked UI/UX team to round them by default to all future designs. Can we please change it?

packages/ui/src/design-system/side-drawer/side-drawer-overlay.css.ts
przemyslaw-wlodek: @mrcnk, you had some great ideas around that like using [framer-motion](https://www.framer.com/motion/). I think it's better to ignore the animation for now (MVP) than deliver the wrong approach which will spread around the codebase.

przemyslaw-wlodek (changes requested):

packages/ui/src/design-system/side-drawer/side-drawer-content-card.component.tsx
renanvalentin: In this context, what is the purpose of exporting the Props interface? From my perspective, it doesn't seem necessary, so I suggest leaving it as is.

Regarding naming convention, it may become a bit noisy if there is no ESLINT rule in place to guard it, as personal preferences may come into play.

packages/ui/src/design-system/side-drawer/side-drawer-content-header.component.tsx
renanvalentin: To ensure that the border is always displayed, it is important to include the `<Separator/>` in the `<Header/>`. Otherwise, the SideDrawer may be displayed without the border, which would be incorrect from a UI perspective.

```
<SideDrawer.Content>
    <SideDrawer.Header
      text="Label"
      onBackClick={onBackClick}
      onCloseClick={onCloseClick}
    />
    
    <SideDrawer.Body>
      <SideDrawer.Headline
        title="Section title"
        description="Lorem ipsum dolor sit amet quare id faciam."
      />
    </SideDrawer.Body>
```

packages/ui/src/design-system/side-drawer/side-drawer-content-headline.css.ts
przemyslaw-wlodek: I see your comment. What I'm trying to say is that @mrcnk also spoke with the Design team and told them that using fractions in px values is likely an antipattern. Then I spoke with the Design team and confirmed that Elena will update all our designs to use 2px instead of 1.5px. So could you update it, please?

packages/ui/src/design-system/side-drawer/side-drawer-content-header.component.tsx
przemyslaw-wlodek: Fine. I thought it was `Header` not `SideDrawerHeader` component.

packages/ui/src/design-system/side-drawer/side-drawer-overlay.css.ts
przemyslaw-wlodek: That's just an idea and we can leave it as is but I will create a ticket for future investigation.

packages/ui/src/design-system/side-drawer/side-drawer-content-headline.css.ts
renanvalentin: > I see your comment. What I'm trying to say is that @mrcnk also spoke with the Design team and told them that using fractions in px values is likely an antipattern. Then I spoke with the Design team and confirmed that Elena will update all our designs to use 2px instead of 1.5px. So could you update it, please?

It seems we have a miscommunication problem, my suggestion is to merge using 1.5px as per Figma and create a broader alignment with the UX team. 

If the UX team confirms and updates the designs to match the new spec, we go back and refactor it. It is unlikely this discussion will get sorted out in a few minutes and I want to merge before closing the sprint.

przemyslaw-wlodek (approved):

PR #135, opened 2023-06-16
feat(ui): add simple Logger

przemyslaw-wlodek:
# Checklist

- [x] JIRA - LW-7034
- [x] Screenshots added.

---

## Proposed solution

Work in progress...

## Screenshots
image

### DIFF
--- a/packages/ui/src/logger.ts
+++ b/packages/ui/src/logger.ts
@@ -1,0 +1,64 @@
+type LoggableComponent = 'PieChart';
+type LogLevel = 'error' | 'warn';
+
+interface LogConfig {
+  componentName: string;
+  data?: unknown;
+  message: string[] | string;
+}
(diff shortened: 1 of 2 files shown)


packages/ui/src/logger.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Please ensure that log messages are clear and concise.
+ * Include corresponding data either in the `data` property or as `${interpolation}` in the message.
+ * If message is a `string[]` it will be rendered multiline.
+ */
+const getLogger = (componentName: LoggableComponent): Logger => ({
+  warn: getLogFunctionByLevel('warn', componentName),
+  error: getLogFunctionByLevel('error', componentName),
+});
+
+const getMockLogger: typeof getLogger = () => ({
```
mchappell, line 58: Instead of adding mock loggers, why not just return empty if production?

PR #145, opened 2023-06-19
feat(ui): create <SearchBox /> component

renanvalentin:
# Checklist

- [x] JIRA - LW-6619
- [x] Screenshots added.

---

## Proposed solution

```jsx
  <SearchBox
            placeholder="Search"
            value={value}
            onChange={setValue}
          />
```

## Testing

The test check if the value is filled properly and cleared after clicking on the clear button.

## Screenshots

capture-1c836bc4

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/496/5316078396/index.html) for [4ed24fce](https://github.com/input-output-hk/lace/pull/145/commits/4ed24fce6d4a3a48107a17e1bd970ac4985db88c)
| Total | 13     | 6      | 0       | 0     | 19    | ❌     |


packages/ui/src/design-system/search-box/search-box.component.tsx
```diff
@@ -0,0 +1,102 @@
...
+        disabled={disabled}
+      />
+      {shouldShowClearButton ? (
+        <ClearButton
+          onClick={handleClearClick}
+          active={shouldShowClearButton}
+          aria-label={clearButtonAriaLabel}
+          ref={clearButtonReference}
+        />
+      ) : undefined}
```
pczeglik-iohk, line 99: I'm just curious, is there any particular reason we are returning `undefined` instead of `null`. More often than not I see `null` in such code, but I believe there is a slight difference and would like to learn :)

packages/ui/src/design-system/search-box/search-box.component.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const handleClick = (): void => {
+    if (inputReference.current) {
+      inputReference.current.focus();
+    }
+  };
+
+  const handleClearClick = (
+    event: Readonly<MouseEvent<HTMLButtonElement>>,
+  ): void => {
+    event.stopPropagation();
```
pczeglik-iohk, line 46: should we always stop propagation? or only in case a developer provided `onClear` callback function?

renanvalentin: The `onClear` should be provided. Otherwise, the behavior would be broken. This makes me think that the property should be required.

pczeglik-iohk (approved):

PR #180, opened 2023-06-26
feat(ui): create <FlowCard /> component

renanvalentin:
# Checklist

- [x] JIRA - LW-6620
- [x] Screenshots added.

---

## Proposed solution

A set of components to display the Flow Card in different ways.

```jsx
import { FlowCard } from '@lace/ui'

          <FlowCard.Card>
            <FlowCard.Profile
              imageSrc={cardanoImage}
              name="Title"
              description="Subtitle"
            />
            <FlowCard.Details title="amount" subtitle="Value" />
          </FlowCard>
        </PopUpView>
```

## Screenshots

capture-58796631

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/637/5382178332/index.html) for [20e5be3f](https://github.com/input-output-hk/lace/pull/180/commits/20e5be3f8a657af01e8269f16bcb4197ebb9fb67)
| Total | 33     | 3      | 0       | 0     | 36    | ❌     |


packages/ui/src/design-system/flow-card/flow-card.component.tsx
```diff
@@ -0,0 +1,15 @@
...
+import * as cx from './flow-card.css';
+
+import type { OmitClassName } from '../../types';
+
+type Props = OmitClassName<'div'> & {
+  children: React.ReactNode;
+};
+
+export const Card = ({ children, ...props }: Readonly<Props>): JSX.Element => (
+  <div className={cx.container} {...props}>
```
szymonmaslowski, line 12: Please use the `Card.Greyed` component as a base as it already provides a unified card styling that covers the flow card case.

renanvalentin: Thanks for suggesting the `Card.Greyed`. They have different background colors in dark mode: Flow Card -> #2F2F2F, Greyed #2E2E2E

Screenshot 2023-06-27 at 09 13 58

It would be interesting to have "base" card components listed as references in Figma. Would you be willing to discuss this with the UX team for future components? 

Screenshot 2023-06-27 at 09 15 48

szymonmaslowski: > Flow Card -> #2F2F2F, Greyed #2E2E2E

Then the `Card.Greyed` should have the `#2F2F2F` color from what I can see, because the other case where it was used which is the delegation card also has the `#2F2F2F` color. That is strange, probably it changed from the time I developed the Card component. Could you please change the background color of the `Card.Greyed` to `#2F2F2F` and use it in Flow Card?

> It would be interesting to have "base" card components listed as references in Figma. Would you be willing to discuss this with the UX team for future components?

It would be great, agree. Unfortunatelly, I won't be able to handle that because of limited time.

przemyslaw-wlodek: Apologies if I'm mistaken but as far as I understand it's better to clarify the color with the UX team than introduce a new component. Please let's try to keep it consistent and reusable. If it's not ready to be delivered in the `ui-toolkit` then please keep it in the project itself (like we did for some of the work with the staking package).

renanvalentin: @przemyslaw-wlodek The Cards section on Figma contains:

- Education Card
- New address card
- Upload / download card
...
- Flow Card

Here's a good example: the `Education Card` and the `Flow Card`. One is colored `mid-black`, while the other is `dark grey`. Both of these are valid design tokens.

As I understood, you have the desire to build the Cards section using a Card Skeleton (padding, border radius, etc.) ? 
If you go through the Cards section, you will notice different variants of card styling. That's why I mentioned the need for the reference Card linked at the bottom of the specification:

Screenshot 2023-06-27 at 09 15 48


We would need someone to take ownership and coordinate this topic with the UX team to achieve this. I don't have this capacity right now. 

My proposed solution re-uses the design tokens to build the Flow Card specifically and is not a general-purpose solution.

przemyslaw-wlodek: Ok, now I understand your concerns and I'm happy that you've reused as much as possible. Then from my POV, we can get this sorted in the next PI. Thanks

szymonmaslowski: Hey @renanvalentin. During the development of the `Card` component I identified the common parts of all the card components present in figma and I consulted and agreed on them with the UX team, so the `Card` component is already designed to be used as a base for all the cards as it fits all all the cases. It would be awesome to have the Card component present in figma and referenced under each of the card cases for the calrity and transparecy. We can start that initiative once we have some free capacity.

The intention of the `Card` component is to provide a base styling only for areas that are common for all the card cases and those are:
 - box shadow
 - border and border radius
 - background color

All the rest style properties like paddings differ between the cases so they should be specified for each case separately. 

Regarding your example of the `Education Card` vs the `Flow Card` they follow different base variants: the `Education Card` is a specific case of the `Card.Outlined` component which has the mid-black background and the `Flow Card` is a case of `Card.Greyed` compoent with the dark-grey-plus background color.

Please take a look on the card background color config for the dark mode:
image

 As you could notice in the storybook, there are three variants of the `Card` component:
 - `Card.Elevated` (a base for the `Address card` or `Staking funds card`)
 - `Card.Greyed` (a base for the `Delegation card` or `Flow card`
 - `Card.Outlined` (a base for majority of cards, e.g. `Educational card` or `New address card`).

renanvalentin: @szymonmaslowski can we resolve this conversation since this is being discussed on Figma with the UX team?

PR #219, opened 2023-07-05
chore(extension): fix cursor position for unformatted values in coin input

vetalcore:
# Checklist

- [x] https://input-output.atlassian.net/browse/LW-6469
- [ ] Screenshots added.

---

## Proposed solution

- fire `focus({ cursor: 'end' });` event in case user clicks outside of input but within the clickable area;
- fire `focus({ cursor: 'end' });` event in case user clicks on input but `compactValue` differs from the actual one

## Testing

## Screenshots

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/843/5484473895/index.html) for [c123fdd5](https://github.com/input-output-hk/lace/pull/219/commits/c123fdd57eaf85a7c7eb16152f82b85ce944c2ba)

### DIFF
--- a/packages/core/src/ui/components/AssetInput/AssetInput.tsx
+++ b/packages/core/src/ui/components/AssetInput/AssetInput.tsx
@@ -89,11 +89,20 @@ export const AssetInput = ({
 
   useLayoutEffect(() => {
     if (focused) {
-      inputRef.current.focus({
-        cursor: 'end'
+      // place cursor at the end of the input in case user clicks outside of input but within the clickable area
+      inputRef.current.focus(document.activeElement !== inputRef.current.input ? { cursor: 'end' } : undefined);
+    }
+  }, [compactValue, focused, value]);
+
+  const onClick = useCallback(() => {
+    // place cursor at the end of the input in case the formatted(compact) value is different from the real one
+    if (value !== compactValue && !focused) {
+      setTimeout(() => {
+        inputRef.current.focus({ cursor: 'end' });
       });
     }
-  }, [focused]);
+  }, [compactValue, value, focused]);
+
   const { t } = useTranslate();
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/core/src/ui/components/AssetInput/AssetInput.tsx
```diff
@@ -89,11 +89,18 @@ export const AssetInput = ({
...
-        cursor: 'end'
-      });
+      // place cursor at the end of the input in case user clicks outside of input but within the clickable area
+      inputRef.current.focus(document.activeElement !== inputRef.current.input ? { cursor: 'end' } : undefined);
+    }
+  }, [compactValue, focused, value]);
+
+  const onClick = useCallback(() => {
+    // place cursor at the end of the input in case the formatted(compact) value is different from the real one
+    if (value !== compactValue) {
```
lgobbi-atix, line 99: What about `displayValue`? Shouldn't we check that too?

vetalcore: in our case value and this one would be equal

mchappell (approved):

lgobbi-atix (approved):

miorsufianiohk:
Hi @vetalcore ,

I need a confirmation from @jackrousian or yourself please that this behaviour is correct as it is not mentioned in requirement (as attached below):

If the **display value was the same as the real value**, the first time I clicked in between numbers, the cursor was always positioned at the end of the value. When I clicked again the second time, then only the cursor was positioned where I wanted it to be (ie: in between the numbers). 

When testing **display value was the same as the real value**, I expected the first time I clicked in between the numbers, the cursor would be positioned immediately where I wanted it to be. (not at the end)

[Please see video.](https://drive.google.com/file/d/1EPDgtB8CLbY-9Uh6-Z0DGurJXb_GlwcB/view?usp=sharing)


cursorRequirements

PR #235, opened 2023-07-07
feat: persists HandleResolution to addressBookContext

VanessaPC:
# Checklist

- [x] [JIRA-7379] 
- [ ] Screenshots added.

---

## Proposed solution

Adds `HandleResolution` to the `AddressBookContext` so that later on we can save and compare both values for a validation on the client side. 

## Testing

- Run the app make sure ADA Handle works without any regression.
- Address book renders 
- Profile renders and they both render the address and the handle. 

## Screenshots

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/858/5487037446/index.html) for [b81d60a3](https://github.com/input-output-hk/lace/pull/235/commits/b81d60a385606e54c1803abcf4ded26a959b9974)
| Total | 28     | 8      | 0       | 0     | 36    | ❌     |

### DIFF
--- a/apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressDetailDrawer.tsx
+++ b/apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressDetailDrawer.tsx
@@ -77,7 +77,7 @@ export const AddressDetailDrawer = ({
     () => ({
       name: validateWalletName,
       address: validateWalletAddress,
-      handle: async (value: string) => await validateWalletHandle(value, handleResolver)
+      handle: async (value: string) => await validateWalletHandle({ value, handleResolver })
     }),
     [handleResolver]
   );
--- a/apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
+++ b/apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
@@ -4,11 +4,13 @@ import { Wallet } from '@lace/cardano';
 import { DbQueries } from '../hooks';
 import { AddressRecordParams } from '@src/features/address-book/context';
 import { sortTabletByName } from '../helpers';
+import { Cardano, HandleResolution } from '@cardano-sdk/core';
 
 export interface AddressBookSchema {
   id: number;
   name: string;
-  address: string;
+  address: string | Cardano.PaymentAddress;
+  handleResolution?: HandleResolution;
   network: Wallet.Cardano.NetworkMagics;
 }
 
@@ -16,7 +18,7 @@ export const addressBookSchema: VersionedSchema = {
   table: 'addressBook',
   indexedFields: {
     1: ['++id', 'name', 'address'],
-    2: ['++id', 'name', 'address', 'network']
+    2: ['++id', 'name', 'address', 'network', 'handleResolution']
   }
 };
 
--- a/apps/browser-extension-wallet/src/utils/validators/address-book.ts
+++ b/apps/browser-extension-wallet/src/utils/validators/address-book.ts
@@ -8,6 +8,7 @@ import { addressErrorMessage, nameErrorMessage } from '@lib/storage/helpers';
 import { TOAST_DEFAULT_DURATION } from '@hooks/useActionExecution';
 import ErrorIcon from '@assets/icons/address-error-icon.component.svg';
 import { HandleProvider, HandleResolution } from '@cardano-sdk/core';
+import { isHandle } from '@lace/core';
 
 const MAX_ADDRESS_BOOK_NAME_LENGTH = 20;
 
@@ -19,7 +20,6 @@ export const verifyHandle = async (
 ): Promise<ValidationResult & { handles?: HandleResolution[] }> => {
   try {
     const resolvedHandles = await handleResolver.resolveHandles({ handles: [value.slice(1)] });
-
     if (resolvedHandles.length === 0) {
       return { valid: false };
     }
@@ -59,11 +59,19 @@ export const validateWalletAddress = (address: string): string => {
   return !isValid ? i18n.t('browserView.addressBook.form.incorrectCardanoAddress') : '';
 };
 
-export const validateWalletHandle = async (value: string, handleResolver: HandleProvider): Promise<string> => {
-  const res = (await verifyHandle(value, handleResolver)).handles;
-  if (!res) {
-    return i18n.t('general.errors.incorrectHandle');
+type validateWalletHandleArgs = {
+  value: string;
+  handleResolver: HandleProvider;
+};
+
+export const validateWalletHandle = async ({ value, handleResolver }: validateWalletHandleArgs): Promise<string> => {
+  const response = await verifyHandle(value, handleResolver);
+  const handles = response.handles;
+
+  if (!handles) {
+    throw new Error(i18n.t('general.errors.incorrectHandle'));
   }
+
   return '';
 };
 
@@ -124,4 +132,19 @@ export const hasAddressBookItem = (
     ];
 
   return [false, undefined];
+};
+
+export const getAddressToSave = async (
+  address: AddressBookSchema | Omit<AddressBookSchema, 'id'> | Omit<AddressBookSchema, 'id' | 'network'>,
+  handleResolver: HandleProvider
+): Promise<AddressBookSchema | Omit<AddressBookSchema, 'id'> | Omit<AddressBookSchema, 'id' | 'network'>> => {
+  if (isHandle(address.address)) {
+    const result = await verifyHandle(address.address, handleResolver);
+
+    if (result.valid) {
+      return { ...address, handleResolution: result.handles[0] };
+    }
+  }
+
+  return address;
 };
--- a/apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressForm/AddressForm.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressForm/AddressForm.tsx
@@ -32,7 +32,7 @@ export const AddressForm = ({ initialValues, onConfirmClick }: AddressFormProps)
     () => ({
       name: validateWalletName,
       address: validateWalletAddress,
-      handle: async (value: string) => await validateWalletHandle(value, handleResolver)
+      handle: async (value: string) => await validateWalletHandle({ value, handleResolver })
     }),
     [handleResolver]
   );
--- a/apps/browser-extension-wallet/src/features/address-book/components/AddressBook.tsx
+++ b/apps/browser-extension-wallet/src/features/address-book/components/AddressBook.tsx
@@ -20,6 +20,8 @@ import {
 } from '@providers/AnalyticsProvider/analyticsTracker';
 import { useAnalyticsContext } from '@providers';
 import { AddressDetailsSteps } from './AddressDetailDrawer/types';
+import { getAddressToSave } from '@src/utils/validators';
+import { useHandleResolver } from '@hooks';
 
 const scrollableTargetId = 'popupAddressBookContainerId';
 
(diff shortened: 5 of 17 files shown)


apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
```diff
@@ -20,6 +22,11 @@ export const addressBookSchema: VersionedSchema = {
   }
 };
 
+export const isAddressAHandle = (value: Cardano.PaymentAddress | string): value is string => value.startsWith('$');
+
+export const isAddressAPaymentAddress = (value: Cardano.PaymentAddress | string): value is Cardano.PaymentAddress =>
+  !value.startsWith('$');
```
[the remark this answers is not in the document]
VanessaPC: 8ecce2a0cafb7e513925205c7014434508e60439 should address this

apps/browser-extension-wallet/src/utils/validators/address-book.ts
```diff
@@ -7,19 +7,18 @@ import { ToastProps } from '@lace/common';
...
 import ErrorIcon from '@assets/icons/address-error-icon.component.svg';
-import { HandleProvider, HandleResolution } from '@cardano-sdk/core';
+import { HandleProvider, HandleResolution, Cardano } from '@cardano-sdk/core';
 
 const MAX_ADDRESS_BOOK_NAME_LENGTH = 20;
 
 const hasWhiteSpace = (s: string) => s.trim() !== s;
 
-export const verifyHandle = async (
+export const hasHandleOwnerChanged = async (
```
lucas-barros, line 16: I'm confused about the function name. I would expect `hasHandleOwnerChanged` to return a boolean, but it's returning handles.

apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressForm/AddressForm.tsx
```diff
@@ -32,7 +32,7 @@ export const AddressForm = ({ initialValues, onConfirmClick }: AddressFormProps)
     () => ({
       name: validateWalletName,
       address: validateWalletAddress,
-      handle: async (value: string) => await validateWalletHandle(value, handleResolver)
+      handle: async (value: string) => await validateWalletHandle(value, undefined, handleResolver)
```
lucas-barros, line 35: I would consider turning the arguments to an object.

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
```diff
@@ -20,6 +22,13 @@ export const addressBookSchema: VersionedSchema = {
   }
 };
 
+const matchHandleRegex = /\$[\w.\-]+$/g;
+export const isAddressAHandle = (value: Cardano.PaymentAddress | string): value is string =>
```
lucas-barros, line 26: It seems these are not used anywhere.

lucas-barros (changes requested):

apps/browser-extension-wallet/src/utils/validators/address-book.ts
VanessaPC: Well, it returns a validation and the handles when needed. 
It returns a boolean value based on whether the owner of the handle has changed, and the handles, because in the cases in which the owner has changed and we want to persist the new information we need the handle data. 
So, to me, it makes more sense to call it `hasHandleOwnerChanged` or something similar than `verifyHandles` because... verify handles isn't explicitly telling us what is proving, so when we are reading the code I'd rather know what is doing? 
Does this make sense? Also, are there other names you think could describe it better? or do you think it should be the previous one?

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
VanessaPC: They aren't used here. That's right. I added this to be used in the PR of this ticket: LW-7128, where they are used, because even though they aren't applied here they still fell into context for me. But I don't mind removing them and adding them one PR later if you'd rather.

greatertomi: Yes. I'll suggest removing them and re-adding them later when they are used. 

Also, I'll suggest you move this and `isAddressAPaymentAddress` to somewhere in the [utils](https://github.com/input-output-hk/lace/tree/cdefe01a7417a38f056fc0384c2c7507f073470a/apps/browser-extension-wallet/src/utils) folder. They fit better there than in the `lib/storage` folder.

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
```diff
@@ -20,6 +22,13 @@ export const addressBookSchema: VersionedSchema = {
...
+  matchHandleRegex.test(value);
+
+export const isAddressAPaymentAddress = (value: Cardano.PaymentAddress | string): value is Cardano.PaymentAddress =>
```
greatertomi, line 29: This is also not used. Same comment here applies

greatertomi (changes requested):

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
VanessaPC: 8d5bb920bcda7e00ee165cd9ae5228769825b011 should address this.

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
VanessaPC: 8d5bb920bcda7e00ee165cd9ae5228769825b011 should address this.

greatertomi (approved):

apps/browser-extension-wallet/src/utils/validators/address-book.ts
lucas-barros: > It returns a boolean value based on whether the owner of the handle has changed

I don't see this part. It seems the return is not related to whether the handle owner has changed or not, just if it's able to resolve the handle or not. Maybe I'm missing something.

VanessaPC: I think the set up in preparation for the following PR is what's causing the confusion. I reverted. 

4f942090a669a336799e6545ce9ae2ec2851b1c0 should resolve this.

apps/browser-extension-wallet/src/lib/translations/en.json
```diff
@@ -85,7 +85,8 @@
       "incorrectAddress": "Incorrect address",
       "incorrectHandle": "Handle not found",
       "wrongNetworkAddress": "This address belongs to a different network",
-      "wrongAddressEra": "Byron addresses are not supported, please use a Shelley address instead."
+      "wrongAddressEra": "Byron addresses are not supported, please use a Shelley address instead.",
+      "handleConflict": "The is a conflict with the handle. We received this {{receivedAddress}} but expected {{actualAddress}}."
```
lucas-barros, line 89:
```suggestion
      "handleConflict": "There is a conflict with the handle. We received this {{receivedAddress}} but expected {{actualAddress}}."
```

lucas-barros (changes requested):

apps/browser-extension-wallet/src/lib/translations/en.json
VanessaPC: 0ae0937cbd1fd21ce4decb3202e22f50242eebcb should resolve this

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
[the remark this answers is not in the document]
VanessaPC: This was decided to revert and add on the next PR.

VanessaPC:
In the middle of the feedback loop my `main` branch and this branch stopped working likely because of cache issues, that's what triggered me merging main into the branch.

apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressDetailDrawer.tsx
```diff
@@ -77,7 +77,7 @@ export const AddressDetailDrawer = ({
     () => ({
       name: validateWalletName,
       address: validateWalletAddress,
-      handle: async (value: string) => await validateWalletHandle(value, handleResolver)
+      handle: async (value: string) => await validateWalletHandle({ value, address: undefined, handleResolver })
```
rhyslbw, line 80:
```suggestion
      handle: async (value: string) => await validateWalletHandle({ value, handleResolver })
```
?

apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressForm/AddressForm.tsx
```diff
@@ -32,7 +32,7 @@ export const AddressForm = ({ initialValues, onConfirmClick }: AddressFormProps)
     () => ({
       name: validateWalletName,
       address: validateWalletAddress,
-      handle: async (value: string) => await validateWalletHandle(value, handleResolver)
+      handle: async (value: string) => await validateWalletHandle({ value, address: undefined, handleResolver })
```
rhyslbw, line 35:
```suggestion
      handle: async (value: string) => await validateWalletHandle({ value, handleResolver })
```

rhyslbw (changes requested):

apps/browser-extension-wallet/src/lib/storage/models/address-book.ts
```diff
@@ -4,11 +4,13 @@ import { Wallet } from '@lace/cardano';
 import { DbQueries } from '../hooks';
 import { AddressRecordParams } from '@src/features/address-book/context';
 import { sortTabletByName } from '../helpers';
+import { Cardano, HandleResolution } from '@cardano-sdk/core';
 
 export interface AddressBookSchema {
   id: number;
   name: string;
-  address: string;
+  address: string | Cardano.PaymentAddress;
```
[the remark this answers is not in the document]
VanessaPC: Currently this will be out of sync when the user enters a handle that comes, on the validation step, with a new resolved address. 
So they will be out of sync. Because until the user 'updates', the previous address will not match the new address from the handle. And this check will run on opening the book. 
`address: string` is used everywhere, not just in the address book, but all through transactions, in the form. So If I change it to make it match the type above, the db will change, the forms will change, and the transactions will as well. And this is a far bigger refactor.

apps/browser-extension-wallet/src/utils/validators/address-book.ts
```diff
@@ -7,19 +7,18 @@ import { ToastProps } from '@lace/common';
...
   value: string,
```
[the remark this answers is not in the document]
VanessaPC: I can't do this in this PR because I don't have the address saved to Provider, that work is in the next PR. In the one were I show the warning. Not because I don't have the address persisted which now I do, but simply because the logic isn't set up in the addressbook because it belongs on the other one.

apps/browser-extension-wallet/src/utils/validators/address-book.ts
```diff
@@ -59,12 +58,26 @@ export const validateWalletAddress = (address: string): string => {
   return !isValid ? i18n.t('browserView.addressBook.form.incorrectCardanoAddress') : '';
 };
 
-export const validateWalletHandle = async (value: string, handleResolver: HandleProvider): Promise<string> => {
-  const res = (await verifyHandle(value, handleResolver)).handles;
-  if (!res) {
-    return i18n.t('general.errors.incorrectHandle');
+export const validateWalletHandle = async (
+  value: string,
+  address: Cardano.PaymentAddress,
```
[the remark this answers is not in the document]
VanessaPC: https://github.com/input-output-hk/lace/pull/235#discussion_r1256286598 
This is where I pass the address and do the check. 
And at this point throw the error if they conflict. Like we saw on the demo. 
At this point if the address  is undefined, which now it is, because we don't have the logic yet on the components for this... we simply avoid error and it behaves like before.

apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressForm/AddressForm.tsx
VanessaPC: 85ba81cb0a41d5d43cc05dca2f7b305125fb49c9 fixes this

apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressDetailDrawer.tsx
VanessaPC: 85ba81cb0a41d5d43cc05dca2f7b305125fb49c9 fixes this

rhyslbw (approved):

lucas-barros (changes requested):

apps/browser-extension-wallet/src/utils/validators/address-book.ts
```diff
@@ -59,11 +58,33 @@ export const validateWalletAddress = (address: string): string => {
...
-export const validateWalletHandle = async (value: string, handleResolver: HandleProvider): Promise<string> => {
-  const res = (await verifyHandle(value, handleResolver)).handles;
-  if (!res) {
-    return i18n.t('general.errors.incorrectHandle');
+type validateWalletHandleArgs = {
+  value: string;
+  handleResolver: HandleProvider;
+};
+
+export const getHandleResolution = async ({
```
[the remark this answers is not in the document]
lucas-barros: We would not be changing `validateWalletHandle` or `verifyHandle`. Just remove `getHandleResolution` and replace with `verifyHandle`.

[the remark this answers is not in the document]
lucas-barros: > It's the same as with a piece of code that gets repeated 3 times so far, but then.. it needs to change once more once I switch to where I need the work.
But if you remove the duplication now then you would only need to change it in one place later. 

The issue I see is if in the mean time someone touches that code it would be harder to refactor or may lead to conflicts. I don't see a benefit to resolve this in another PR, but I'm not enforcing anything so feel free to do as you think is best.

lucas-barros (approved):

dmitrii-gaico (approved):

PR #236, opened 2023-07-07
Separate events being developed and production-ready events

jplorek-atix:
# Checklist

- [x] JIRA - \<[LW-7348](https://input-output.atlassian.net/browse/LW-7348)>
- [ ] Screenshots added.

---

## Proposed solution

This PRs implements a environment variable `PRODUCTION_MODE_TRACKING` to enable and disable post hog events in production, also we have two maps with post hog projects tokens, one for production and the other for development.
The idea is not mix event that are being developed and track wrong user activity in production.

When `PRODUCTION_MODE_TRACKING` is `true`, we will use the map with the project tokens for production, if not true we use development project tokens.

## Testing

- [ ] Check if right PostHog projects are used depending on `PRODUCTION_MODE_TRACKING` environment value (`false` or `true`).
- [ ] Check that by default, events are sent to instances suffixed with `(development/testing)`

[LW-7348]: https://input-output.atlassian.net/browse/LW-7348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser-extension-wallet/src/providers/AnalyticsProvider/postHog/config.ts
```diff
@@ -3,13 +3,21 @@ import { PostHogConfig } from 'posthog-js';
...
-export const NETWORK_ID_TO_POSTHOG_TOKEN_MAP: Record<Wallet.Cardano.NetworkMagic, string> = {
+export const DEV_NETWORK_ID_TO_POSTHOG_TOKEN_MAP: Record<Wallet.Cardano.NetworkMagic, string> = {
   [Wallet.Cardano.NetworkMagics.Mainnet]: 'phc_gH96Lx5lEVXTTWEyytSdTFPDk3Xsxwi4BqG88mKObd1',
   [Wallet.Cardano.NetworkMagics.Preprod]: 'phc_Xlmldm6EYSfQVgB9Uxm3b2xC1noDlgFFXpF9AJ6SMfJ',
   [Wallet.Cardano.NetworkMagics.Preview]: 'phc_e8SaOOWpXpNE59TnpLumeUjWm4iv024AWjhQqU406jr'
 };
 
+export const PRODUCTION_NETWORK_ID_TO_POSTHOG_TOKEN_MAP: Record<Wallet.Cardano.NetworkMagic, string> = {
+  // TODO: add the correct project token
+  [Wallet.Cardano.NetworkMagics.Mainnet]: '1',
```
jplorek-atix, line 16: We might want to add this to env variables, by doing so, we will have more control over the projects where we want to send the events

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -49,7 +53,14 @@ export class AnalyticsTracker {
...
   }
 
-  async sendEventToPostHog(action: PostHogAction, properties: Record<string, string | boolean> = {}): Promise<void> {
+  async sendEventToPostHog(
+    action: PostHogAction,
+    properties: Record<string, string | boolean> = {},
+    flow?: PostHogFlows // create an enum for this
+  ): Promise<void> {
+    // adds an extra check to not track flows in production that are not approved yet, at the same time, if we find an issue with a flow in production, we can disable
+    // should we do this only when production tracking mode is enabled?
```
jplorek-atix, line 62: What I mean with this is add something like 
```
const shouldIgnoreEvent = PRODUCTION_TRACKING_MODE_ENABLED ? !ENABLED_TRACKING_FLOWS.includes(flow) : false
if (shouldIgnoreEvent) return;
```

So, if we are in development, all flows will be enabled by default, we don't need to configure anything, we just add the events and the tracking should continue working as before.

We only need to enable the flows that we want to track on production.

marcin-mazurek: I think it's a bit beyond scope of the ticket :) Details discussed on Slack - I would personally focus only on switching between dev and prod projects in this PR.

apps/browser-extension-wallet/src/providers/AnalyticsProvider/postHog/config.ts
marcin-mazurek: Personally I don't have a strong opinion on declaring them via environment variables, those are anyway public tokens attached to every request, and they are unlikely to change. In Matomo we had site IDs and we defined them in JS config files them too. @DominikGuzei what's your take on that?

Also, can you provide here the actual tokens from PostHog?

apps/browser-extension-wallet/.env.example
```diff
@@ -41,6 +41,8 @@ TERMS_OF_USE_URL=https://www.lace.io/lace-terms-of-use.pdf
 # events tracking
 MATOMO_API_ENDPOINT=https://matomo.cw.iog.io/matomo.php
 PUBLIC_POSTHOG_HOST=https://eu.posthog.com
+ENABLED_TRACKING_FLOWS=onboarding,send,nft
+PRODUCTION_MODE_TRACKING=false
```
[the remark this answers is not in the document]
marcin-mazurek: The goal is to set this variable automatically during the release process, so we don't have to provide it manually

.github/workflows/release.yml
```diff
@@ -36,6 +36,16 @@ jobs:
         run: |
           sed -i "s/\"version\": \"[0-9.]*\"/\"version\": \"${{ inputs.version }}\"/g" apps/browser-extension-wallet/manifest.json
 
+      - name: Enable production tracking
+        run: |
+          if grep -q "PRODUCTION_MODE_TRACKING=true" apps/browser-extension-wallet/.env; then
```
marcin-mazurek, line 41: `.env` is in `.gitignore` so I those changes won't be commited when a PR is created. Looking at the current release, we seem to be editing `.env.default` when preparing a release build: https://github.com/input-output-hk/lace/pull/216/files

What I started to wonder about is, whether this workflow is actually used? 🤔 Do you happen to know, from your previous experience? I was never involved in the release process, but I can't see any workflow runs: https://github.com/input-output-hk/lace/actions/workflows/release.yml

In any case, I asked for clarification here: https://input-output-rnd.slack.com/archives/C01S2DD5B0D/p1689024991210549

.github/workflows/release.yml
```diff
@@ -36,6 +36,16 @@ jobs:
...
+            echo "Production tracking is enabled"
+          elif grep -q "PRODUCTION_MODE_TRACKING=false" apps/browser-extension-wallet/.env; then
+            sed -i "s/\"PRODUCTION_MODE_TRACKING=false\"/\"PRODUCTION_MODE_TRACKING=true\"/" apps/browser-extension-wallet/.env
```
marcin-mazurek, line 44: This command fails for me (on macOS 12.5) - can you double check? I tried other cases and those work fine, just the sed syntax doesn't seem to work

```
➜  lace git:(feat/lw-6791-opted-out-tracking) sed -i "s/\"PRODUCTION_MODE_TRACKING=false\"/\"PRODUCTION_MODE_TRACKING=true\"/" apps/browser-extension-wallet/.env
sed: 1: "apps/browser-extension- ...": command a expects \ followed by text
```

.github/workflows/release.yml
marcin-mazurek: As discussed on Slack - let's remove it for now.

marcin-mazurek (approved):

bslabiak (approved):

PR #260, opened 2023-07-12
feat(staking): enable sign step in staking drawer

przemyslaw-wlodek:
# Checklist

- [x] JIRA - LW-7096
- [x] Screenshots added.

---

## Proposed solution

Enable multi-staking transaction signing step. This PR also enables the simplest, incomplete draft of multi-staking flow.

## Screenshots
image

### DIFF
--- a/packages/staking/src/features/drawer/StakePoolConfirmation.tsx
+++ b/packages/staking/src/features/drawer/StakePoolConfirmation.tsx
@@ -130,7 +130,7 @@ const MultipleStakePoolConfirmationBody = ({
             subText={<span>{stakePool.ticker}</span>}
           />
           <div className={styles.itemData}>
-            <Ellipsis beforeEllipsis={10} afterEllipsis={8} text={stakePool.id} ellipsisInTheMiddle />;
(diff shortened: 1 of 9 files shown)


mrcnk (approved):

packages/staking/src/features/drawer/StakePoolConfirmation.tsx
```diff
@@ -303,7 +306,7 @@ export const StakePoolConfirmationFooter = ({ popupView }: StakePoolConfirmation
       setIsConfirmingTx(true);
       try {
         await signAndSubmitTransaction();
-        const isDelegating = !!(rewardAccounts && delegationDetails);
+        const isDelegating = !!(rewardAccounts && (delegationDetails || draftPortfolio.length > 0));
```
szymonmaslowski, line 309: I think this var representing the case when user already submitted the delegation and it is pending or already passed, so Probably we should not check for selections here.
`rewardAccounts` represents the on chain delegation state and delegationDetails is a data extracted from the `rewardAccounts`

przemyslaw-wlodek: Thanks a lot for this catch! We need to refine it. It's all subject to greater refactoring though.

PR #304, opened 2023-07-21
feat(extension): use public key hash as id for opted in users

jplorek-atix:
# Checklist

- [ ] JIRA - \<[LW-7574](https://input-output.atlassian.net/browse/LW-7574)>
- [ ] Screenshots added.
- [ ] Fix Broken unit tests
- [ ] Add unit test for new methods added in user id service
---

## Proposed solution

This PR implements the hashed wallet public key as id for opted in users, to be able to keep tracking across browsers. For opted out users, the ID keeps being randomized

## Testing

- Start any of the setup flows and complete all the steps. 
- On the analytics step, agree to be tracked. 
- Go to post hog dashboard.
- For all steps before wallet creation you will see a random ID in post hog, after the wallet is created you will see an Alias event with a new ID. 
- Click on the ID and go to the events tab, you should see the previous event for the random ID, linked to this new ID

## Screenshots

[LW-7574]: https://input-output.atlassian.net/browse/LW-7574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@d1terqjryk7mu9.cloudfront.net/linux/chrome/1110/5648265212/index.html) for [b9005b86](https://github.com/input-output-hk/lace/pull/304/commits/b9005b867002ef682f67328fe7a32776b55c411a)
| Total | 31     | 4      | 0       | 0     | 35    | ❌     |

### DIFF
--- a/apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
+++ b/apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
@@ -1,11 +1,11 @@
 /* eslint-disable no-magic-numbers */
 /* eslint-disable import/imports-first */
 /* eslint-disable prettier/prettier */
-const mockHashId =
+const mockWalletBasedId =
   '15d632f6b0ab82c72a194d634d8783ea0ef5419c8a8f638cb0c3fc49280e0a0285fc88fbfad04554779d19bec4ab30e5afee2f9ee736ba090c2213d98fe3a475';
 const mockBlake2b = jest.fn(() => ({
   update: () => ({
-    digest: () => mockHashId
+    digest: () => mockWalletBasedId
   })
 }));
 
@@ -32,13 +32,13 @@ describe('userIdService', () => {
       };
       const { getStorageMock } = generateStorageMocks(store);
       const userIdService = new UserIdService(getStorageMock);
-      expect(await userIdService.getId()).toEqual(store.userId);
+      expect(await userIdService.getRandomizedUserId()).toEqual(store.userId);
     });
 
     it('should generate a random user id if none was stored', async () => {
       const { getStorageMock } = generateStorageMocks({});
       const userIdService = new UserIdService(getStorageMock);
-      expect(Buffer.from(await userIdService.getId(), 'hex')).toHaveLength(USER_ID_BYTE_SIZE);
+      expect(Buffer.from(await userIdService.getRandomizedUserId(), 'hex')).toHaveLength(USER_ID_BYTE_SIZE);
     });
   });
 
@@ -47,7 +47,7 @@ describe('userIdService', () => {
       const { getStorageMock, setStorageMock } = generateStorageMocks({});
       const userIdService = new UserIdService(getStorageMock, setStorageMock);
       await userIdService.makePersistent();
-      const temporaryUserId = await userIdService.getId();
+      const temporaryUserId = await userIdService.getRandomizedUserId();
       expect(setStorageMock).toHaveBeenCalledWith({
         usePersistentUserId: true,
         userId: temporaryUserId
@@ -72,12 +72,12 @@ describe('userIdService', () => {
         userId: undefined
       });
       // it should still retain the previously stored user id in memory
-      expect(await userIdService.getId()).toEqual(store.userId);
+      expect(await userIdService.getRandomizedUserId()).toEqual(store.userId);
       // simulate a session timeout
       jest.advanceTimersByTime(SESSION_LENGTH + 1);
       // assert new temporary user id being generated after session timeout
-      expect(await userIdService.getId()).not.toEqual(store.userId);
-      expect(Buffer.from(await userIdService.getId(), 'hex')).toHaveLength(USER_ID_BYTE_SIZE);
+      expect(await userIdService.getRandomizedUserId()).not.toEqual(store.userId);
+      expect(Buffer.from(await userIdService.getRandomizedUserId(), 'hex')).toHaveLength(USER_ID_BYTE_SIZE);
     });
   });
 
@@ -96,18 +96,18 @@ describe('userIdService', () => {
       // simulate an almost session timeout
       jest.advanceTimersByTime(SESSION_LENGTH - 1);
       // it should still retain the previously stored user id in memory
-      expect(await userIdService.getId()).toEqual(store.userId);
+      expect(await userIdService.getRandomizedUserId()).toEqual(store.userId);
 
       await userIdService.extendLifespan();
       // simulate an almost session timeout
       jest.advanceTimersByTime(SESSION_LENGTH - 1);
-      expect(await userIdService.getId()).toEqual(store.userId);
+      expect(await userIdService.getRandomizedUserId()).toEqual(store.userId);
 
       // simulate a session timeout
       jest.advanceTimersByTime(1);
       // assert new temporary user id being generated after session timeout
-      expect(await userIdService.getId()).not.toEqual(store.userId);
-      expect(Buffer.from(await userIdService.getId(), 'hex')).toHaveLength(USER_ID_BYTE_SIZE);
+      expect(await userIdService.getRandomizedUserId()).not.toEqual(store.userId);
+      expect(Buffer.from(await userIdService.getRandomizedUserId(), 'hex')).toHaveLength(USER_ID_BYTE_SIZE);
     });
   });
 
@@ -120,33 +120,34 @@ describe('userIdService', () => {
     });
   });
 
-  describe('getting hash ID', () => {
-    it('should generate hash ID if keyAgentsByChain is defined', async () => {
+  describe('getting wallet based user ID', () => {
+    it('should generate hash ID if keyAgentsByChain is defined and is opted in user', async () => {
       const { getStorageMock, setStorageMock, clearStorageMock } = generateStorageMocks({
-        keyAgentsByChain: mockKeyAgentsByChain
+        keyAgentsByChain: mockKeyAgentsByChain,
+        usePersistentUserId: true
       });
       const userIdService = new UserIdService(getStorageMock, setStorageMock, clearStorageMock);
-      const hashId = await userIdService.getHashId('Preview');
-      expect(hashId).toEqual(mockHashId);
+      const walletBasedId = await userIdService.getUserId(1);
+      expect(walletBasedId).toEqual(mockWalletBasedId);
       expect(mockBlake2b).toHaveBeenCalledTimes(2);
     });
 
     it('should not generate hash ID twice - should retrieve hash from memory', async () => {
       mockBlake2b.mockClear();
       const { getStorageMock, setStorageMock, clearStorageMock } = generateStorageMocks({
-        keyAgentsByChain: mockKeyAgentsByChain
+        keyAgentsByChain: mockKeyAgentsByChain,
+        usePersistentUserId: true
       });
       const userIdService = new UserIdService(getStorageMock, setStorageMock, clearStorageMock);
-      await userIdService.getHashId('Preview');
-      const hashId = await userIdService.getHashId('Preview');
-      expect(hashId).toEqual(mockHashId);
-      expect(mockBlake2b).toHaveBeenCalledTimes(2);
+      const walletBasedId = await userIdService.getUserId(1);
+      expect(walletBasedId).toEqual(mockWalletBasedId);
+      expect(mockBlake2b).not.toHaveBeenCalled();
     });
 
     it('should not generate hash ID if keyAgentsByChain is not defined', async () => {
       const { getStorageMock, setStorageMock, clearStorageMock } = generateStorageMocks({});
       const userIdService = new UserIdService(getStorageMock, setStorageMock, clearStorageMock);
-      expect(await userIdService.getHashId('Preview')).toBeUndefined();
+      expect(await userIdService.getRandomizedUserId()).toBeUndefined();
     });
   });
 });
--- a/apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
+++ b/apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
@@ -1,13 +1,10 @@
-/* eslint-disable no-magic-numbers */
-/* eslint-disable import/imports-first */
-/* eslint-disable prettier/prettier */
-const mockWalletBasedId =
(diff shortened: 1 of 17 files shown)


apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -54,6 +83,7 @@ export class UserIdService implements UserIdServiceInterface {
   async makeTemporary(): Promise<void> {
     console.debug('[ANALYTICS] Converting user ID into temporary');
     await this.setStorage({ usePersistentUserId: false, userId: undefined });
+    this.userId = undefined;
```
jplorek-atix, line 86: If we not remove the user ID from memory, when the user opt out, we will keep sending events with the ID that has already been linked to the permanent user ID, since we are using the userId as alias when the user opt in

Would it be better to have field for alias ID or is it ok to keep using the userId field?

marcin-mazurek: When the user opts out, for the duration of that session, same ID should be still used. When the session ends and another one starts, another ID must be used.

We can have separate fields for randomUserId and walletBasedUserId, it would improve readability

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -21,6 +28,29 @@ export class UserIdService implements UserIdServiceInterface {
...
+    const keyAgentsByChain = (await this.getStorage())?.keyAgentsByChain;
+
+    if (!keyAgentsByChain) {
+      console.debug('[ANALYTICS] Key agents not found - Wallet not created yet');
+      return this.hashId;
+    }
+
+    if (!this.hashId) {
+      console.debug('[ANALYTICS] Hash ID not found - generating new one');
+      const hash = hashExtendedAccountPublicKey(keyAgentsByChain[chainName].keyAgentData.extendedAccountPublicKey);
```
marcin-mazurek, line 47: Should we leave a comment indicating we're calling this function twice by design?

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -21,6 +28,29 @@ export class UserIdService implements UserIdServiceInterface {
     private sessionLength: number = SESSION_LENGTH
   ) {}
 
+  async getIsPersistentId(): Promise<boolean> {
+    const { usePersistentUserId } = await this.getStorage();
+    return usePersistentUserId;
+  }
+
+  async getHashId(chainName: Wallet.ChainName): Promise<string> {
```
marcin-mazurek, line 36: How about:
```suggestion
  async getWalletBasedUserId(chainName: Wallet.ChainName): Promise<string> {
```

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -11,6 +17,7 @@ export const USER_ID_BYTE_SIZE = 8;
 
 export class UserIdService implements UserIdServiceInterface {
   private userId?: string;
+  private hashId?: string;
```
marcin-mazurek, line 20: How about:
```suggestion
  private walletBasedUserId?: string;
```

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -21,6 +28,29 @@ export class UserIdService implements UserIdServiceInterface {
     private sessionLength: number = SESSION_LENGTH
   ) {}
 
+  async getIsPersistentId(): Promise<boolean> {
```
marcin-mazurek, line 31: How about just:
```suggestion
  async isPersistent(): Promise<boolean> {
```

apps/browser-extension-wallet/src/providers/AnalyticsProvider/postHog/PostHogClient.ts
```diff
@@ -74,11 +86,29 @@ export class PostHogClient {
...
+    // Gets temporary user id, this will be used as alias for opted in user, for opted out user keeps as distinct_id
+    const userId = await this.userIdService.getId();
+    // Checks if wallet has been created
+    const hasTheWalletBeenCreated = !!hashId;
+    const isOptedInUserWithHashId = hasTheWalletBeenCreated && isPersistentId;
+    const idsMetadata = isOptedInUserWithHashId
+      ? {
+          distinct_id: hashId,
+          alias_id: userId
+        }
```
marcin-mazurek, line 102: Just checking - do we need to send both of them? I thought we could use the new ID under `distinct_id` property as soon as it's aliased with `posthog.alias()`.

In general - it would be nice if we could encapsulate this whole logic inside of UserIdService to decouple PostHogClient from user ID management

marcin-mazurek (changes requested):

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -31,6 +31,8 @@ export class AnalyticsTracker {
     // eslint-disable-next-line unicorn/prefer-ternary
     if (status === EnhancedAnalyticsOptInStatus.OptedIn) {
       await this.userIdService?.makePersistent();
+      // At this point, the wallet has been created, so, we are able to link the temporary ID with the hash ID
```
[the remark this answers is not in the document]
jplorek-atix: yes, you are right, I removed this comment, but, I left the `sendAliasEvent` call here, because we are checking if the hash id exist before sending the alias event, so, in onboarding flow, is not going to be sent, it is only going to be sent if the status change from setting section

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
```diff
@@ -1,7 +1,17 @@
+/* eslint-disable no-magic-numbers */
+/* eslint-disable import/imports-first */
```
DominikGuzei, line 2: let's move this const below the imports to get rid of this ignore rule?

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
```diff
@@ -103,4 +113,134 @@ describe('userIdService', () => {
...
+        keyAgentsByChain: mockKeyAgentsByChain,
+        usePersistentUserId: true
+      });
+      const userIdService = new UserIdService(getStorageMock, setStorageMock, clearStorageMock);
+      const walletBasedId = await userIdService.getwalletBasedUserId(1);
+      expect(walletBasedId).toEqual(mockWalletBasedId);
+      expect(mockHashExtendedAccountPublicKey).toHaveBeenCalledTimes(2);
+    });
+
+    it('should not run hashExtendedAccountPublicKey, should retrieve ID from memory', async () => {
```
DominikGuzei, line 131: small nitpick for test description: `'should not run hashExtendedAccountPublicKey but retrieve ID from memory'`

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
```diff
@@ -103,4 +113,134 @@ describe('userIdService', () => {
...
+      // call getwalletBasedUserId to get wallet based ID from memory
+      const walletBasedId = await userIdService.getwalletBasedUserId(1);
+      expect(walletBasedId).toEqual(mockWalletBasedId);
+      expect(mockHashExtendedAccountPublicKey).not.toHaveBeenCalled();
+    });
+
+    it('should not generate wallet based ID if keyAgentsByChain is not defined', async () => {
+      const { getStorageMock, setStorageMock, clearStorageMock } = generateStorageMocks();
+      const userIdService = new UserIdService(getStorageMock, setStorageMock, clearStorageMock);
+      const walletBasedId = await userIdService.getwalletBasedUserId(1);
```
DominikGuzei, line 154: hm, shouldn't we throw an exception in this case? Is there any valid use case where `keyAgentsByChain` is not defined?

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -21,7 +29,30 @@ export class UserIdService implements UserIdServiceInterface {
...
+      const chainName = getChainNameByNetworkMagic(networkMagic);
+      const extendedAccountPublicKey = keyAgentsByChain[chainName].keyAgentData.extendedAccountPublicKey;
+      this.walletBasedUserId = this.generateWalletBasedUserId(extendedAccountPublicKey);
+    }
+
+    console.debug(`[ANALYTICS] getwalletBasedUserId() called (current Wallet Based ID: ${this.walletBasedUserId})`);
+    return this.walletBasedUserId;
+  }
+
+  async getRandomizedUserId(): Promise<string> {
```
DominikGuzei, line 55: I'm not sure if `getwalletBasedUserId` and `getRandomizedUserId` should be public API interface? To me it seems like internal implementation detail and if we offer three different methods to get an ID it can easily be used in the wrong way? To me it seems like we should only expose `getUserId` and how that's created is internal logic of the class.

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
marcin-mazurek: +1

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.test.ts
jplorek-atix: in onboarding flow until we don't reach the last step `keyAgentsByChain` is not defined since the wallet has not been created yet

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -1,16 +1,24 @@
...
 import { USER_ID_SERVICE_BASE_CHANNEL, UserIdService as UserIdServiceInterface } from '@lib/scripts/types';
 import randomBytes from 'randombytes';
+import { getChainNameByNetworkMagic } from '@src/utils/chain';
 
 // eslint-disable-next-line no-magic-numbers
 export const SESSION_LENGTH = Number(process.env.SESSION_LENGTH_IN_SECONDS || 1800) * 1000;
 export const USER_ID_BYTE_SIZE = 8;
 
 export class UserIdService implements UserIdServiceInterface {
   private userId?: string;
```
marcin-mazurek, line 20:
```suggestion
  private randomizedUserId?: string;
```

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
```diff
@@ -21,7 +29,30 @@ export class UserIdService implements UserIdServiceInterface {
     private sessionLength: number = SESSION_LENGTH
   ) {}
 
-  async getId(): Promise<string> {
+  async getwalletBasedUserId(networkMagic: Wallet.Cardano.NetworkMagic): Promise<string | undefined> {
+    const { keyAgentsByChain, usePersistentUserId } = await this.getStorage();
+
+    if (!keyAgentsByChain) {
+      console.debug('[ANALYTICS] Key agents not found - Wallet not created yet');
+      return this.walletBasedUserId;
```
marcin-mazurek, line 37: Could we explicitly return undefined instead (also in line 42)? 
```suggestion
      return undefined;
```

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
jplorek-atix: @DominikGuzei I agree with you, I would probably leave `getRandomizedUserId` public for now since is used in Matomo and the functionality of wallet based ID is not applicable for there.

or maybe we can add some extra check in `getUserId` to find out if it was call from Matomo. What do you think?

marcin-mazurek (changes requested):

apps/browser-extension-wallet/src/lib/scripts/background/services/userIdService.ts
marcin-mazurek: Yeah, in this case, for now we should leave `getRandomizedUserId()` public. We can make it private when we get rid of Matomo.

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -40,6 +42,10 @@ export class AnalyticsTracker {
     await this.postHogClient?.sendPageNavigationEvent();
   }
 
+  async #sendAliasEvent(): Promise<void> {
+    await this.postHogClient?.sendAliasEvent();
```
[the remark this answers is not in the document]
jplorek-atix: @marcin-mazurek I ended up making this public, as I'm using it in [WalletSetupWizard.tsx](https://github.com/input-output-hk/lace/pull/304/files#diff-cdc8ed49ef37bc5910fcc1587654a09bcc942f30e675bbafabd28028da9ce683R247) when the user clicks `go to my wallet` button

DominikGuzei (approved):
LGTM 🚢  great work 🥇

apps/browser-extension-wallet/src/providers/AnalyticsProvider/postHog/PostHogClient.ts
```diff
@@ -49,6 +48,17 @@ export class PostHogClient {
...
   }
 
+  async sendAliasEvent(): Promise<void> {
+    const { id, alias } = await this.userIdService.getAliasProperties(this.chain.networkMagic);
+    // If one of this does not exist, should not send the alias event
+    if (!alias || !id) {
+      console.debug('[ANALYTICS] IDs were not found');
+      return;
+    }
+    console.debug('[ANALYTICS] Linking temporary ID with permanent user ID');
```
marcin-mazurek, line 58:
```suggestion
    console.debug('[ANALYTICS] Linking randomized ID with wallet-based ID');
```

marcin-mazurek (approved):
Well done 🚀 🚢 
Just one tiny suggestion to improve the logging message :)

milagi (approved):

marcin-mazurek:
@jplorek-atix can you resolve merge conflicts and merge? 🙏

PR #334, opened 2023-07-29
fix(extension): save discovered addresses into local storage

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-7693)
- [ ] Screenshots added.

---

## Proposed solution

In order to fix the issue there are in general two changes:
 - listening on the addresses change
 - blocking the UI until addresses received

There was a fair point from @mirceahasegan that the new addresses might be created as a result of some user flows .e.g multi-staking. Because of that we need to expect new addresses to arrive not only on the app initialisation flows (wallet creation/restoration or load from keyAgentData) but rather during the whole app life time.

Because of the above, the smartest way to fix the bug and have a future-proof implementation is to have a global listener for an addresses changes that will update:
 - the `knowAddresses` of the `keyAgentData` (local storage + `walletStore`)
 - the `knowAddresses` of the `keyAgentsByChain` (background storage)
 - the `keyAgent` object of the cardanoWallet object (`walletStore`)

Additionally in order to prevent the ui from starting without address discovery being complete the addresses listener sets a flag `addressesDiscoveryCompleted` which is require to render the lace.

That logic applies for all of the entry points (popup, browser view, dapp connector), so there is a new hook called `useAppInit()` containing the common logic and used by all of the entry points.

### Problem with locking the wallet
TLDR by locking the wallet we forget all the addresses that were discovered so it has to happen again.

There is yet another place having the stale `keyAgent`'s `knownAddresses` which is the `lock` object. That object is an encrypted `keyAgentsByChain` object and it gets created and stored during initial wallet creation. The problem is that when the user unlocks the wallet we decrypt the lock object and store the `keyAgentsByChain` and `keyAgentData` objects that does not have the updated `knownAddresses` property. We could update the `lock` object but we need user password for that.

## Testing

**Scenario 1: Already restored wallet which is currently broken gets fixed** (Repeat for each of network)
1. Have a wallet restored with the version from main
2. Open browser dev console->Application->localStorage->keyAgentData and ensure the `knownAddresses` property is an empty array
3. Stay with the dev tools localStorage open
4. Build lace from this branch
5. Refresh UI

Expected:
1. the `knownAddresses` property of na `keyAgentData` is an array with at least one property
2. the Spinner disappears after `knownAddresses` gets populated.

**Scenario 2: Restored/created wallet is correct from the beginning**
1. Have a lace with no wallet restored
2. Open lace
3. Open browser dev console->Application->localStorage
4. Restore or create wallet

Expected:
1. the `knownAddresses` property of na `keyAgentData` is an array with at least one property
2. the Spinner disappears after `knownAddresses` gets populated.

**Scenario 3: Unlocking the wallet** (Repeat for each of network)
1. Have a wallet restored with already discovered known addresses
2. Open UI
3. Open browser dev console->Application->localStorage
4. Lock the wallet
5. Unlock the wallet

Expected:
1. the `knownAddresses` property of na `keyAgentData` is an array with at least one property
2. the Spinner disappears after `knownAddresses` gets populated.

**Scenario 4: All three UIs doesn't hang on the initial loader**
Just check popup, browser view and dapp connector if it properly loads the UI (does not hang on the spinner)

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1313/5753463310/index.html) for [f1cf8215](https://github.com/input-output-hk/lace/pull/334/commits/f1cf8215c2de20e9341e8e15dfae9c1a2543d389)
| Total | 34     | 2      | 0       | 0     | 36    | ❌     |


apps/browser-extension-wallet/src/hooks/useAppInit.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    const walletManager = new WalletManagerUi({ walletName: process.env.WALLET_NAME }, { logger: console, runtime });
+    setWalletManagerUi(walletManager);
+  }, [setWalletManagerUi]);
+
+  useEffect(() => {
+    if (!walletManagerUi?.wallet || subscription.current) return () => void 0;
+
+    subscription.current = walletManagerUi?.wallet.addresses$.subscribe((knownAddresses) => {
```
szymonmaslowski, line 21: We can use `walletManagerUi.wallet` object to create a subscription as this object is available all the time, it doesn't get switched with any other wallet abject, it maintains its streams no matter what is happening on the UI.

apps/browser-extension-wallet/src/hooks/useAppInit.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useWalletStore } from '@stores';
+import { getValueFromLocalStorage } from '@utils/local-storage';
+import { useEffect } from 'react';
+import { runtime } from 'webextension-polyfill';
+
+export const useAppInit = (): void => {
+  const { setAddressesDiscoveryCompleted, setWalletManagerUi, walletManagerUi } = useWalletStore();
+  const { updateKeyAgentData } = useWalletManager();
+
+  useEffect(() => {
```
szymonmaslowski, line 13: This useEffect repeats for all of the UIs (popup, browser view, dapp connector)

apps/browser-extension-wallet/src/hooks/useAppInit.ts
```diff
@@ -0,0 +1,36 @@
...
+    const walletManager = new WalletManagerUi({ walletName: process.env.WALLET_NAME }, { logger: console, runtime });
+    setWalletManagerUi(walletManager);
+  }, [setWalletManagerUi]);
+
+  useEffect(() => {
+    if (!walletManagerUi?.wallet) return () => void 0;
+
+    const subscription = walletManagerUi.wallet.addresses$.subscribe(async (knownAddresses) => {
```
szymonmaslowski, line 21: We can use `walletManagerUi.wallet` object to create a subscription as this object is available all the time, it doesn't get switched with any other wallet abject, it maintains its streams no matter what is happening on the UI.

mchappell (approved):

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
```diff
@@ -357,21 +376,37 @@ export const useWalletManager = (): UseWalletManager => {
...
+  const updateKeyAgent = useCallback(
     async (newKeyAgentData: Wallet.KeyManagement.SerializableKeyAgentData) => {
       const backgroundStorage = await backgroundService.getBackgroundStorage();
+      const { name } = getValueFromLocalStorage('wallet');
 
       if (!backgroundStorage) throw new Error("Couldn't access background storage");
       const { keyAgentsByChain } = backgroundStorage;
 
+      keyAgentsByChain[environmentName].keyAgentData = newKeyAgentData;
+      // TODO: updateWalletLock so after unlocking the discovery does not get triggered
```
mirceahasegan: After a discussion with @szymonmaslowski, we came up with a few thoughts on wallet-lock and known addresses:
1. In the short run, discovery on wallet unlock might be acceptable if it also blocks the UI until it is complete, similar to the restore wallet flow.
2. We should also consider the reason behind erasing known addresses and storing them encrypted on wallet lock. Known addresses may be ok to store even if the wallet is locked, considering they are already public info. We could analyze what other data we're storing unencrypted on wallet-lock to make this decision. This is for future PR/improvement.

mirceahasegan (commented):
LGTM! 🚀

mrcnk (approved):

oldGreg5 (approved):

PR #355, opened 2023-08-03
[LW-7570]: update handle when sending tokens

shawnbusuttil:
# Checklist

- [ ] JIRA - LW-7570
- [ ] Screenshots added.

---

## Proposed solution

## Testing

1. Add one of your own handles to the address book and send it to another address (can skip if a previously saved handle has changed hands).
2. Go to send flow and proceed to add this handle from the address book.
3. Expected: A warning with a review option should appear post-verification. 
4. Click the call to action and you should be taken to an update/delete address flow.
5. Take an action - Update or Delete.
6. Expected: You should be taken back to the address input screen.

## Screenshots

https://github.com/input-output-hk/lace/assets/27725122/41923d92-4bec-469f-bd68-77f1cec88cbe

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1345/5783623371/index.html) for [4b46e913](https://github.com/input-output-hk/lace/pull/355/commits/4b46e9138890373dda973194632b7261013f5699)
| Total | 35     | 1      | 0       | 0     | 36    | ❌     |


apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.tsx
```diff
@@ -0,0 +1,158 @@
...
+  ({ isPopupView }: { isPopupView?: boolean }): React.ReactElement => {
+    const { t } = useTranslation();
+    const [actualAddressHasBeenCopied, setActualAddressHasBeenCopied] = useState(false);
+    const [expectedAddressHasBeenCopied, setExpectedAddressHasBeenCopied] = useState(false);
+    const handleResolver = useHandleResolver();
+    const { list: addressList } = useAddressBookContext();
+    const validatedAddressStatus = useUpdateAddressStatus(addressList as AddressBookSchema[], handleResolver);
+    const { addressToEdit } = useAddressBookStore();
+
+    console.log('addr2::', addressToEdit);
```
greatertomi, line 30: Remove this `console`

apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.tsx
```diff
@@ -0,0 +1,158 @@
...
+    };
+    const handleMouseLeaveActualAddress = () => {
+      setActualAddressHasBeenCopied(false);
+    };
+
+    const expectedAddressCopyText = expectedAddressHasBeenCopied ? 'general.button.copied' : 'general.button.copy';
+    const actualAddressCopyText = actualAddressHasBeenCopied ? 'general.button.copied' : 'general.button.copy';
+
+    return (
+      <>
```
greatertomi, line 65: This is an extra unnecessary fragment

apps/browser-extension-wallet/src/features/send/components/Send.tsx
```diff
@@ -64,7 +65,15 @@ export const Send = (): React.ReactElement => {
             <HeaderTitle popup />
           </div>
         }
-        footer={shouldDisplayFooter ? <Footer isPopupView openContinueDialog={toggleContinueDialog} /> : undefined}
+        footer={
+          shouldDisplayFooter ? (
```
greatertomi, line 69:
```suggestion
          shouldDisplayFooter && (
```

apps/browser-extension-wallet/src/hooks/useSendEvent.ts
```diff
@@ -0,0 +1,20 @@
+import { useCallback } from 'react';
+import { useAnalyticsContext } from '@providers';
+import { AnalyticsEventActions, AnalyticsEventCategories } from '@providers/AnalyticsProvider/analyticsTracker';
+
+type SendEventFn = (name: string, value?: number) => void;
+
+export const useSendEvent = (): SendEventFn => {
```
greatertomi, line 7: This hook was not used? Can it be deleted?

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/Form/AddressInput.tsx
```diff
@@ -195,6 +194,7 @@ export const AddressInput = ({ row, currentNetwork, isPopupView }: AddressInputP
   const isAddressInputValueValid = validationObject.name || validationObject.address;
 
   useEffect(() => {
+    // todo: debounce this
```
greatertomi, line 197: Why not just debounce this in this PR?

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransactionLayout.tsx
```diff
@@ -59,11 +60,17 @@ export const SendTransactionLayout = ({ children, isPopupView }: SendTransaction
         shouldAssetPickerDisplayFooter
           ? {
               // eslint-disable-next-line @typescript-eslint/no-empty-function
```
greatertomi, line 62:
```suggestion
```

greatertomi (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/Form/AddressInput.tsx
shawnbusuttil: This was not me.

apps/browser-extension-wallet/src/hooks/useSendEvent.ts
shawnbusuttil: Ah yes. I committed this by mistake but was going to do some refactoring which uses it.

apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressChangeDetailDrawer.tsx
```diff
@@ -142,144 +44,12 @@ export const AddressChangeDetailDrawer = ({
...
-                        </Text>
-                      </div>
-                    </div>
-                  </CopyToClipboard>
-                </div>
-              </Timeline.Item>
-            </Timeline>
-          </div>
-        </>
+        <AddressChangeDetail isPopupView={popupView} />
```
VanessaPC, line 51: Why aren't we passing `initialValues` here to the `AddressChangeDetail` component? those contain the expected and actual address, so no calculations need to be done with `useMemo` at least when it's coming from the `AddressBook`. I think there should be a way to send the same details from the SendFlow here, because we have the same details being calculated in the previous step, now we would be doing them twice, and the UI looks slower while we are doing the calculations because it shows blank. 
If the only reason to keep this component now is to use a Drawer, we should probably delete it and move the functionality to `AddressDetailDrawer` and do the needed logic for the couple of calls on the buttons there.

apps/browser-extension-wallet/src/hooks/useSendEvent.ts
VanessaPC: If we are using a hook for analytics, we probably want to be able to send the `action, category` props, so it can be used elsewhere.

apps/browser-extension-wallet/src/utils/validators/address-book.ts
```diff
@@ -108,9 +108,9 @@ type ensureHandleOwnerHasntChangedArgs = {
 export const ensureHandleOwnerHasntChanged = async ({
   handleResolution,
   handleResolver
-}: ensureHandleOwnerHasntChangedArgs): Promise<void> => {
+}: ensureHandleOwnerHasntChangedArgs): Promise<boolean> => {
```
VanessaPC, line 111: Why are we returning now a boolean? this logic hasn't changed anywhere else where it's being used and it's not used anywhere new, so can we leave it as it was before?

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/Form/AddressInput.tsx
```diff
@@ -233,6 +233,19 @@ export const AddressInput = ({ row, currentNetwork, isPopupView }: AddressInputP
     setSection(sectionsConfig[section]);
   };
 
+  const handleAddressReview = () => {
+    setAddressToEdit({
+      name: addressInputValue.name,
+      address: handle,
+      handleResolution: addressInputValue.handleResolution
+    });
+    setSection({ currentSection: Sections.ADDRESS_CHANGE, prevSection: Sections.FORM });
```
VanessaPC, line 242: can we pass to this section the `actual` and `expected` addressed, in this way we wouldn't need further calculations one step after.

VanessaPC (changes requested):
My flags are turned on, but I can't see the button.  
Screenshot 2023-08-08 at 12 47 25

apps/browser-extension-wallet/src/hooks/useSendEvent.ts
shawnbusuttil: Updated.

apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressChangeDetailDrawer.tsx
shawnbusuttil: Because since the `AddressChangeDetail` component is now being used in two places, and as you said, the send flow works differently, I thought it should be part of the common logic.

nix/lace-blockchain-services/internal/lace-blockchain-services/openapi.json
```diff
@@ -10,9 +10,7 @@
       "url": "http://127.0.0.1:12345"
     }
   ],
-  "tags": [
-    { "name": "default" }
-  ],
+  "tags": [{ "name": "default" }],
```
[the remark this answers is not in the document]
shawnbusuttil: This came from main? Reverting anyway.

apps/browser-extension-wallet/src/utils/validators/address-book.ts
shawnbusuttil: Reverting. This was also part of conflict resolution I think.

apps/browser-extension-wallet/src/features/address-book/components/AddressDetailDrawer/AddressChangeDetailDrawer.tsx
VanessaPC: That's fine. But then `AddressChangeDetailDrawer` is now a redundancy of drawers... because it's only used to present this part that contains the actual UI and logic for the job. So either we can clean it up here or open a ticket for more cleanup.

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/Form/AddressInput.tsx
VanessaPC: I thought we had passed a prop like `name` somewhere else.

apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.module.scss
```diff
@@ -0,0 +1,110 @@
...
+  
+  .footer {
+    gap: size_unit(1);
+  }
+}
+
+.copyAddressButton {
+  width: 100%;
+  font-weight: $font-weight-bold !important;
+}
```
greatertomi, line 96: This style is not used too

apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.module.scss
```diff
@@ -0,0 +1,110 @@
...
+
+.popupView {
+  .addressWrapper {
+    flex-direction: column;
+  }
+  
+  .footer {
+    gap: size_unit(1);
+  }
+}
```
greatertomi, line 91: Couldn't find where this is used too

apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.module.scss
```diff
@@ -0,0 +1,110 @@
...
+
+.footer {
+  display: flex;
+  flex-direction: column;
+  gap: size_unit(2);
+  
+  button {
+    font-weight: $font-weight-bold !important;
+  }
+}
```
greatertomi, line 110: This too. Couldn't find where it was used.

greatertomi (changes requested):

apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.module.scss
```diff
@@ -0,0 +1,110 @@
...
+  .addressDetails {
+    color: var(--text-color-secondary);
+    @include text-body-medium;
+    letter-spacing: -0.015em;
+  }
+}
+
+.sideTimeline {
+  margin: 0 !important;
+}
```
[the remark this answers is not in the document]
shawnbusuttil: It is used. I specifically put it there to override the one coming from the Timeline module. Because it was rendering with a big space to the left.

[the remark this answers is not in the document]
VanessaPC: @shawnbusuttil  good catch! do we want to log this as a bug? this space might be affecting the other place where it's used. cd

apps/browser-extension-wallet/src/features/address-book/components/AddressBook.tsx
```diff
@@ -151,17 +153,10 @@ export const AddressBook = withAddressBookContext(() => {
           onCancelClick={() => {
             setAddressToEdit({} as AddressBookSchema);
             setIsAddressDrawerOpen(false);
+            setSection({ currentSection: Sections.FORM });
           }}
           initialValues={addressToEdit}
-          expectedAddress={validatedAddressStatus[addressToEdit.address]?.error?.expectedAddress ?? ''}
```
VanessaPC, line 156: This expected And actual address where here so that we wouldn't be doing calculations in the drawers. They are smaller components that just render and pass values to.. they don't need to be doing calculations. 
The large calculations should be done in the parent components where can't extract them from. 
So this was correct at this level. 
If we do this here, we need to get the same address from the Send flow, but on the send flow we are calculating already the handle verification so at that point we also already have an actual and expected address to send to the drawer.

apps/browser-extension-wallet/src/hooks/useValidateAddressStatus.tsx
```diff
@@ -0,0 +1,27 @@
...
+  expectedAddress: string;
+  actualAddress: string;
+};
+
+export const useValidateAddressStatus = (address: string): AddressStatusProps => {
+  const handleResolver = useHandleResolver();
+  const { list: addressList } = useAddressBookContext();
+  const validatedAddressStatus = useUpdateAddressStatus(addressList as AddressBookSchema[], handleResolver);
+
+  const expectedAddress = useMemo(
```
VanessaPC, line 17: I don't understand the use of `useMemo`, this should only be used when we need to allocate memory aside for a large calculation, extracting data from the store isn't one. Let's lighten this.

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransaction.tsx
```diff
@@ -15,46 +15,62 @@ import { AssetPicker } from './AssetPicker';
...
+      [Sections.SUCCESS_TX]: <TransactionSuccess />,
+      [Sections.FAIL_TX]: <TransactionFail />,
+      [Sections.ADDRESS_LIST]: <AddressList isPopupView={isPopupView} scrollableTargetId={scrollableTargetId} />,
+      [Sections.ADDRESS_FORM]: <AddressForm isPopupView={isPopupView} />,
+      [Sections.ASSET_PICKER]: <AssetPicker isPopupView={isPopupView} />,
+      [Sections.ADDRESS_CHANGE]: (
+        <AddressChangeDetail
+          name={name}
+          isPopupView={isPopupView}
+          expectedAddress={expectedAddress}
```
VanessaPC, line 66: if we have this here, why do we have to then calculate it in the drawer? can we find a way to send it through? else in this case I would leave it.

VanessaPC (changes requested):

apps/browser-extension-wallet/src/features/address-book/components/AddressChangeDetail.module.scss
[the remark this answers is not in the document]
shawnbusuttil: The only other place is the onboarding flow right?

apps/browser-extension-wallet/src/features/address-book/components/AddressBook.tsx
shawnbusuttil: After my change, the component now shows the addresses immediately. Are you saying that we should pass them two levels down from the parent?

apps/browser-extension-wallet/src/hooks/useValidateAddressStatus.tsx
shawnbusuttil: Will remove this.

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransaction.tsx
shawnbusuttil: There is no drawer in this case. There are only component mapping to each section.

apps/browser-extension-wallet/src/features/address-book/components/AddressBook.tsx
VanessaPC: I am saying that one of the levels is now not needed, because that drawer does no calculations at all and its props are almost the same as the other drawer because the Timeline part of it has been abstracted. So, I think we should calculate it here, and pass it down to the drawer. The drawer has no reason to do a calculation.

greatertomi (approved):

VanessaPC:
I am testing your branch, but can't send an NFT, all the flags are enabled, and the app is rebuilt and restarted, but this is what I get, though my balance is 3000 ish tADA. 
Screenshot 2023-08-15 at 13 13 41

VanessaPC (approved):
Great work!

dmitrii-gaico (changes requested):
@shawnbusuttil Hey!
Could you please take a look why it is showing the same Address for Previous/New while Send flow. If reviewing the Handle from the Address Book then it works fine.
Address Book:
image
While Send Flow:
image

dmitrii-gaico (approved):

mchappell (approved):

PR #402, opened 2023-08-15
feat(staking): LW-7864 add migrating stake pools banner

refi93:
# Checklist

- [x] JIRA - \ [LW-7864](https://input-output.atlassian.net/browse/LW-7864)
- [x] Screenshots added.

---

## Motivation

Having already some delegation or starting a new one  user won’t  see immediately on the overview page untill the transaction gets submited which may confuse them.

## Proposed solution

While the delegation transaction is pending, we'd show to the user a banner letting them know the staking portfolio is about to change. There are two scenarios to consider:

1. User is already delegating, so the message is about migrating stake pools
2. User is not delegating, so this is their first delegation. In that case we show the same banner informing them that their staking portfolio will appear soon

## Testing

1. Test banner for scenario 1 by delegating from a wallet with no previous delegations, check that while the transaction is pending, in the "Staking" tab the screenshotted banner would appear.
2. Test banner for scenario 2 by picking some set of pools to delegate to, delegate to them and check that while the transaction is pending, the screenshotted banner would appear.

## Note

There is an edge-case if user delegates to the same portfolio they are currently delegating to, in which case the transaction is identified as a "self" transaction and therefore no banner is shown due to us listnening just for pending `delegation*` transaction types. So, arguably, this edge-case is handled by this PR reasonably (no real change to the portfolio, so no reason to show a banner about a pending "change"). Showing some other message to the user warning them about the redundancy of this action (delegating to the same portfolio) is part of this task in the backlog: https://input-output.atlassian.net/browse/LW-7641

## Screenshots
First pending delegation:
Screenshot 2023-08-15 at 22 50 19

Pending delegation for existing portfolio:
Screenshot 2023-08-15 at 22 49 42

[LW-7864]: https://input-output.atlassian.net/browse/LW-7864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1484/5871642164/index.html) for [4e8da942](https://github.com/input-output-hk/lace/pull/402/commits/4e8da942ba0f36e5f2760271ccc52628c6962a96)
| Total | 33     | 2      | 0       | 0     | 35    | ❌     |


apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
```diff
@@ -54,15 +58,60 @@ export const MultiDelegationStaking = (): JSX.Element => {
...
+        fiatCurrency,
+        cardanoFiatPrice: priceResult.cardano?.price,
+        sendAnalytics
+      }));
+    setWalletActivitiesObservable(result);
+  }, [fiatCurrency, getWalletActivitiesObservable, sendAnalytics, priceResult]);
+
+  useEffect(() => {
+    fetchWalletActivities();
+  }, [fetchWalletActivities]);
```
refi93, line 109: this code is already repeated three times in places related to the Activity and is apparently needed to trigger fetching of the activity data/be reactive to its changes. I considered refactoring/deduplicating it but it would touch areas we, as the purple squad don't own (AFAIK) so I concluded the most sensible for now is to repeat the code 1:1 and should anybody refactor this in the future, they would at least have an easier job by finding the impacted places by a simple string match in the codebase.

Places where this logic is already repeated: https://github.com/search?q=repo%3Ainput-output-hk%2Flace%20setWalletActivitiesObservable&type=code

packages/staking/src/features/outside-handles-provider/types.ts
```diff
@@ -89,6 +89,7 @@ export type OutsideHandlesContextValue = {
       priceVariationPercentage24h: number;
     };
   };
+  hasPendingDelegationTransaction?: boolean;
```
refi93, line 92: made this optional to not force the popup to load activity (from which this flag is inferred) which would probably look arbitrary there given it's not used and would just result in redundant network requests

apps/browser-extension-wallet/src/stores/slices/wallet-activities-slice.ts
```diff
@@ -49,7 +49,11 @@ interface FetchWalletActivitiesPropsWithSetter extends FetchWalletActivitiesProp
 }
 
 export type FetchWalletActivitiesReturn = Observable<Promise<AssetActivityListProps[]>>;
-const DelagatioinTransactionTypes = new Set(['delegation', 'delegationRegistration', 'delegationDeregistration']);
+export const DelegationTransactionTypes: Set<TransactionType> = new Set([
```
refi93, line 52: added explicit typing for better type-safety, i.e. to ensure TS would warn us if any of the types in the set isn't a valid `TransactionType`

packages/staking/src/features/overview/Overview.tsx
```diff
@@ -49,6 +66,16 @@ export const Overview = () => {
           status={currentPortfolio.length === 1 ? 'simple-delegation' : 'multi-delegation'}
         />
       </Box>
+      {hasPendingDelegationTransaction && (
+        <Box mb={'$40'}>
+          <Banner
+            withIcon
+            customIcon={<InfoCircleOutlined className={styles.infoIcon} />}
```
refi93, line 73: this need to style explicitly the info icon to be purple seems a bit cumbersome to me, but I didn't find any reusable "purple info" icon, did I perhaps miss something?

packages/staking/src/features/overview/Overview.tsx
```diff
@@ -25,7 +29,20 @@ export const Overview = () => {
     setIsDrawerVisible(true);
   };
 
-  if (currentPortfolio.length === 0) return <Text.SubHeading>Start staking</Text.SubHeading>;
+  if (currentPortfolio.length === 0)
+    return (
+      <>
+        <Text.SubHeading>Start staking</Text.SubHeading>
```
refi93, line 35: hide this if banner is shown

packages/staking/src/features/outside-handles-provider/types.ts
refi93: @przemyslaw-wlodek let's double-check if we want the banner in the popup view or not

refi93: checked offline with Przemek and popup is out of scope indeed

apps/browser-extension-wallet/src/views/browser-view/features/assets/components/AssetDetailsDrawer/AssetDetailsDrawer.tsx
```diff
@@ -58,29 +55,9 @@ const renderAssetDetails =
           : AnalyticsEventNames.ViewTokens.VIEW_TOKEN_TX_DETAILS_BROWSER
       });
     }, [analytics, popupView]);
-
-    const fetchWalletActivities = useCallback(async () => {
-      const result =
-        fiatCurrency &&
-        fiatPrice &&
```
refi93, line 65: @przemyslaw-wlodek @szymonmaslowski the deduplication of the wallet activities fetching logic seemed straightforward perhaps except this place where I had to decide between keeping the `fiatPrice` passed from outside as a parameter to this function vs retrieving it within the `useWalletActivities()` hook.

I finally went with the former as it seems more convenient for the caller and also considering that apparently the original code wasn't really consistent, because the `fiatCurrentcy` was obtained from the store vs the `fiatPrice` passed from outside, and the refactored `useWalletActivities` hook makes the source of these two (related) properties consistent

apps/browser-extension-wallet/src/hooks/useWalletActivities.ts
```diff
@@ -0,0 +1,53 @@
...
+import { useCallback, useEffect, useState } from 'react';
+import { useFetchCoinPrice } from './useFetchCoinPrice';
+import { StateStatus, useWalletStore } from '@src/stores';
+import { AssetActivityListProps } from '@lace/core';
+
+type UseWalletActivitiesProps = {
+  sendAnalytics: () => void;
+};
+
+export const useWalletActivities = ({
```
refi93, line 12: @przemyslaw-wlodek @szymonmaslowski I ended up deduplicating the wallet activity fetching logic into this hook. In the end it seemed quite straightforward but maybe it turns out to go against some practices/patterns in the codebase I'm not yet aware of. If this refactor turns out to "open a can of worms" don't hesitate to point it out and I can drop this commit

apps/browser-extension-wallet/src/views/browser-view/features/assets/components/AssetDetailsDrawer/AssetDetailsDrawer.tsx
refi93: what I'm unsure about is that this new `useWalletActivities()` hook basically introduces an additional way of accessing the activities and related data, so a bit of ambiguity, but I'd nevertheless consider it an improvement to the existing approach of repeating the fetching boilerplate pretty much anywhere the activity data was needed so far.

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
```diff
@@ -54,15 +57,35 @@ export const MultiDelegationStaking = (): JSX.Element => {
     fetchStakePools: state.fetchStakePools,
     networkInfo: state.networkInfo,
     fetchNetworkInfo: state.fetchNetworkInfo,
-    blockchainProvider: state.blockchainProvider
+    blockchainProvider: state.blockchainProvider,
+    getWalletActivitiesObservable: state.getWalletActivitiesObservable
```
szymonmaslowski, line 61: This line is redundant

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
```diff
@@ -54,15 +57,35 @@ export const MultiDelegationStaking = (): JSX.Element => {
...
   const { fiatCurrency } = useCurrencyStore();
   const { executeWithPassword } = useWalletManager();
   const [multidelegationFirstVisit, { updateLocalStorage: setMultidelegationFirstVisit }] = useLocalStorage(
     MULTIDELEGATION_FIRST_VISIT_LS_KEY,
     true
   );
 
+  const hasPendingDelegationTransaction = flatMap(walletActivities, ({ items }) => items).some(
+    ({ type, status }) => DelegationTransactionTypes.has(type) && status === 'sending'
+  );
```
szymonmaslowski, line 87: Let's move this logic to staking. The more we encapsulate inside, the better. Ideally Staking should be a "plug and play" component.

I would suggest creating a helper function in the overview directory called `hasPendingDelegationTransaction` and accepting `walletActivities`, so it could be reused by the `Overview` and `OverviewPopup` components.

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
```diff
@@ -77,6 +100,7 @@ export const MultiDelegationStaking = (): JSX.Element => {
         delegationStoreDelegationTxFee: delegationTxFee,
         delegationStoreSelectedStakePool: selectedStakePool,
         fetchCoinPricePriceResult: priceResult,
+        hasPendingDelegationTransaction,
```
szymonmaslowski, line 103: Moving the `hasPendingDelegationTransaction` logic to the staking we will need to pass the `walletActivities` instead. We need to do that for the `MultiDelegationStakingPopup` component as well.

packages/staking/src/features/overview/Overview.css.ts
```diff
@@ -0,0 +1,6 @@
+import { style } from '@vanilla-extract/css';
+import { theme } from '../theme';
+
+export const bannerInfoIcon = style({
+  color: `${theme.colors.$bannerInfoIconColor} !important`,
```
szymonmaslowski, line 5: Are we able to set the color other way than using important? Maybe increasing the specificity?

szymonmaslowski (changes requested):
Great Job 💪

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
przemyslaw-wlodek: I agree, but initially I discussed this with you and we decided to put it here. 😄 Now I see no harm in moving it (we will need to duplicate the `DelegationTransactionTypes` though). This comment is just to make it clear for @refi93 and @xdzurman.

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
refi93: The task still states that the popup view should have the banner, too, but @przemyslaw-wlodek confirmed with me offline that the popup view is left out of scope after discussing it with Fernando. The reason is that the user needs to be in the expanded view anyway to perform the pool migration. @przemyslaw-wlodek do you have a link to Fernando's comment or was it just a private conversation? Nevertheless, once I have the link to the comment, I'll update the task to prevent further confusion, sorry for that.

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
refi93: > I would suggest creating a helper function in the overview directory

I'd consider generalizing the `mapPortfolioToDisplayData.ts` file to `helpers.ts` and include there also the `hasPendingDelegationTransaction` helper, in order to not "spam" the overview folder with a new file for each helper, does that sound good to you?

packages/staking/src/features/overview/Overview.css.ts
refi93: actually, I just tried it without `!important` and it worked fine :D But I'm surprised because I remember trying the same before without success and as I've seen uses of "!important" in similar contexts, e.g. here: https://github.com/input-output-hk/lace/blob/d4801af6b7740f1d649f17324248b52d28ecc9a2/apps/browser-extension-wallet/src/features/dapp/components/Connect.module.scss#L145 I just went with putting there `!imporant` and it seemed to solve the problem. But apparently it works now, so I must have overlooked something subtle, so nevermind

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
refi93: done in https://github.com/input-output-hk/lace/pull/402/commits/f0b69bd90b45f748612093ff10d73afcb86a5db2

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
refi93: addressed in https://github.com/input-output-hk/lace/pull/402/commits/f0b69bd90b45f748612093ff10d73afcb86a5db2

packages/staking/src/features/overview/Overview.css.ts
refi93: done in https://github.com/input-output-hk/lace/commit/84f48d98d4ccca00c5d496794575fd120b72fa60

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
refi93: done in https://github.com/input-output-hk/lace/commit/84f48d98d4ccca00c5d496794575fd120b72fa60

apps/browser-extension-wallet/src/features/delegation/components/MultiDelegationStakingPopup.tsx
```diff
@@ -105,6 +121,7 @@ export const MultiDelegationStakingPopup = (): JSX.Element => {
         walletStoreFetchNetworkInfo: fetchNetworkInfo,
         walletStoreNetworkInfo: networkInfo,
         walletStoreBlockchainProvider: blockchainProvider,
+        walletStoreWalletActivities: walletActivities,
```
refi93, line 124: I checked the network requests in the popup and empirically, the popup was loading the wallet activities on load even before this change, so this doesn't result in any additional requests (which would be sad, because we currently don't use the activity for the popup as we don't show any banner there)

przemyslaw-wlodek (approved):
💪🏻 great work! Thank you @refi93 :)

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
szymonmaslowski: @refi93 TBH I would keep both in separate files as IMO it is a good practise (similar to having a single react compoennt per file). I am ok with having helpers file for now. We definitely should establish some guidelines and write them down 😉

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
szymonmaslowski: If it was about my preference I would definitely include the banner in the popup view even if we cannot stake there. That is just an info for the user so he is not surprised with what he sees there.

packages/staking/src/features/outside-handles-provider/types.ts
```diff
@@ -93,6 +94,7 @@ export type OutsideHandlesContextValue = {
   submittingState: SubmittingState;
   walletStoreGetKeyAgentType: () => string;
   walletStoreInMemoryWallet: Wallet.ObservableWallet;
+  walletStoreWalletActivities: AssetActivityListProps[];
```
szymonmaslowski, line 97: Our intention with those names was to include the data source in them, so e.g. `useWalletStore` hook exposes the `InMemoryWallet` object so the property name here is called `walletStoreInMemoryWallet`. Following this that property here should be called just wallet activities 😉 It's shit, I know 😄 It is there only temporary to help us understand all the data sources of the Staking once we have some more time.

That is not something for which we should block merging of the PR, I am just saying for an informative purpose 😉

szymonmaslowski (approved):
Well done!

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/MultiDelegationStaking.tsx
refi93: ok, I'll leave the "helpers" file for now, IMHO I wouldn't look at helper functions as "react components", components usually are "composite" structures (combining hooks, styles, ...) and hence usually have some boilerplate around which may look a bit confusing if mixed. But such pure functions are usually pretty simple and separating them into files may make the codebase just look more cluttered to the "naked eye", also if some pattern emerges in the helpers it is a bit harder to notice as they are scattered around, hence making it harder to group them nicely which would otherwise aid readability, but if we agree on the pattern of one function per file, that's fine by me, too, of course, but I'd rather align on that on a team level as you mentioned

ManusMcCole (approved):

refi93:
sorry I had to force-push this branch to revert a broken merge commit that was the result of merging this branch into a local version of `main` that contained additional commits I've been expreimenting with. The commit 2684f50 I force pushed was already approved in this PR

danielmain (approved):

PR #415, opened 2023-08-18
[LW-7486] refactor onAddressSave to a hook

VanessaPC:
# Checklist

- [x] JIRA - LW-7486
- [ ] Screenshots added.

---

## Proposed solution
Refactor some of the duplicated code to achieve less repetition. Add tests to the code. 

## Testing
Make sure that saving an address works from the address book and the send flow. 

## Screenshots

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1673/5960722789/index.html) for [c073cc35](https://github.com/input-output-hk/lace/pull/415/commits/c073cc35b04dc4243e9afe82b5039ec4c2ed336e)
| Total | 33     | 2      | 0       | 0     | 35    | ❌     |


apps/browser-extension-wallet/src/hooks/useOnAddressSave.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { t } = useTranslation();
+  const { utils } = useAddressBookContext();
+  const { saveRecord: saveAddress, updateRecord: updateAddress } = utils;
+  const handleResolver = useHandleResolver();
+
+  const onSaveAddressActions = async (
+    address: AddressBookSchema | Omit<AddressBookSchema, 'id' | 'network'> | Omit<AddressBookSchema, 'id'>,
+    addressToEdit: AddressBookSchema | Omit<AddressBookSchema, 'id' | 'network'>
+  ) => {
+    analytics.sendEventToMatomo({
```
greatertomi, line 33: Shouldn't there be `await` here?

greatertomi (approved):
Nice refactor. Well done. 💪🏽

apps/browser-extension-wallet/src/hooks/useOnAddressSave.ts
VanessaPC: I don't want to block saving the actual address saving or update actions, the analytics events will get sent either way, and we don't wait for any response because .. well, we don't do anything with it. 
The part we have to await is for the address to save.

apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressBook/AddressBook.tsx
```diff
@@ -27,8 +25,7 @@ import {
   AnalyticsEventNames
 } from '@providers/AnalyticsProvider/analyticsTracker';
 import { AddressDetailsSteps } from '@src/features/address-book/components/AddressDetailDrawer/types';
-import { useHandleResolver, useUpdateAddressStatus } from '@hooks';
-import { getAddressToSave } from '@src/utils/validators';
+import { useHandleResolver, useOnAddressSave, useUpdateAddressStatus } from '@hooks';
```
shawnbusuttil, line 28: Small favour: Do you think you can rename the directory to 'address-book'?

shawnbusuttil (approved):

apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressBook/AddressBook.tsx
VanessaPC: Yea, I will do that.

VanessaPC: Hey, I tried the renaming of this dir. here, but it was adding quite a bit of a diff. So, are you ok with me mentioning this in my following ticket which touches more files in the address-book in general?

iadmytro (approved):

gabriela-ponce:
@VanessaPC I found that when switching the network in the address book, you can see in some contacts that the alert about the address changing is displayed. Also you can click on it, but if you wait for a few second it would disappear (as it occurs when the handle didn't change the address). Check [this](https://drive.google.com/file/d/1c2VZyWmwex0JWT05MGja3Y1XXY_e_gDT/view?usp=drive_link) video.

VanessaPC:
> @VanessaPC I found that when switching the network in the address book, you can see in some contacts that the alert about the address changing is displayed. Also you can click on it, but if you wait for a few second it would disappear (as it occurs when the handle didn't change the address). Check [this](https://drive.google.com/file/d/1c2VZyWmwex0JWT05MGja3Y1XXY_e_gDT/view?usp=drive_link) video.

This is a really good find. This happens because when we switch. networks we aren't handling the in between status as it's handled on the dashboard. We should be clearing the state of the data in the app and likely add that sort of loading / blue screen Dashboard has, I believe, at least a clearing of the state. I created this ticket and appended your video LW-8177.

apps/browser-extension-wallet/src/views/browser-view/features/adress-book/components/AddressBook/AddressBook.tsx
VanessaPC: On a second thought, I don't think this directory should be modified, the reason being, two folders above we already have an `address-book/components/` folder, and the folder in here is within a component folder. React does say their components folders should be camel case, and I would avoid duplicating (or in this case triplicating) address-book, what do you think?

PR #361, opened 2023-08-04
feature(staking): lw-7093 multi staking popup view

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-7093)
- [ ] Proper tests implemented
- [ ] Screenshots added.

---

## Proposed solution

Staking popup view implementation according to designs: [figma](https://www.figma.com/file/C1g4ndWSv5u53GMlTeQ0nN/Lace-Browser---Pop-up-view-UI-Design?node-id=215%3A67546&mode=dev)

## Testing

Open popup and go to staking

## Screenshots

image
image
image
image
image

---
# Allure report
`allure-report-publisher` generated test report!

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1462/5852366385/index.html) for [2d34c344](https://github.com/input-output-hk/lace/pull/361/commits/2d34c34414f00a2f8e746663f7a916e7eb94f8bd)
|       | passed | failed | skipped | flaky | total | result |
|-------|--------|--------|---------|-------|-------|--------|
| Total | 34     | 1      | 0       | 0     | 35    | ❌     |

PR #416, opened 2023-08-18
feat(staking): LW-8017 disable multi-delegation for HW wallets

refi93:
# Checklist

- [x] JIRA - \ [LW-8017](https://input-output.atlassian.net/browse/LW-8017)
- [x] Screenshots added.

---

## Proposed solution

Replace the static check for environment variable with a `useIsMultidelegationEnabled()` hook that can retrieve the wallet key type (in memory / hw wallet) at runtime, allowing us to decide whether to show the legacy delegation UI (compatible with HW wallets) or the new multi-delegation UI

## Testing

* If `USE_MULTI_DELEGATION` env flag is set to `false`, there should be no perceivable change to the Staking section UI, the legacy UI should be shown
* If `USE_MULTI_DELEGATION` env flag is set to `true`, for "in memory" wallets the new multi-staking UI should be shown when visting the "Staking" section, both in the extended and popup view
* If `USE_MULTI_DELEGATION` env flag is set to `true`, for "hardware" wallets the user should see the legacy single staking view both in the extended and popup view, despite the feature flag being enabled
* check that staking interface has no apparent abnormalities in dark nor light mode (changed imports of styles in the PR which in theory may have impact)

## Screenshots

New multidelegation UI should be visible only to in-memory wallets:
Screenshot 2023-08-18 at 18 04 41

UI should fall back to single staking mode if using a hardware wallet:
Screenshot 2023-08-18 at 18 04 41

[LW-8017]: https://input-output.atlassian.net/browse/LW-8017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1553/5904725029/index.html) for [8c0a4fe1](https://github.com/input-output-hk/lace/pull/416/commits/8c0a4fe1208c9df9cc71ea465a4ffa07812de32b)


apps/browser-extension-wallet/src/views/browser-view/index.tsx
```diff
@@ -16,20 +16,17 @@ import '@lib/i18n';
 import 'antd/dist/antd.css';
 import '../../styles/index.scss';
 import 'normalize.css';
+// Disabling import/no-unresolved as it is not aware of the "exports" entry
+// https://github.com/import-js/eslint-plugin-import/issues/1810
+// eslint-disable-next-line import/no-unresolved
+import '@lace/staking/index.css';
```
refi93, line 22: @szymonmaslowski I wondered if this css may somehow interfere with the "single staking" UI and that's why the import was conditioned by the env var, but given that the file was directly imported up until recently without explicit mention of issues (changed in this PR: https://github.com/input-output-hk/lace/pull/361/files#diff-23dce96c0ea286e4c6d9f7a90d63d48f9c1621b46485870d940518cd745ce9d9L7) I'd assume this worked fine and the only reason was bundle size optimization - am I right? If so, the bigger bundle size seems like a fair tradeoff for me for solving the problem of hw wallet users - I didn't notice any app slowdown myself. Was there perhaps some other, stronger reason for the conditional import?

Moreover, the compiled css has apparently keys with unique prefixes so based on that I also assume this direct import should be harmless. As far as I tried, I didn't notice any broken UI in the single delegation interface after this change

apps/browser-extension-wallet/src/hooks/useIsMultiDelegationEnabled.ts
```diff
@@ -0,0 +1,13 @@
+import { Wallet } from '@lace/cardano';
+import { useWalletStore } from '@src/stores';
+import { useMemo } from 'react';
+
+export const useIsMultiDelegationEnabled = (): boolean => {
+  const { getKeyAgentType } = useWalletStore();
+  const inMemoryWallet = useMemo(
+    () => getKeyAgentType() === Wallet.KeyManagement.KeyAgentType.InMemory,
```
refi93, line 8: @szymonmaslowski I've seen this logic repeated in the codebase quite a lot (dozens of times), I wouldn't refactor it now due to the widespread impact, but from looking at the codebase it seems to me this could be a fine candidate for a custom selector into the `walletStore()`, like `stakePoolResultsSelector`, what do you think? If so, I guess we can make it a backlog refactoring task and get to it after the release

szymonmaslowski: Agree, good idea!

apps/browser-extension-wallet/src/hooks/useIsMultiDelegationEnabled.ts
```diff
@@ -0,0 +1,13 @@
+import { Wallet } from '@lace/cardano';
+import { useWalletStore } from '@src/stores';
+import { useMemo } from 'react';
+
+export const useIsMultiDelegationEnabled = (): boolean => {
```
szymonmaslowski, line 5: How about calling it `useMultiDelegationEnabled`?

apps/browser-extension-wallet/src/views/browser-view/index.tsx
szymonmaslowski: The bundle size is not a concern for us as all the sources are loaded during installation. I guess the purpose of the conditional was to be "more correct". The important thing thought is that we need it to be loaded very late due to the problem with the styles ordering we encountered. I guess this is the reason we used require. I might be wrong though or not remembering the entire reason behind. I wish I documented it with a comment...

Anyway, please check the overview page in dark mode. If headers are white-ish then we are fine 😉

szymonmaslowski (approved):
Well done!

xdzurman (approved):

refi93:
> Anyway, please check the overview page in dark mode. If headers are white-ish then we are fine

@szymonmaslowski Didn't notice any difference between `main` and this PR in the overview screen so I guess it's fine

this branch:
Screenshot 2023-08-21 at 22 31 47

main:
Screenshot 2023-08-21 at 22 28 11

apps/browser-extension-wallet/src/hooks/useIsMultiDelegationEnabled.ts
refi93: sure will do 👍

apps/browser-extension-wallet/src/hooks/useIsMultiDelegationEnabled.ts
refi93: added this task into the backlog: https://input-output.atlassian.net/browse/LW-8109

ManusMcCole (approved):
Looks good @refi93 👍

PR #418, opened 2023-08-18
[WIP]Feat/feature flag post hog client implementation

jplorek-atix:
# Checklist

- [ ] JIRA - \<[LW-7866](https://input-output.atlassian.net/browse/LW-7866)>
- [ ] Screenshots added.

---

## Proposed solution

This PR implements Post Hog feature flag. For this a `FeatureFlags` class has been implemented, this receive a post hog client as a dependency to be able to communicate with post hog. If the client is not provided the local set of feature flags will be used, also, we can explicitly configure if we want to use the local flags/build time flags. For this there is a env variable that can be set.

**Runtime flag/PostHog flag configuration:**
To be able to fetch flags from PostHog, `onFeatureFlags` method is used, this register an event listener that allow us to fetch the flags periodically. After the event listener is registered we start calling `reloadFeatureFlags` to fetch the flags.

**Error Handling:**
Post Hog does not provide documentation on how they handle network errors, but, by looking the code implementation, seems that they just catch the error and return undefined, there are some case where the console the error as well.
[Here is the send request implementation](https://github.com/PostHog/posthog-js/blob/master/src/posthog-core.ts#L643-L706), this send request method is used when `reloadFeatureFlags` is called

## Testing

To test this, a flag call `NFT_FOLDERS` has been set in post hog. 
This flag is configured to be visible only for user with the property `user_tracking_type = enhanced`, so, only opted in user will be able to see this feature

**Opted in**

- As an opted in user in lace, go to the NFT section, the NFT folder feature is enabled

**Opted out**

- As an opted out user in lace, go to the NFT section, the NFT folder feature is disabled

**Disable the flag for all type of users**

- Go to Post Hog Feature Flags section, you will see a table with the configured flags.
- On the last column of the table there will be three dot button, click it, a menu will be display,
- Click over _**Disable feature flag**_
- Go to lace, the NFT folder feature should not be visible for opted in and opted out users

## Screenshots

[LW-7866]: https://input-output.atlassian.net/browse/LW-7866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1556/5906275019/index.html) for [e1a4219e](https://github.com/input-output-hk/lace/pull/418/commits/e1a4219e56416e8ecbd03dc07999976052577d8e)


apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -61,6 +62,10 @@ export class AnalyticsTracker {
     await this.userIdService?.extendLifespan();
   }
 
+  getPostHogFeatureFlag(callback: (flags: Array<string>) => void): Subscription {
+    return this.postHogClient?.subscribeToRemoteFlags(callback);
```
jplorek-atix, line 66: Right now, we expose the flags subscriber through the analytics tracker since we need the posthog client for this. But for me, the feature flag should be handled outside the analytics context, so we might want to refactor this and move the post hog client initialization somewhere else

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
...
+      this.loadLocalFlags();
+    } else {
+      this.#enabledFlags = new Subject();
+      this.startPollingFlagsFromPostHog();
+    }
+  }
+
+  startPollingFlagsFromPostHog(): void {
+    this.pollFlagsFromPostHog();
+    this.posthogInstance.onFeatureFlags((flags) => {
```
jplorek-atix, line 22: Loads the flags for the first time and register an event listener, so, every time `posthogInstance.reloadFeatureFlags()` is called, this will emit a new value to `#enabledFlags`

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
+import { config } from '@src/config';
+import { PostHog } from 'posthog-js';
+import { Subject, BehaviorSubject, Subscription } from 'rxjs';
+
+const FEATURE_FLAGS_POLLING_INTERVAL = 10_000;
```
jplorek-atix, line 5: We can add an env variable for this

apps/browser-extension-wallet/src/providers/AnalyticsProvider/postHog/PostHogClient.ts
```diff
@@ -36,7 +40,10 @@ export class PostHogClient {
       disable_persistence: true,
       disable_cookie: true,
       persistence: 'memory',
-      property_blacklist: ['$autocapture_disabled_server_side', '$device_id', '$time']
+      property_blacklist: ['$autocapture_disabled_server_side', '$device_id', '$time'],
+      loaded: (posthogInstance: PostHog) => {
+        this.#featureFlagsInstance = new FeatureFlags(posthogInstance, USE_LOCAL_FLAGS);
```
jplorek-atix, line 45: This allows us to initialize the feature flags after Post Hog client initialization

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
...
+    });
+  }
+
+  subscribeToRemoteFlags(callback: (params: Array<string>) => void): Subscription {
+    return this.#enabledFlags.subscribe(callback);
+  }
+
+  private pollFlagsFromPostHog() {
+    setInterval(() => {
+      this.posthogInstance.reloadFeatureFlags();
```
jplorek-atix, line 33: This will fetch the post hog flag config after a period of time

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
jplorek-atix: The value of flags is an array of enabled posthog flags, all disabled flags will not be present in this array

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const analyticsContext = useContext(FeatureFlagsContext);
+  if (analyticsContext === null) throw new Error('context not defined');
+  return analyticsContext;
+};
+
+// eslint-disable-next-line arrow-body-style
+export const FeatureFlagsProvider = ({ children }: FeatureFlagsProviderProps): React.ReactElement => {
+  const analytics = useAnalyticsContext();
+  const [enabledFeature, setEnabledFeature] = useState<Array<string>>();
+  const [flagsHaveBeenLoaded, setFlagsHaveBeenLoaded] = useState(false);
```
jplorek-atix, line 29: The load process of the flag is not immediate and can cause to render the incorrect component until the proper flag state is loaded, so, we can use this state to display a loader and avoid this behavior

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
```diff
@@ -0,0 +1,50 @@
...
+      setEnabledFeature(flags);
+      setFlagsHaveBeenLoaded(true);
+    });
+
+    return () => {
+      if (subscription) subscription.unsubscribe();
+    };
+  }, [analytics]);
+
+  const isFeatureEnabled: FeatureFlags['isFeatureEnabled'] = (flag) =>
```
jplorek-atix, line 42: Also, we can expose a method calling `posthogInstance.getFeatureFlag(FeatureFlagkey)`. This will be useful if we want to implement A/B testing https://posthog.com/tutorials/webflow-ab-tests#create-an-experiment-in-posthog

apps/browser-extension-wallet/src/providers/FeatureFlags/types.ts
```diff
@@ -0,0 +1,13 @@
+export type Flags =
```
marcin-mazurek, line 1: Maybe we could extend this type to act as configuration defining available runtime feature flags, possible values and the default value? Regarding possible values, I think we need to have a bit more futureproof structure. While this spike is purely about A/B testing where booleans should do the trick, but in the future, we will be also looking at multivariate tests (eg. three cases A/B/C). The default value would be the value used in case PostHog is unavailable, if it's an enum then it probably should be explicitly defined

marcin-mazurek (commented):
Hey @jplorek-atix, excellent job 🚀 Left a bunch of comments, please take a look.

One more question: did you think if there is any potential refactoring possible for the current build time flagging mechanism? I imagined we could have some central config where we could easily configure switch between making build time or runtime, but it would need to be something Webpack understands, so it can remove code from the final bundle. So I'm not sure if this is feasible and makes sense, but just checking, as this was one of the acceptance criteria.

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
jplorek-atix: When adding an experiment in Post Hog you can configure different types of test groups, when you call this `getFeatureFlag(<experiment-flag-key>)`, this will return one of the group randomly, so, we can display one thing or another based on the group. What I meant here, is that what I did might not be the best implementation and we can improve this by using `getFeatureFlag` method

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
```diff
@@ -0,0 +1,50 @@
...
+    flagsHaveBeenLoaded && enabledFeature.includes(flag);
```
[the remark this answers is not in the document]
jplorek-atix: 1. Yes, but I'm not totally sure how do we want to cache this, do we want to use background storage for this?
2. Yes, `enabledFeatures` will be an empty array and `isFeatureEnabled` will be false

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
...
+    }, FEATURE_FLAGS_POLLING_INTERVAL);
+  }
+
+  private loadLocalFlags() {
+    const { FEATURE_FLAGS } = config();
+    const enabledFlags = [];
+
+    for (const [key, isEnabled] of Object.entries(FEATURE_FLAGS)) {
```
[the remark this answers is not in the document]
marcin-mazurek: Yes, but we should treat them separate.
Environment variable flags will be used for development process when working on large feature eg. Trezor.
Once the feature is ready, we can roll it out gradually with runtime flags.

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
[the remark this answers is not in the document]
marcin-mazurek: > Yes, but I'm not totally sure how do we want to cache this, do we want to use background storage for this?
Yes, background storage. We don't have to do it in this PR but in final implementation, let's document it though 🙏

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
marcin-mazurek: As agreed on the call - can you remove this interval? 🙏

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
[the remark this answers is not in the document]
marcin-mazurek: Added as requirement in [LW-8239](https://input-output.atlassian.net/browse/LW-8239), closing.

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
marcin-mazurek: Captured in [LW-8239](https://input-output.atlassian.net/browse/LW-8239), thanks

apps/browser-extension-wallet/src/providers/FeatureFlags/FeatureFlags.ts
marcin-mazurek: I can see it's removed, thanks :)

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -38,14 +40,19 @@ export class AnalyticsTracker {
     }
   }
 
-  async setOptedInForEnhancedAnalytics(status: EnhancedAnalyticsOptInStatus): Promise<void> {
+  async setOptedInForEnhancedAnalytics(
+    status: EnhancedAnalyticsOptInStatus,
+    chain?: Wallet.Cardano.ChainId
```
marcin-mazurek, line 45: Do we have to pass this as an additional parameter? PostHogClient should be aware of the current chain ID as it exposes a method called `setChain` which we call whenver it changes.

apps/browser-extension-wallet/src/providers/FeatureFlags/context.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const useFeatureFlagsContext = (): FeatureFlags => {
+  const analyticsContext = useContext(FeatureFlagsContext);
+  if (analyticsContext === null) throw new Error('context not defined');
+  return analyticsContext;
+};
+
+// eslint-disable-next-line arrow-body-style
+export const FeatureFlagsProvider = ({ children }: FeatureFlagsProviderProps): React.ReactElement => {
+  const analytics = useAnalyticsContext();
+  const [flagsHaveBeenLoaded, setFlagsHaveBeenLoaded] = useState(false);
```
marcin-mazurek, line 24: How about using this shorter form?
```suggestion
  const [flagsLoaded, setFlagsLoaded] = useState(false);
```

DominikGuzei (approved):
Amazing work 🎉

marcin-mazurek:
Well done, great POC - let's keep this PR open as we will be making this production ready.

jplorek-atix:
Comments will be address in ticket [LW-8239](https://input-output.atlassian.net/browse/LW-8239)

tomislavhoracek (approved):
🎉

marcin-mazurek:
Closing as solution will be finalised in https://github.com/input-output-hk/lace/pull/537

PR #430, opened 2023-08-22
fix(extension): reset user analytics status when wallet is removed

jplorek-atix:
# Checklist

- [ ] JIRA - \<[LW-8122](https://input-output.atlassian.net/browse/LW-8122)>
- [ ] Screenshots added.

---

## Proposed solution

When the wallet was removed the `optedInForEnhancedAnalytics` state in `AnalyticsProvider/context.tsx` was never set to its default value again. Because of this, the effect that is calling `analyticsTracker.setOptedInForEnhancedAnalytics(optedInForEnhancedAnalytics)` was never triggered after the user accepts analytics again, since `optedInForEnhancedAnalytics` never changed the previous value (ACCEPTED)

To solve this, a listener to the current route path is added, so, when the user is redirected to `/setup`, we can reset the state of `optedInForEnhancedAnalytics` to `REJECTED`

[LW-8122]: https://input-output.atlassian.net/browse/LW-8122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1773/6016681932/index.html) for [28ccbab7](https://github.com/input-output-hk/lace/pull/430/commits/28ccbab75de4f01e3a49cf155329d58acf33e9af)


DominikGuzei (approved):
Great catch, simple solutuion 👍

marcin-mazurek (commented):
From what I remember, we should be handling this in [useWalletManager hook](https://github.com/input-output-hk/lace/blob/main/apps/browser-extension-wallet/src/hooks/useWalletManager.ts#L299) - can you check why is it not working?

marcin-mazurek (approved):
I'm still not 100% happy 😄 With this approach, we have some leaky abstraction - I think that `AnalyticsProvider` shouldn't necessarily be aware of Settings module. But it's probably the easiest solution we can do without larger refactoring, so let's go with it for now, but let's review it after the release and see how we could improve it.

marcin-mazurek:
@bslabiak can you re-test it please?

bslabiak (approved):

PR #375, opened 2023-08-08
feat(extension): add custom person properties

jplorek-atix:
# Checklist

- [ ] JIRA - \<[LW-7651](https://input-output.atlassian.net/browse/LW-7651)>
- [ ] Proper tests implemented
- [ ] Screenshots added.

---

## Proposed solution

This PR implements custom person properties

The following properties are linked once to the user id and will be visible in the `Set Once` list of properties

- initial_hardware_wallet_model
- initial_firmware_version
- initial_cardano_app_version

The `user_tracking_type` property is sent in every user event and will be visible in the `Set` list of properties

## Testing

To test the `user_tracking_type` properties is needed to send an event, this property will appears in the list `Set` and can have two values 

- enhanced: opted in user
- basic: opted out user

The Hardware Wallet properties `initial_hardware_wallet_model`, `initial_firmware_version` and `initial_cardano_app_version` will be set only once and will be sent in `onboarding | hardware wallet | all done | go to my wallet | click` event, under the `Set Once` property list

[LW-7651]: https://input-output.atlassian.net/browse/LW-7651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
# Allure report
`allure-report-publisher` generated test report!

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1695/5968229741/index.html) for [85cd6e39](https://github.com/input-output-hk/lace/pull/375/commits/85cd6e39b9c48e98979899182f6390c6b8572a6b)
|       | passed | failed | skipped | flaky | total | result |
|-------|--------|--------|---------|-------|-------|--------|
| Total | 34     | 1      | 0       | 0     | 35    | ❌     |

PR #440, opened 2023-08-23
feat(extension): add flag for disable opted out user tracking

jplorek-atix:
# Checklist

- [ ] JIRA - \<[LW-8082](https://input-output.atlassian.net/browse/LW-8082)>
- [ ] Screenshots added.

---

## Proposed solution

This PR implements a flag for enable/disable tracking for opted out users.
This adds two environment variables, one for Post Hog and another for Matomo

`USE_POSTHOG_ANALYTICS_FOR_OPTED_OUT`
`USE_MATOMO_ANALYTICS_FOR_OPTED_OUT`

## Testing

**Disable opted out users tracking**
By default the tracking for opted out user is disabled.
Try to send an event as an opted out user, there shouldn't be any event in Post Hog and any network request to Post Hog in the network tab.

**Enable opted out users tracking**
set `USE_POSTHOG_ANALYTICS_FOR_OPTED_OUT` to `true` in the `.env` file.
Send an event as an opted out user, there should be an event in Post Hog and a network request to Post Hog in the network tab.

### DIFF
--- a/apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
+++ b/apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
@@ -6,25 +6,34 @@ import {
   PostHogProperties
 } from './types';
 import { Wallet } from '@lace/cardano';
-import { MatomoClient } from '../matomo';
-import { POSTHOG_ENABLED, PostHogClient } from '../postHog';
+import { MatomoClient, MATOMO_OPTED_OUT_EVENTS_DISABLED } from '../matomo';
+import { POSTHOG_ENABLED, POSTHOG_OPTED_OUT_EVENTS_DISABLED, PostHogClient } from '../postHog';
 import { getUserIdService } from '@providers/AnalyticsProvider/getUserIdService';
 import { UserIdService } from '@lib/scripts/types';
 
+interface AnalyticsTrackerArgs {
+  extensionParams: {
+    chain: Wallet.Cardano.ChainId;
+    view?: ExtensionViews;
+  };
+  analyticsDisabled?: boolean;
+  isPostHogEnabled?: boolean;
+  isOptedInUser?: boolean;
+}
 export class AnalyticsTracker {
   protected matomoClient?: MatomoClient;
   protected postHogClient?: PostHogClient;
   protected userIdService?: UserIdService;
+  protected isOptedInUser?: boolean;
 
-  constructor(
-    extensionParams: {
-      chain: Wallet.Cardano.ChainId;
-      view?: ExtensionViews;
-    },
+  constructor({
+    extensionParams,
     analyticsDisabled = false,
-    isPostHogEnabled = POSTHOG_ENABLED
-  ) {
+    isPostHogEnabled = POSTHOG_ENABLED,
+    isOptedInUser = false
+  }: AnalyticsTrackerArgs) {
     if (analyticsDisabled) return;
+    this.isOptedInUser = isOptedInUser;
     this.userIdService = getUserIdService();
     this.matomoClient = new MatomoClient(extensionParams.chain, this.userIdService);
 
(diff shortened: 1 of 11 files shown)


apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -44,19 +53,23 @@ export class AnalyticsTracker {
   }
 
   async sendPageNavigationEvent(): Promise<void> {
+    if (POSTHOG_OPTED_OUT_EVENTS_DISABLED && !this.isOptedInUser) return;
```
marcin-mazurek, line 56: 1. Could we use UserIdService changes from #375 to avoid introducing additional property?
```suggestion
    if (POSTHOG_OPTED_OUT_EVENTS_DISABLED && this.userIdService.getTrackingType() === UserTrackingType.Basic) return;
```
2. Could we extract this line to a helper method to reduce duplication?

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -6,25 +6,34 @@ import {
...
   protected userIdService?: UserIdService;
+  protected isOptedInUser?: boolean;
 
-  constructor(
-    extensionParams: {
-      chain: Wallet.Cardano.ChainId;
-      view?: ExtensionViews;
-    },
+  constructor({
+    extensionParams,
```
marcin-mazurek, line 30: Just wondering, what's the benefit of having arguments here in form of a nested object instead of flat one? We don't seem to be passing `extensionParams` as object anywhere, we're anyway accessing each property individually, so we might as well declare all params as one flat structure

marcin-mazurek (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/settings/components/SettingsSecurity.tsx
```diff
@@ -37,13 +37,15 @@ export const SettingsSecurity = ({
...
   const backgroundService = useBackgroundServiceAPIContext();
+  const analytics = useAnalyticsContext();
 
   const showPassphraseVerification = process.env.USE_PASSWORD_VERIFICATION === 'true';
 
   const handleAnalyticsChoice = (isOptedIn: boolean) => {
     setEnhancedAnalyticsOptInStatus(
       isOptedIn ? EnhancedAnalyticsOptInStatus.OptedIn : EnhancedAnalyticsOptInStatus.OptedOut
     );
+    analytics.setIsOptedInUser(isOptedIn);
```
DominikGuzei, line 48: seems like the unit tests cover a different scenario (passing `isUserOptedIn` via constructor) instead of using this method?

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
jplorek-atix: Now that you mention, there is no benefit on having `extensionParams` as an object. I've changed this to a flat structure

apps/browser-extension-wallet/src/views/browser-view/features/settings/components/SettingsSecurity.tsx
jplorek-atix: yes, I refactored this, and now we are using the `userIdService` to get the user tracking mode

apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts
```diff
@@ -44,19 +49,27 @@ export class AnalyticsTracker {
   }
 
   async sendPageNavigationEvent(): Promise<void> {
+    const shouldOmitEvent = await this.shouldOmitSendEventToPostHog();
+    if (shouldOmitEvent) return;
```
jplorek-atix, line 53: @marcin-mazurek @DominikGuzei 
I added the helper for this, but, unfortunately I keep repeating the if statement in all methods. Do you have any suggestion on how to improve this?

marcin-mazurek: It's good, I don't think there is anything more we can do :) Btw. you can skip `await` before `shouldOmitSendEventToPostHog` and `getUserTrackingType`, they are synchronous.

marcin-mazurek: If you remove await you can probably write it as a oneliner:

```
if (this.shouldOmitSendEventToPostHog()) return;
```

marcin-mazurek: Follow up - as discussed on Slack, it's not possible as all background services call are asynchronous

marcin-mazurek (approved):

DominikGuzei (approved):

bslabiak (approved):

PR #448, opened 2023-08-25
feat(staking) LW-8145  show percentages instead of fractions in the multistaking portfolio draft

refi93:
# Checklist

- [x] JIRA - [LW-8145](https://input-output.atlassian.net/browse/LW-8145)
- [x] Screenshots added.

---

## Proposed solution

1. replaced the fraction with percentage by formatting the fraction value with the pre-existing `formatPercentages()` helper
2. though not strictly needed (we split stake evenly, up to 5 pools max, and show integer percentages), I found that the `formatPercentages` helper wouldn't work as expected for decimal places beyond `2` and the rounding logic may end up rounding up percentages above half which in our case wouldn't be desired as the percentages shown may add up beyond `100%` which looks odd. To address these issues I extended the API of the helper to be more explicit (named parameters, explicit rounding strategy) and covered it with unit tests for visibility and future extensibility
3. clarified with Fernando/Mathieu that `1/3` should be shown as `33%`, see comments in the Jira ticket

## Testing

1. In the expanded view, go to Staking -> Browse pools
2. Pick 5 pools (current maximum), click "Next"
4. check that in the summary screen the percentages are shown properly - should not contain any decimals. You can try removing pools to gradually go through the case of 5,4,3,2,1 pools

## Screenshots

before:
Screenshot 2023-08-25 at 10 57 32

after:
Screenshot 2023-08-25 at 10 55 45

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1699/5974146099/index.html) for [6ca01d91](https://github.com/input-output-hk/lace/pull/448/commits/6ca01d91b2b6af00c3dcf2c923475509e94b387b)

[LW-8145]: https://input-output.atlassian.net/browse/LW-8145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/common/src/ui/lib/format-number.ts
```diff
@@ -3,5 +3,25 @@ import { Percent } from '@cardano-sdk/util';
 
 export const DEFAULT_DECIMALS = 2;
 
-export const formatPercentages = (number: number | Percent, decimalPlaces: number = DEFAULT_DECIMALS): string =>
-  (Math.round(number.valueOf() * 100 * 100) / 100).toFixed(decimalPlaces);
+type FormatPercentagesOptions = {
+  decimalPlaces?: number;
+  rounding?: 'down' | 'half_up';
```
refi93, line 8: I considered making this parameter an enum but in the end I didn't like the redundancy of having to import the enum to use it in whatever code would be using this helper and property, and also unless the enum was a string enum (requiring redundant declarations of the enum values, it would have less type safety as raw numbers would also be accepted by TS. https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant

If the agreed on approach is to prefer enums, I can refactor this, of course

packages/common/src/ui/lib/__tests__/format-number.test.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable no-magic-numbers */
+import { formatPercentages } from '../format-number';
+
+describe('Number formatters', () => {
+  describe('getCapitalizedInitial', () => {
```
refi93, line 5: note: I just noticed I forgot to update the test name, will fix together with other potential comments

packages/common/src/ui/lib/format-number.ts
```diff
@@ -3,5 +3,25 @@ import { Percent } from '@cardano-sdk/util';
...
+  value: number | Percent,
+  { decimalPlaces = DEFAULT_DECIMALS, rounding = 'half_up' }: FormatPercentagesOptions = {}
+): string => {
+  const unroundedValue = value.valueOf() * Math.pow(10, decimalPlaces) * 100;
+  let roundedValue: number;
+  switch (rounding) {
+    case 'down':
+      roundedValue = Math.floor(unroundedValue);
+      break;
+    case 'half_up':
```
xdzurman, line 21: I'd consider changing `half_up` to `default` or `nearest`, the snake case strikes my eye a little bit. Also if a need arises to add the `ceil` option, it'd be more natural

xdzurman (approved):

przemyslaw-wlodek (approved):
@refi93 great work :) I approved but please update the test note and union casing 🍺

oldGreg5 (approved):

PR #455, opened 2023-08-28
feat(staking): LW-8126 add returned deposit info to de-registration tx confirm

xdzurman:
**based on 7674 since this PR requires the `depositReclaim` logic**

# Checklist

- [x] JIRA - [link](https://input-output.atlassian.net/browse/LW-8126)
- [x] Screenshots added.
- [x] [Figma design](https://www.figma.com/file/RPN3V4Et7hMbXkV5ZIF75i/Lace-Browser---Full-view-UI-Design?type=design&node-id=24821-55816&mode=design&t=Yl5wOxtJWqc4jrRQ-4)

---

## Proposed solution

According to figma designs, render the returned deposit and total fee effect with a `+` sign.

## Testing

Stake from N pools to N-1 pools to trigger the UI which shows the `Transaction return` and `Total` rows in transaction confirm screen. 

## Screenshots

image

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1839/6047213905/index.html) for [5d6f16b0](https://github.com/input-output-hk/lace/pull/455/commits/5d6f16b0d7f80e4a9fcb38cd1a0adaf48de2ee97)


packages/staking/src/features/drawer/StakePoolConfirmation.tsx
```diff
@@ -104,6 +104,67 @@ const StakePoolConfirmationBody = ({
   );
 };
 
+type StakePoolDepositReclaimDetailsProps = {
+  delegationTxDeposit: number;
+};
+
+const StakePoolDepositReclaimDetails = ({
```
szymonmaslowski, line 111: Suggestion: IMO there is no big value in creating this component. I would rather try to create some building block components/functions for avoid code repetition (e.g. for abstracting usage of `renderAmountInfo`) and put all the content in the same component

packages/staking/src/features/i18n/types.ts
```diff
@@ -88,10 +88,17 @@ type KeysStructure = {
...
+        title: '';
+      };
+      transactionReturn: {
+        title: '';
+      };
+      transactionTotal: {
         title: '';
       };
       theAmountYoullBeChargedForRegisteringYourStakeKey: '';
+      theAmountYoullBeAwardedForDeregisteringYourStakeKey: '';
```
szymonmaslowski, line 101: Wow these are really long keys... 😄

szymonmaslowski (approved):
Left some suggestions 😉

packages/staking/src/features/drawer/StakePoolConfirmation.tsx
xdzurman: I extracted the `renderAmountInfo` functionality into a component. I made it stateless and pass all props to it - however I think a cleaner version would be to make it use the outsideHandles and get cardanoCoin, fiatCurrency, cardanoFiatPrice from a hook to avoid duplicity. Although, I'm not sure if this would be as performant, since many instances of this component would call the outside handles hook. We had a problem with this on other project, but what do you think?

Also, I disagree that a standalone component does not bring much value - it helps it deal with txDepositReclaim, totalTxEffect on its own without having to consider existence of these (thanks to it being render only when `delegationTxDeposit < 0` and keep this functionality well isolated

PR #465, opened 2023-08-30
ci(all): Fix labeler not matching staking paths

refi93:


.github/workflows/labeler.yml
```diff
@@ -1,6 +1,6 @@
 name: 'Label PRs'
 on:
-  - pull_request_target
+  - pull_request
```
refi93, line 3: temporary change to force using the workflow definition from this PR, see https://stackoverflow.com/questions/74143357/github-test-workflow-in-branch

.github/workflows/labeler.yml
```diff
@@ -9,7 +9,10 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
       - uses: actions/labeler@v4
         with:
           repo-token: '${{ secrets.GITHUB_TOKEN }}'
           sync-labels: true
+          dot: true
```
refi93, line 17: needed to recognize files under the `.github` folder:  

> This action uses [minimatch](https://www.npmjs.com/package/minimatch) to apply glob patterns. For historical reasons, paths starting with dot (e.g. .github) are not matched by default. You need to set dot: true to change this behavior. See [Inputs](https://github.com/actions/labeler#inputs) table below for details.

https://github.com/actions/labeler

.github/labeler.yml
```diff
@@ -5,13 +5,16 @@ browser:
...
+  - any: ['packages/staking/**/*']
+  - any: ['.github/workflows/packages-staking.yml']
+
 # UI Toolkit
 ui-toolkit:
   - any: ['packages/ui/**/*']
 # E2E
 e2e:
-  - any: ['features/**/*', 'packages/e2e-tests/**/*']
+  - any: ['features/**/*']
```
refi93, line 16: now that I'm checking the codebase it seems there's no `features` folder anymore so I guess this could be dropped @przemyslaw-wlodek

.github/labeler.yml
```diff
@@ -5,13 +5,15 @@ browser:
...
-  - any: ['packages/staking/**/*', '.github/workflows/packages-staking.yml']
+  - any: ['packages/staking/**/*']
+  - any: ['.github/workflows/packages-staking.yml']
+
 # UI Toolkit
 ui-toolkit:
   - any: ['packages/ui/**/*']
 # E2E
 e2e:
-  - any: ['features/**/*', 'packages/e2e-tests/**/*']
```
refi93, line 14: now that I'm checking the codebase it seems there's no features folder anymore so I guess this could be dropped @przemyslaw-wlodek

note: in theory, this may match any "features" subfolder (not sure TBH), but as far as I searched the codebase, the "features" folder(s) found seemed to be plain UI stuff rather than anything concerning E2E tests

przemyslaw-wlodek (approved):
Thank you very much @refi93 for fixing this! 🚀🍻
P.S. When I added the previous `any` element I was just following the existing pattern. It was not looking suspicious 🥲

mchappell (approved):

PR #472, opened 2023-08-31
Fix(extension): LW-7358 fix incorrect time being shown for activities 

refi93:
# Checklist

- [x] JIRA - LW-7358
- [x] Screenshots added.

---

## Proposed solution

Date/Time was formatted as UTC instead of local time resulting in the times shown in the UI being off by user's time offset, so the fix consists in extending the time/date formatting helpers to allow for utc vs local time and use local time in activity. The formatters were refactored to require explicitly stating the time standard (local vs UTC) to make the intention clear to the caller/reader and avoid future confusion.

These helpers were also used to format data about blocks (used in activity details modal) and UTC was used there apparently intentionally, so to make that clear, this PR involves also refactoring the formatted time fields to explicitly contain "utc" in their name to make this clear to the reader and avoid confusion as the values are plain strings.

I also noticed that we needlessly append "AM/PM" to otherwise 24H-formatted times in the activity and the transaction times in the details modal are 12H-formatted, so I unified both to 24H (while keeping the transaction detail time in UTC) - agreed with PM on that, see JIRA task comments for context

## Testing

Visit "Activity" tab and check that the time of activities corresponds to the actual (local) time they happened at (e.g. via block explorer). Check also that UTC time shown in the activity details is correct

## Screenshots

Before:
Screenshot 2023-08-31 at 16 57 07

Screenshot 2023-08-31 at 16 57 14

After:
Screenshot 2023-08-31 at 16 56 41
Screenshot 2023-08-31 at 16 56 48

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/1873/6074315542/index.html) for [fe1e006e](https://github.com/input-output-hk/lace/pull/472/commits/fe1e006efbb87cdcd51412c2a084ffa842c49806)


apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/pending-tx-transformer.ts
```diff
@@ -83,7 +83,11 @@ export const txTransformer = ({
     direction
   });
   const outputAmount = new BigNumber(coins.toString());
-  const timestamp = formatTime(time, 'HH:mm:ss A');
+  const timestamp = formatTime({
+    date: time,
+    format: 'HH:mm:ss',
```
mrcnk, line 88:
```suggestion
    format: DEFAULT_TIME_FORMAT,
```

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/__tests__/pending-tx-transformer.test.ts
```diff
@@ -110,7 +110,11 @@ describe('Testing tx transformers utils', () => {
           }
         ],
         assetsNumber: 2,
-        timestamp: formatTime(time, 'HH:mm:ss A')
+        timestamp: formatTime({
+          date: time,
+          format: 'HH:mm:ss',
```
mrcnk, line 115:
```suggestion
          format: DEFAULT_TIME_FORMAT,
```

apps/browser-extension-wallet/src/api/transformers.ts
```diff
@@ -120,8 +120,8 @@ export const blockTransformer = (block: Wallet.BlockInfo): TransactionDetail['bl
   confirmations: block.confirmations.toString(),
   size: block.size.toString(),
   transactions: block.txCount.toString(),
-  date: formatDate(block.date, 'MM/DD/YYYY'),
-  time: `${formatTime(block.date, 'h:mm:ss A')} UTC`,
+  utcDate: formatDate({ date: block.date, format: 'MM/DD/YYYY', type: 'utc' }),
+  utcTime: `${formatTime({ date: block.date, format: 'HH:mm:ss', type: 'utc' })} UTC`,
```
mrcnk, line 124:
```suggestion
  utcTime: `${formatTime({ date: block.date, format: DEFAULT_TIME_FORMAT, type: 'utc' })} UTC`,
```

mrcnk (changes requested):

mrcnk:
Please use the format template in constant and then we're good to go 🎉

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/pending-tx-transformer.ts
refi93: given these formats are equal to the default time/date format, I'd consider just dropping the format parameter from the function invocation completely, as the formatDate/Time falls back to it anyway. This should make it clear to the reader that some default format is used without being overly verbose, what do you think?

mrcnk: Yes, makes sense to drop importing this format in all the places, but just make default format function that bundles it 👍

refi93: indeed, for the tests it IMHO makes sense to be explicit (didn't have that in mind in my original comment), but I'll remove the explicit default format from the application code

mrcnk (approved):

apps/browser-extension-wallet/src/utils/format-date.ts
```diff
@@ -4,14 +4,20 @@ import utc from 'dayjs/plugin/utc';
...
-    .utc()
+    .utc(type === 'local')
     .format(format ?? DEFAULT_DATE_FORMAT);
 
-export const formatTime = (date: Date | string | number, format?: string): string =>
+export const formatTime = ({ date, format, type }: FormatDateTimeParams): string =>
   dayjs(date)
-    .utc()
-    .format(format ?? DEFAULT_TIME_FORTMAT);
+    .utc(type === 'local')
```
szymonmaslowski, line 22: I think this implementation is problematic. In case user passes type `local` and a format which include the time zone then as a result we get modified timestamp indicating different point in time. Wouldn't it be more correct to just skip the `.utc()` in case of a local time?

szymonmaslowski (changes requested):

apps/browser-extension-wallet/src/utils/format-date.ts
refi93: good point, will fix it 👍

refi93: I was just testing the dayjs lib and it doesn't seem to behave as I originally expected (`utc(false)` disabling casting to `utc`)

```javascript
var dayjs = require("dayjs")
var utc = require('dayjs/plugin/utc');
dayjs.extend(utc);


console.log(
  dayjs(Date()).format('HH:mm:ss'), // prints 13:26:17
  dayjs(Date()).utc(false).format('HH:mm:ss') // prints 11:26:17 (expected 13:26:17, too need to understand why yet)
)
```
the fix in this PR solves the issue though I'm not sure right now if it just isn't compensating for some timezone/standard inconsistency in the underlying data, I will investigate it deeper in a follow up issue, shouldn't be a blocker for this PR - thanks for the catch @szymonmaslowski

oldGreg5 (approved):

apps/browser-extension-wallet/src/utils/format-date.ts
refi93: @szymonmaslowski made follow-up issue to investigate this more in depth: https://input-output.atlassian.net/browse/LW-8325

ManusMcCole (approved):
Great job @refi93

szymonmaslowski (approved):

PR #515, opened 2023-09-08
fix(ui): [LW-6898] fix incorrect tab button active state and tab navigation

refi93:
# Checklist

- [x] JIRA - LW-6898
- [x] Screenshots added.

---

## Proposed solution

Fix tab label coloring and border to be black when button is active/focused. While fixing the purple border around the active & not focused tab button I realized that the tab navigation doesn't work as one would normally expect, because on TAB key press only one of the tabs got focused and the other skipped, so this PR is fixing also that

## Testing

In the expanded view go to the "Staking" section and try clicking between tabs, try navigation by pressing TAB/Enter and check that it behaves reasonably (correct tab is focused - purple border shown and after pressing ENTER it gets selected)

Check that the tab label/border colors match the FIgma designs (see Jira tickets for reference)

## Screenshots

There are multiple states of the tab button, I'd just recommend checking Figma and verifying that the UI in this PR matches that

### DIFF
--- a/packages/staking/src/features/staking/Navigation.tsx
+++ b/packages/staking/src/features/staking/Navigation.tsx
@@ -16,14 +16,30 @@ export const Navigation = ({ children }: NavigationProps) => {
   }));
   const { t } = useTranslation();
   const onValueChange = (value: string) => {
+    console.log(value);
     if (isValueAValidSubPage(value)) setActivePage(value);
   };
 
   return (
     <>
-      <SubNavigation.Root aria-label={t('root.nav.title')} value={activePage} onValueChange={onValueChange}>
-        <SubNavigation.Item name={t('root.nav.overviewTitle')} value={Page.overview} data-testid="overview-tab" />
-        <SubNavigation.Item name={t('root.nav.browsePoolsTitle')} value={Page.browsePools} data-testid="browse-tab" />
+      <SubNavigation.Root
+        aria-label={t('root.nav.title')}
+        value={activePage}
+        onValueChange={onValueChange}
+        tabIndex={-1}
+      >
+        <SubNavigation.Item
+          name={t('root.nav.overviewTitle')}
+          value={Page.overview}
+          data-testid="overview-tab"
+          tabIndex={0}
+        />
+        <SubNavigation.Item
+          name={t('root.nav.browsePoolsTitle')}
+          value={Page.browsePools}
+          data-testid="browse-tab"
+          tabIndex={0}
+        />
       </SubNavigation.Root>
       {children(activePage)}
     </>
--- a/packages/ui/src/design-system/sub-navigation/sub-navigation-item.css.ts
+++ b/packages/ui/src/design-system/sub-navigation/sub-navigation-item.css.ts
@@ -29,7 +29,7 @@ export const disabled = style([
 
 const labelContainerFocused = {
   borderRadius: vars.radius.$tiny,
-  boxShadow: `0 0 0 1px ${vars.colors.$sub_navigation_container_outlineColor}`,
+  boxShadow: `0 0 0 3px ${vars.colors.$sub_navigation_container_outlineColor}`,
 };
 
 export const labelContainer = style([
@@ -44,9 +44,8 @@ export const labelContainer = style([
   }),
   {
     selectors: {
-      [`${root}:focus:not(:active) &`]: labelContainerFocused,
-      [`${root}:focus:not(${active}) &`]: labelContainerFocused,
-      [`${root}:focus-visble:not(:active) &`]: labelContainerFocused,
+      [`${root}:focus-visible:not(${active}) &`]: labelContainerFocused,
+      [`${root}:focus-visible:not(:active) &`]: labelContainerFocused,
     },
   },
 ]);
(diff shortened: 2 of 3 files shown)


packages/ui/src/design-system/sub-navigation/sub-navigation.component.tsx
```diff
@@ -27,6 +27,7 @@ export const SubNavigation = ({
     defaultValue={defaultValue}
     value={value}
     onValueChange={onValueChange}
+    activationMode="manual"
```
refi93, line 30: without this, upon pressing TAB the tab got automatically selected (the default behavior)

see radix-ui code for reference: https://github.com/radix-ui/primitives/blob/28bebf2c6992d056244845c898abeff45dec2871/packages/react/tabs/src/Tabs.tsx#L74

packages/ui/src/design-system/sub-navigation/sub-navigation-item.css.ts
```diff
@@ -44,9 +44,8 @@ export const labelContainer = style([
   }),
   {
     selectors: {
-      [`${root}:focus:not(:active) &`]: labelContainerFocused,
-      [`${root}:focus:not(${active}) &`]: labelContainerFocused,
-      [`${root}:focus-visble:not(:active) &`]: labelContainerFocused,
+      [`${root}:focus-visible:not(${active}) &`]: labelContainerFocused,
```
refi93, line 47: keeping just `focus-visible` because as per [Figma](https://www.figma.com/file/qGPfPiE6J7Oa7R1I1me6Vi/Lace-Browser---UI-Toolkit?node-id=251%3A12370&mode=dev) I understood that the purple border should be shown only when "focused" and "inactive" which I guess is really possible only when navigating with the keyboard by pressing TAB

blog on `focused` vs `focused-visible` elaborating on this exact usecase (TAB navigation):
https://pawelgrzybek.com/the-difference-between-css-focus-and-focus-visible-pseudo-class/

packages/staking/src/features/staking/Navigation.tsx
```diff
@@ -16,14 +16,30 @@ export const Navigation = ({ children }: NavigationProps) => {
...
+        aria-label={t('root.nav.title')}
+        value={activePage}
+        onValueChange={onValueChange}
+        tabIndex={-1}
+      >
+        <SubNavigation.Item
+          name={t('root.nav.overviewTitle')}
+          value={Page.overview}
+          data-testid="overview-tab"
+          tabIndex={0}
```
refi93, line 35: I had to manually set `tabIndex`-es here because without that the TAB only focused the currently selected TAB and subsequent pressing of the tab skipped somewhere else, rendering keyboard navigation useless

packages/ui/src/design-system/sub-navigation/sub-navigation-item.css.ts
```diff
@@ -29,7 +29,7 @@ export const disabled = style([
 
 const labelContainerFocused = {
   borderRadius: vars.radius.$tiny,
-  boxShadow: `0 0 0 1px ${vars.colors.$sub_navigation_container_outlineColor}`,
+  boxShadow: `0 0 0 3px ${vars.colors.$sub_navigation_container_outlineColor}`,
```
refi93, line 32: From [Figma](https://www.figma.com/file/qGPfPiE6J7Oa7R1I1me6Vi/Lace-Browser---UI-Toolkit?node-id=251%3A12370&mode=dev) I understood that the border is supposed to be 3px thick, fixing it here

packages/ui/src/design-system/sub-navigation/sub-navigation-item.css.ts
```diff
@@ -60,17 +59,23 @@ export const label = style([
       [`${root}:hover &`]: {
         color: vars.colors.$sub_navigation_item_label_color_hover,
       },
-      [`${root}:active &`]: {
+      [`${root}:focus-visible:not(${active}) &`]: {
```
refi93, line 62: note that buttons are active only when pressed, but the tabs are of course active beyond that, which is handled by the `[data-state="active"]` (i.e. `${active}`) selector, hence this duplication

packages/staking/src/features/staking/Navigation.tsx
```diff
@@ -16,14 +16,30 @@ export const Navigation = ({ children }: NavigationProps) => {
   }));
   const { t } = useTranslation();
   const onValueChange = (value: string) => {
+    console.log(value);
```
przemyslaw-wlodek, line 19: I guess it's a leftover :)

packages/staking/src/features/staking/Navigation.tsx
```diff
@@ -16,14 +16,30 @@ export const Navigation = ({ children }: NavigationProps) => {
...
+          name={t('root.nav.overviewTitle')}
+          value={Page.overview}
+          data-testid="overview-tab"
+          tabIndex={0}
+        />
+        <SubNavigation.Item
+          name={t('root.nav.browsePoolsTitle')}
+          value={Page.browsePools}
+          data-testid="browse-tab"
+          tabIndex={0}
```
przemyslaw-wlodek, line 41: should it be `1`?

refi93: `tabIndex="0"` is the default value allowing sequential navigation. I considered putting there increasing numbers but my understanding is that it could potentially clash with other unrelated elements with positive tab-indices, so tabIndex={0} should be a safe default if we just want the element to be "tab-navigable" in some reasonable order

> tabindex="0" means that the element should be focusable in sequential keyboard navigation, after any positive tabindex values. The focus navigation order of these elements is defined by their order in the document source.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

packages/staking/src/features/staking/Navigation.tsx
refi93: ok I see `no-console` is disabled because webpack strips console.logs anyway: https://github.com/input-output-hk/lace/blob/8733e5f356e9f09c0e6822db05c5211bbf485455/.eslintrc.js#L32 - given that there aren't that many instances of `console.log()` I guess we may reconsider that in favor of just explicitly disabling the rule in the relatively few places that it's done

szymonmaslowski: personally I would disable just console.log but allow for other types of logs (info, error etc.) because the thing is that we tend to use console logs for debuging so they are usually a left overs. In case we would like to have some log we could use e.g. info and that would be for sure an intended usage ;)

packages/ui/src/design-system/sub-navigation/sub-navigation-item.css.ts
refi93: I double-checked if it's really needed to add a selector for `:active` and seems like it makes a difference only for the `labelContainer` (purple border around the tab button when navigated with the keyboard), on click the `state="active"` property is apparently applied to the button as soon as I press the mouse button so no need to rely on the css selector, so I'm removing the rest 👍 cc @przemyslaw-wlodek

przemyslaw-wlodek (approved):

oldGreg5 (approved):

PR #523, opened 2023-09-11
feat(staking): present on-chain data during staking management [LW-8267]

przemyslaw-wlodek:
# Checklist

- [x] JIRA - LW-8267
- [x] Screenshots added.

---

## Proposed solution
Present on-chain data in staking management flow. Split the stake equally while building a new portfolio or modifying stake pools in the current portfolio.

## Testing
Please run full regression on new/current portfolio management, adding/removing pools etc.

## Screenshots
Before:
image
image

After:
image

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2082/6241758072/index.html) for [824481f6](https://github.com/input-output-hk/lace/pull/523/commits/824481f6b6f89dced99594c000cf660dacdc96c2)


packages/staking/src/features/drawer/StakePoolPreferences.tsx
```diff
@@ -47,12 +48,15 @@ export const StakePoolPreferences = () => {
...
-  const displayData = draftPortfolio.map(({ name, id }, i) => ({
+  const displayData = draftPortfolio.map(({ name = '-', weight, id }, i) => ({
     color: PIE_CHART_DEFAULT_COLOR_SET[i] as PieChartColor,
     id,
-    name: name || '',
-    value: 1,
+    name,
+    weight,
   }));
+  const createRemovePoolFromPortfolio = (poolId: Wallet.Cardano.PoolIdHex) => () => {
```
xdzurman, line 57:
```suggestion
  const handleRemovePoolFromPortfolio = (poolId: Wallet.Cardano.PoolIdHex) => () => {
```
Personally I liked the original name more. Create and Remove in succession might disturb the reader

packages/staking/src/features/store/delegationPortfolio.ts
```diff
@@ -54,6 +54,9 @@ export const useDelegationPortfolioStore = create(
         set((store) => {
           if (store.activeManagementProcess === PortfolioManagementProcess.None) return;
           store.draftPortfolio = store.draftPortfolio.filter((pool) => pool.id !== id);
+          store.draftPortfolio.forEach((pool) => {
+            pool.weight = 1 / store.draftPortfolio.length;
+          });
```
xdzurman, line 59: In this file, there's this same functionality 3 times. I'd propose creating a function, something along the lines of
```ts
const setWeights<T> = (array: T[]): (T & {weight: number})[] =>
    array.map((elem) => ({...elem, weight: 1 / array.length})) 
```
that way, it can be also called in one-line, e.g.
```ts
store.draftPortfolio = setWeights(store.draftPortfolio.filter((pool) => pool.id !== id));
```
or
```ts
selections = setWeights(selections.filter((pool) => pool.id !== id));
```

I'll leave the names of variables and function to you. WDYT?

xdzurman (approved):
**Code review:**
Left comments to address.

**QA review:**
LGTM

packages/staking/src/features/store/delegationPortfolio.ts
```diff
@@ -94,8 +100,11 @@ export const useDelegationPortfolioStore = create(
         });
       },
       unselectPool: ({ id }) =>
-        set((store) => {
-          store.selections = store.selections.filter((pool) => pool.id !== id);
+        set(({ selections }) => {
+          selections = selections.filter((pool) => pool.id !== id);
+          selections.forEach((pool) => {
+            pool.weight = 1 / selections.length;
```
refi93, line 106: this `weight` is inherited from `Cip17Pool` structure and as per CIP-17, weights should be integers, so this can IMHO be confusing, lead to issues if we integrate with cardano-js-sdk portfolio (which would likely follow CIP-17) in the future:

> For each pool, we demand a weight which can capture a certain stake proportion within the portfolio. The value is an integer, and relative to other weights in the portfolio.

https://cips.cardano.org/cips/cip17/

I'd suggest to keep weights fixed to 1 for now, and rather have a separate `targetPercentage` property (if needed). To avoid further confusion, I'd suggest renaming the `weight` property in our `CurrentPortfolioStakePool` object to `targetWeight` and keep it an integer (`1` for now) to stay consistent with CIP-17. If we need to work also with the target percentages, we can introduce a `targetPercentage` property

btw, I did a [similar refactor in my branch](https://github.com/input-output-hk/lace/pull/533/commits/168403f0bd036e24e76f273451234c7f8c364bb7#diff-e194ffdfafa97ef5423fa5f96ae4ccb759d8e1d240e53fd8904e13849a35edf3R85) implementing the drift warning. I'd suggest merging first your PR which is smaller and I can then base mine on top of your:

packages/staking/src/features/drawer/StakePoolPreferences.tsx
przemyslaw-wlodek: I understand but it's not a handler yet. You need to create it dynamically, passing `poolId` so we can rename it to `createHandleRemovePoolFromPortfolio` if you wish 😄

packages/staking/src/features/store/delegationPortfolio.ts
przemyslaw-wlodek: Percentages and weights are interchangeable. The thing is that we receive percentages from the backend in the form of floating point numbers lower than 1 (summing up to 1). Let's leave further refactoring for the next phase of store implementation.

packages/staking/src/features/store/delegationPortfolio.ts
```diff
@@ -12,6 +17,11 @@ const defaultState: DelegationPortfolioState = {
 
 export const MAX_POOLS_COUNT = 5;
 const LAST_STABLE_EPOCH = 2;
+// interchangeable with percentages
```
refi93, line 20: this obviously may not work properly for portfolios configured externally, especially as soon as the portfolio would be stored on chain, created by some other CIP-17 compliant wallet.

In [my PR](https://github.com/input-output-hk/lace/pull/533/commits/6cedd8ab6e9403423c6ca3af37e07554ca232084) I'm actually splitting target weight and percentage into two separate properties, so to not delay this PR, I suggest fixing this in my PR right after merging this PR - do you perhaps see any strong reason to keep weights and percentages interchangeable for the foreseeable future?

refi93: ok noticed your comment, thanks for the clarification: https://github.com/input-output-hk/lace/pull/523#discussion_r1330772465 - nevertheless I'd try to look into it in my PR (I'd give a try to translating these non-compliant weights to percentages in a sensible way which actually shouldn't be much work and calling that inconsistency with CIP-17 out to the SDK team for the time being)

refi93 (approved):
let's keep the assumption that `weight = percentage` for now and I'll revisit it in my PR #533 anyway - I see no quick fix/refactor right now, especially the reconciliation of on-chain state vs target portfolio, taking into account the initial state vs state after editing, seems tricky and would take me at least a few hours which I'd suggest rather spending on #533 and make it play well even with the notion of portfolio drift at once

tested and LGTM 👍

mrcnk (approved):

PR #533, opened 2023-09-14
feat(staking): LW-7966 stake distribution drift warning

refi93:
# Checklist

- [x] JIRA - LW-7966
- [x] Screenshots added.

---

## Proposed solution

Show drift warning as soon as the portfolio total drift exceeds 15%, see Jira task for exact formula and designs

Besides adding the banner this PR also refactors the `DraftPortfolioStakePool` type, introducing an explicit reference to the original "current portfolio" to preserve the functionality introduced in #543, and splits explicitly "percentage" and "targetWeight", with the intent of improving readability and making the code more resilient to bugs due to the nuances between weights and percentages

note: depends on #539 which this PR is based on and should be fully reviewed/merged only after finishing/merging #539 first

## Testing

1. Balance the portfolio across at least 2 stake pools, no "drift warning" should  not be visible
2. Send some significant portion of total funds to the send address
3. as soon as the tx goes through, the "drift warning" banner should appear and be visible until portfolio is balanced again

* test also that portfolio management drawer still works as expected, try adding/removing pools
* test that while submitting delegation "migrating pools" banner is still visible in the extended view (logic also touched in this PR)

## Screenshots

Screenshot 2023-09-21 at 23 46 25
Screenshot 2023-09-21 at 23 48 37

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2148/6267237218/index.html) for [3ef67bed](https://github.com/input-output-hk/lace/pull/533/commits/3ef67bedeff59d633a0b362b0ae63732706c283f)


packages/staking/src/features/store/types.ts
```diff
@@ -42,17 +42,28 @@ export interface StakePoolDetails {
...
+  id: Wallet.Cardano.Cip17Pool['id'];
+  targetWeight: Wallet.Cardano.Cip17Pool['weight'];
   displayData: Wallet.util.StakePool;
 };
 
-export type CurrentPortfolioStakePool = DraftPortfolioStakePool & {
+export type DraftPortfolioStakePool = DraftPortfolioStakePoolBase &
+  (
+    | {
+        basedOnCurrentPortfolio: false;
```
refi93, line 54: I considered having an optional `currentPortfolioRef: {percentage: Wallet.Percent}` property instead, but I found that to be more fragile as it's pretty easy to miss the removal/addition of such property when "detaching/ejecting" the portfolio draft from the current portfolio when user first edits it, whereas the boolean flag IMHO somewhat forces the writer of the code to "stop" and think what to set it to and configure the `DraftPortfolio` appropriately

packages/staking/tsconfig.json
```diff
@@ -12,6 +12,11 @@
     "outDir": "dist",
     "module": "es2020",
     "target": "es2016",
-    "preserveConstEnums": true
-  }
+    "preserveConstEnums": true,
+    "baseUrl": "./src",
+    "paths": {
+      "@assets/*": ["assets/*"]
```
refi93, line 18: this is crucial to finally be able to use assets "hidden" within the UI package which weren't accessible so far. This means we are able to use them without having to duplicate them which I leveraged e.g. for info-icon.component.svg in our banner

packages/ui/src/design-system/control-buttons/control-button.css.ts
```diff
@@ -15,6 +15,7 @@ export const container = recipe({
     border: vars.borders.$button_default,
     borderRadius: vars.radius.$medium,
     outline: 'none',
+    cursor: 'pointer',
```
refi93, line 18: not really related to the task itself, fixes the "Manage" button that until now didn't have the right cursor type on hover (default cursor remained)

packages/staking/src/features/overview/StakingNotificationBanner.tsx
```diff
@@ -0,0 +1,56 @@
...
+          message={t('overview.banners.pendingPoolMigration.title')}
+          description={t('overview.banners.pendingPoolMigration.message')}
+        />
+      );
+    case 'portfolioDrifted':
+      return (
+        <Banner
+          popupView={popupView}
+          withIcon
+          customIcon={<BellIcon className={styles.bannerInfoIcon} />}
```
refi93, line 47: note that in Figma the icon is aligned to the top in the popup view, but in the codebase are already other popup-view banners aligning the icon to the center so I decided to remain consistent with that -  for context see my comment raised in the figma design for the popup view

packages/staking/src/features/overview/OverviewPopup.tsx
```diff
@@ -81,12 +85,24 @@ export const OverviewPopup = () => {
 
   return (
     <>
+      {stakingNotification === 'portfolioDrifted' && (
+        <Box mb="$32">
+          <StakingNotificationBanner
+            notification="portfolioDrifted"
+            onPortfolioDriftedNotificationClick={expandStakingView}
```
refi93, line 92: as per the original task description this should have led to the management drawer but this would be a more complex change because it would require propagating a new kind of message between the popup view and expanded view and parsing it on the other end, made this redirection to the management drawer from popup view into a follow up task (LW-8543)

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -279,24 +279,31 @@ const mapStakePoolToPortfolioPool = ({
...
-  name: stakePool.metadata?.name,
-  ticker: stakePool.metadata?.ticker,
-  weight: 1,
+  targetWeight: 1,
 });
 
-const targetWeight = 100;
-const mapPoolWeights = (pools: DraftPortfolioStakePool[]) =>
-  pools.map<DraftPortfolioStakePool>((pool) => ({ ...pool, weight: Math.round(targetWeight / pools.length) }));
+const ejectDraftFromCurrentPortfolio = (
```
refi93, line 289: this is the function responsible from removing reference to the original "current portfolio" (i.e. the on-chain percentages), which as per the current logic is supposed to be triggered (manually) anytime we mutate the initial state of the draft portfolio, based off the current portfolio (it's idempontent, so no issue running this function also on already "ejected" drafts)

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/atomicStateMutators.ts
```diff
@@ -12,13 +12,15 @@ import {
 
 const missingDraftPortfolioErrorMessage = 'DelegationPortfolioState: Inconsistent state: missing draftPortfolio';
 
-const mapPoolWeights = (pools: DraftPortfolioStakePool[]) =>
-  pools.map<DraftPortfolioStakePool>((pool) => ({ ...pool, weight: Math.round(targetWeight / pools.length) }));
+const ejectDraftFromCurrentPortfolio = (
```
refi93, line 15: this is the function responsible from removing reference to the original "current portfolio" (i.e. the on-chain percentages), which as per the current logic is supposed to be triggered anytime we mutate the initial state of the draft portfolio, based off the current portfolio (it's idempontent, so no issue running this function also on already "ejected" drafts)

packages/staking/src/features/overview/helpers/getPortfolioTotalPercentageDrift.ts
```diff
@@ -0,0 +1,18 @@
...
+export const getPortfolioTotalPercentageDrift = (portfolio: CurrentPortfolioStakePool[]): number => {
+  const totalValue = Wallet.BigIntMath.sum(portfolio.map(({ value }) => value));
+  const totalWeight = sum(portfolio.map(({ targetWeight }) => targetWeight));
+
+  return sum(
+    portfolio.map(({ value, targetWeight }) => {
+      const targetPercentage = (targetWeight / totalWeight) * 100;
+      const currentPercentage = new BigNumber(value.toString()).div(totalValue.toString()).times(100).toNumber();
+      return Math.abs(targetPercentage - currentPercentage);
+    })
```
refi93, line 16: formula from the task description for reference:

```
Sum (n from 1 to Number_of_userPools) of Pool_N original % - Pool_N actual %)
```

of course, the differences need to all be made positive (i.e. use `Math.abs()`), otherwise they would even out

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/types.ts
```diff
@@ -4,17 +4,28 @@ import { Command } from './commands';
...
+type DraftPortfolioStakePoolBase = {
+  id: Wallet.Cardano.Cip17Pool['id'];
+  targetWeight: Wallet.Cardano.Cip17Pool['weight'];
   displayData: Wallet.util.StakePool;
 };
 
-export type CurrentPortfolioStakePool = DraftPortfolioStakePool & {
+export type DraftPortfolioStakePool = DraftPortfolioStakePoolBase &
+  (
+    | {
```
refi93, line 15: I considered having an optional currentPortfolioRef: {percentage: Wallet.Percent} property instead, but IMHO that would be more fragile as it IMHO makes it easier to miss the removal/addition of such property when "transforming" or "ejecting" (upon first editation) the portfolio draft from the current portfolio, whereas the basedOnCurrentPortfolio boolean flag IMHO somewhat forces the writer of the code to "stop" and think what to set it to and configure the DraftPortfolio appropriately

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/types.ts
```diff
@@ -4,17 +4,28 @@ import { Command } from './commands';
...
+  );
+
+export type CurrentPortfolioStakePool = DraftPortfolioStakePoolBase & {
   displayData: Wallet.util.StakePool & {
     lastReward: bigint;
     totalRewards: bigint;
   };
   stakePool: Wallet.Cardano.StakePool;
   value: bigint;
+  percentage: Wallet.Percent;
```
refi93, line 28: it's a fair point that it seems odd that here we express the target as a `targetWeight` and the current portfolio on-chain state as `percentages` - for now I deliberately chose not to try to unify these because for representing the target portfolio on-chain, the percentages don't allow for an unambiguous/lossless representation of them and I'd rather leave such unification (if at all needed) for later - I guess the path would become clear as soon as SDK is updated to persist/return the target portfolio and sliders are implemented in the management UI.

What I mean by "lossless" -> if e.g. we represent the target weight in the state as percentage and do some arithmetics (adding, multiplying, ...), we may end up having a target percentage `0.33......01` (because JS floats) and it may not be correctly translated to a weight in the (re)balancing tx/portfolio - the drift would be very slight, but I'd rather steer clear from any numerical instability if avoidable as it introduces its own class of edge-cases that are very easy to forget

On the other hand, we may choose to represent "percentages" as integers from 0 to 100, but that would lead to issues as soon as somebody chose to increase the precision shown in the UI (need to scale up all values), moreover it would not be consistent with other instances of percentages in the code where these are already encoded as decimal numbers from 0 to 1

So all in all, to avoid this "rabbit hole" I rather avoided having a single unit to represent target vs current distribution of the portfolio, and postpone that decision only after we implement more features/sdk implementation of portfolio persistence is available which should be soon anyway. The conversions are simple to handle, thinking about/dealing with the loss of precision is IMHO harder in general

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/types.ts
```diff
@@ -4,17 +4,28 @@ import { Command } from './commands';
...
+export type DraftPortfolioStakePool = DraftPortfolioStakePoolBase &
+  (
+    | {
+        basedOnCurrentPortfolio: false;
+      }
+    | { basedOnCurrentPortfolio: true; currentPortfolioPercentage: Wallet.Percent }
+  );
+
+export type CurrentPortfolioStakePool = DraftPortfolioStakePoolBase & {
   displayData: Wallet.util.StakePool & {
```
refi93, line 22: @szymonmaslowski made the follow up refactoring issue to de-ducplicate the structure of `Current/DraftPortfolioStakePool` structures, as discussed offline: https://input-output.atlassian.net/browse/LW-8580

packages/staking/src/features/drawer/PoolDetailsCard.tsx
```diff
@@ -39,7 +40,7 @@ export const PoolDetailsCard = ({ name, color, weight, onRemove }: PoolDetailsCa
         <Flex justifyContent="space-between" alignItems="center">
           <Text.Body.Normal weight="$semibold">
             {t('drawer.preferences.stakeValue', {
-              stakePercentage: weight,
+              stakePercentage: formatPercentages(percentage, { decimalPlaces: 0, rounding: 'down' }),
```
refi93, line 43: let's round up - should solve at least one class of edge-cases (e.g. `10.6 | 89.4`)

packages/staking/src/features/overview/OverviewPopup.tsx
szymonmaslowski: Was the redirection from popup to manage drawer an intention of the original task? I can see in the ACs that it is saying about redirecting to the browse pools, however that is incorrect I guess. IMO it should be Overview page instead, as this is about the current portfolio, so your implementation actually achieves the goal, so in this case the follow up task is probably not needed. Thought, please double check that with Fernando 😉

packages/staking/src/features/overview/helpers/getCurrentStakingNotification.ts
```diff
@@ -0,0 +1,26 @@
+import { CurrentPortfolioStakePool } from 'features/store';
```
szymonmaslowski, line 1: Please move this file along with the `StakingNotificationBanner` component to the dedicated `overview/StakingNotificationBanner` dir 🙏 

@edit: Actually please move all of the banner related functions to that dedicated dir 🙏

szymonmaslowski, line 1: Sorry, but IMO we shouldn't have aliases. I would preffer to have a relative paths ;)

packages/staking/src/features/overview/helpers/getCurrentStakingNotification.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { StakingNotificationType } from '../types';
+import type { AssetActivityListProps } from '@lace/core';
```
szymonmaslowski, line 3: This should got to the top of the imports list according to our eslint config. I wonder why it was not fixed automatically... 🤔

packages/staking/src/features/overview/types.ts
```diff
@@ -0,0 +1 @@
+export type StakingNotificationType = 'pendingFirstDelegation' | 'pendingPoolMigration' | 'portfolioDrifted';
```
szymonmaslowski, line 1: This one also should be moved to the banner dir IMO

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/types.ts
```diff
@@ -4,17 +4,28 @@ import { Command } from './commands';
 
 export type StakePoolWithLogo = Wallet.Cardano.StakePool & { logo?: string };
 
-export type DraftPortfolioStakePool = Wallet.Cardano.Cip17Pool & {
+type DraftPortfolioStakePoolBase = {
+  id: Wallet.Cardano.Cip17Pool['id'];
```
szymonmaslowski, line 8: Please consider also adding a `stakePool` property here. Maybe it will enable us to simplify implementation. For the first glance it seems better if current and draft types were more symetric

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -1,16 +1,10 @@
 import { Wallet } from '@lace/cardano';
 import { create } from 'zustand';
 import { immer } from 'zustand/middleware/immer';
-import { CARDANO_COIN_SYMBOL, LAST_STABLE_EPOCH, targetWeight } from './constants';
+import type { ExecuteCommand, Handler } from './stateMachine/types';
```
szymonmaslowski, line 4:
```suggestion
import type { ExecuteCommand, Handler } from './stateMachine';
```

However, I am curious was there somethis wrong with importing those types without the `type` import?

packages/staking/src/features/store/__snapshots__/delegationPortfolio.test.ts.snap
```diff
@@ -25,6 +25,7 @@ exports[`delegationPortfolioStore > sets the current portfolio 1`] = `
     },
     "id": "39deffa1dfcfe192ea0efeb3e9bcd9878190627fb590ec81f390cd6d",
     "name": "8BOOL",
+    "percentage": 0.33,
```
szymonmaslowski, line 28: It looks like this file is no longer used. Could you please remove it?

packages/staking/tsconfig.json
szymonmaslowski: I think it would be more correct to expose those icons instead of implicitly accessing them with ts magic (if I correctly understand the way it works). Sorry, but I am not a fan of aliases and would prefer to avoid them. Let's have a offline discussion about this change.

szymonmaslowski (approved):
Great improvements!

szymonmaslowski (changes requested):
Sorry, a missclick. I wanted to request changes instead of approving right away 😆

packages/staking/src/features/overview/helpers/getCurrentStakingNotification.ts
refi93: actually, as soon as I tried to put the type imports at the top, eslint warned me about an incorrect import order, I guess `features/store` is considered to be "external" by the eslint rule. Nevertheless, in this specific case a simple "fix" is to make the CurrentPortfolioStakePool import a type import

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
refi93: in this specific case there probably isn't much of a benefit, in general it helps the compiler as it is obvious such imports can be dropped, also runtime cyclic dependencies can be avoided this way, especially if some file with type definitions, but used solely for types happens to have a non-type import for whatever reason.

But I agree in this specific case it makes not that much sense, it was rather my "reflex" to isolate type imports, will merge it into a single one

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md

packages/staking/src/features/overview/helpers/getCurrentStakingNotification.ts
refi93: I see, VSCode intellisense actually suggested me that alias 🤔 do you perhaps know of some eslint rule/setting to disable these aliases, if we don't want to use them at all?

packages/common/src/ui/components/Banner/Banner.tsx
```diff
@@ -78,6 +86,11 @@ export const Banner = ({
             {buttonMessage && <Button onClick={onButtonClick}> {buttonMessage} </Button>}
           </div>
         )}
+        {!!onBannerClick && !onButtonClick && (
```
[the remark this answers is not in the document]
refi93: IMHO the banner and button click are two mutually exclusive behaviors and a banner having both would not be good UX, for which we moreover don't have designs. So actually I'd rather go with a discriminating union to prevent accidentally combining both callbacks which doesn't have well defined behavior/UX as of now

refi93:
> Was the redirection from popup to manage drawer an intention of the original task? I can see in the ACs that it is saying about redirecting to the browse pools, however that is incorrect I guess. IMO it should be Overview page instead, as this is about the current portfolio, so your implementation actually achieves the goal, so in this case the follow up task is probably not needed. Thought, please double check that with Fernando 😉

The task originally stated that both for the expanded and popup view the "drift warning" banner should lead to the management drawer. After I raised the need for implementing additional messaging for the popup view with Fernando, we agreed to update the ACs to take the user to the expanded view and I'd say "Browse pools" was introduced there by mistake, should have been Overview actually - I'm clarifying this with him right now

Still, I made a low-priority follow-up task to take the user straight to the management drawer as I believe that to be superior UX, just not worth the additional effort right now

packages/staking/tsconfig.json
refi93: sure, let's discuss it - I dropped the alias, though I'm curious - what issues do you see with aliases? At first sight, I'd say it makes imports neater and somewhat prevents incorrect imports, especially if two files with the same name happen to be present in the project, but TBH, I haven't used them yet, so I'm probably just not aware of the caveats :)

refi93:
@szymonmaslowski ACs updated already to state that Popup view banner should take the user to "Overview" 👍 thanks for the catch

szymonmaslowski (approved):

oldGreg5 (approved):

PR #539, opened 2023-09-18
feat(staking): LW-8422 refactor portfolio store to state machine

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-8422)
- [ ] Screenshots added.

---

## Proposed solution

**Goal: Refactoring from having the app state changes spread across all UI components to the state-machine-like centralised management.**

The zustand based `useDelegationPortfolioStore` exposes now a few setters and a `executeCommand` mutator. That mutator runs the `processPopupViewCases` or `processExpandedViewCases` function depending on the context.

Those functions implement a tree-like structure constructed using three types of data:
 - top level `Flow`s where each Flow represents app user flows like viewing pool details or staking flow
 - middle level `Step`s where each step represents a drawer step within a Flow.
 - bottom level `Command`s representing an "intention" and being emited from the UI.

Thanks to state-machine-like design we define which `Command`s are allowed in which `Flow`s and `Step`s so every user flow is well designed and defined in a central place.

We achieved a type safety with the level of the commands belonging to particular flow/steps - TS complains if we are missing a command handler for the particular state.

Further improvements:
 - Split into multiple files
 - Express state with discriminated union
 - Type safety improvements:
   - Use state discriminated union to validate state mutations. Possibly we could enhance the `handler` function (which is now responsible just for narrowing down the command data type) with a `CurrentState` and `TargetState` type parameters so it would then validate use while changing the state.
   - If possible make TS complaining when you try to use a command that is not defined in a given Step/Flow.
 - move more things from the UI to be controlled by the state machine:
   - submitting the transaction and handling a result
   - constructing the `txBuilder` object
   - etc.
 - minor usage improvements like:
   - passing a string to the `executeCommand` if particular command does not transfer any data (which is a majority of cases) so we don't have to pass a object

Additional improvements that has been made:
 - Reduced coupling with `outsideHandles`
 - Since we touched this part, we fixed an existing bug in prod in `fix(staking): clear password on stake tx failure`

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2155/6271878369/index.html) for [e279a944](https://github.com/input-output-hk/lace/pull/539/commits/e279a944b1670b355718d881eec05e5c3b5ec7f5)


packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,231 @@
...
+
+// Just a wrapper for a simple handler function but narrowing down the data type
+// so we are able to hint TS that for a given command that is the type of passed data
+const handler =
+  <D = any>(handlerBody: Handler<D>): Handler<D> =>
+  (params) =>
+    handlerBody(params);
+
+const helpers = {
+  selectPool: ({ pool, state }: { pool: DraftPortfolioStakePool; state: State }) => {
```
danielmain, line 110: I would not return anything because the function is returning `void`

So I would push only if `!selectionsFull && !alreadySelected`

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,231 @@
...
+  showPoolDetails: ({
+    pool,
+    state,
+    targetMode,
+  }: {
+    pool: Wallet.Cardano.StakePool;
+    state: State;
+    targetMode: Mode;
+  }) => {
+    state.mode = targetMode;
```
danielmain, line 125: Could we keep the helpers pure? And let the `executeCommand` modify the state?

danielmain (commented):
Just left some non-blocking comments

packages/staking/src/features/store/useDelegationPortfolioStore.ts
szymonmaslowski: I guess current implementation is the same. Just a bare return statement is like exiting a function with returning nothing, isn't it?

I considered doing like you suggested in case of potential readability improvement but actually IMO having a condition `if (selectionsFull || alreadySelected)` is easier to reason about than `if (!selectionsFull && !alreadySelected)` so in the end current implementation is simpler IMO. I am eager to change it though if you thing otherwise 😉

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,231 @@
...
+  PoolDetails = 'PoolDetails',
+  CurrentPortfolioManagement = 'CurrentPortfolioManagement',
+  NewPortfolioCreation = 'NewPortfolioCreation',
+}
+
+enum DrawerDefaultSection {
+  DETAIL = 'detail',
+}
+
+enum DrawerManagementSection {
```
refi93, line 21: I'd consider naming this "step" instead to make the sequentiality of these obvious to the reader at first sight

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,231 @@
...
+  ShowDetails = 'ShowDetails',
+}
+
+enum PoolDetailsCommand {
+  Select = 'Select',
+}
+
+enum DrawerPreferencesCommand {
+  UpdateWeight = 'UpdateWeight',
+}
```
refi93, line 48: I'd consider coupling the commands with the expected data and making it a union of objects, otherwise it gets pretty tricky for the reader to understand what are the commands parametrized by which I believe is important to get the bigger picture

```suggestion
type DrawerPreferencesCommand = {
  name: 'UpdateWeight',
  params: { weight: number }
} | ...
```

I believe this should also make it pretty straightforward to come up with the "command" callback signature below

```
// WIP: trying to match command to its corresponding data/payload

```

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,231 @@
+import { DelegatedStake } from '@cardano-sdk/wallet';
+import { Wallet } from '@lace/cardano';
+import { create } from 'zustand';
+import { immer } from 'zustand/middleware/immer';
+import { MAX_POOLS_COUNT } from './delegationPortfolio';
+import { CurrentPortfolioStakePool, DraftPortfolioStakePool } from './types';
+
+enum Mode {
```
refi93, line 8: I'd consider renaming `Mode -> Section` to `Flow -> Step` which would imho be more descriptive to the reader at first sight, if I understood correctly what we are trying to capture here -> "Multistep forms, with the special case of single-step "forms" with no actions"

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,231 @@
...
+
+enum PoolDetailsCommand {
+  Select = 'Select',
+}
+
+enum DrawerPreferencesCommand {
+  UpdateWeight = 'UpdateWeight',
+}
+
+type Command = OverviewCommand | BrowsePoolsCommand | PoolDetailsCommand | DrawerPreferencesCommand;
```
refi93, line 50: I'd consider making this an interface and be explicit in the commands that these are implementing this interface, which would then have the benefit of ensuring consistency of the structure and should in theoery allow even for extracting the individual Commands into separate files and generalize the handler in a way that adding a new Command wouldn't force the generic logic to be updated

```suggestion
interface Command = { name: string, params: unknown };
```

packages/staking/src/features/store/useDelegationPortfolioStore.ts
szymonmaslowski: Totally agree 👍

packages/staking/src/features/store/useDelegationPortfolioStore.ts
refi93: note, actually an ordinary `type` should do, an interface may mislead the reader into thinking that classes are involved which isn't the case

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+  ? DraftPortfolioStakePool
+  : C extends PoolDetailsCommand.Select
+  ? DraftPortfolioStakePool
+  : C extends DrawerPreferencesCommand.UpdateWeight
+  ? { poolId: Wallet.Cardano.PoolIdHex; weight: number }
+  : undefined;
+
+type State = {
+  currentPortfolio: CurrentPortfolioStakePool[];
+  draftPortfolio: DraftPortfolioStakePool[];
```
refi93, line 66: Could you clarify what's the difference between the "draft portfolio" and "selection"? From this file alone/code I briefly looked up it wasn't obvious to me and I'm wondering whether these couldn't somehow be merged or at least their relation be made clearer by some more descriptive name

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+  Select = 'Select',
+}
+
+enum DrawerPreferencesCommand {
+  UpdateWeight = 'UpdateWeight',
+}
+
+type Command = OverviewCommand | BrowsePoolsCommand | PoolDetailsCommand | DrawerPreferencesCommand;
+
+type DataOfCommand<C extends Command> = C extends OverviewCommand.ShowDetails
```
refi93, line 52: I'd consider making a similar `SectionOfMode` (`StepOfFlow`) type and leverage it in the state to ensure consistency between the Flow and Step in the state which I'd find valuable not only from readability perspective but also as one more safeguard against bugs

packages/staking/src/features/store/useDelegationPortfolioStore.ts
refi93: I see pretty much the same question was answered on Slack, pasting your answer here @szymonmaslowski 

> draftPortfolio is there because we have two different flows (current pools management and new portfolio creation) used by the same UI. When we initiate one of those flows we copy either currentPortfolio or selections to the draftPortfolio and the UI (drawer) reads from single sorce. This also allows us to have a copy of the original state (currentPortfolio or selections) so we can modify the copy preserving the original state untouched. Consider the scenario: we have some selections but instead of submitting them we go to overview and click manage button so we enter the management of the current portfolio but thanks to copying the currentPortfolio to draft we preserve selections untouched, so user can cancel managemet of current portfolio and continue with selections instead.

refi93: what would perhaps help me get this nuance at first sight would be if `selections` was called `browsePoolsSelections`

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+  BrowsePools = 'BrowsePools',
+  CurrentPoolDetails = 'CurrentPoolDetails',
+  PoolDetails = 'PoolDetails',
+  CurrentPortfolioManagement = 'CurrentPortfolioManagement',
+  NewPortfolioCreation = 'NewPortfolioCreation',
+}
+
+enum DrawerDefaultSection {
+  DETAIL = 'detail',
+}
```
przemyslaw-wlodek, line 19: Shouldn't this be a simple type? & Nit: please name enum key/value with same casing `detail = 'detail'` for consistency and to allow easy "mirror-like" access to the value (which usually helps in type-guards) etc.

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+  ? { poolId: Wallet.Cardano.PoolIdHex; weight: number }
+  : undefined;
+
+type State = {
+  currentPortfolio: CurrentPortfolioStakePool[];
+  draftPortfolio: DraftPortfolioStakePool[];
+  mode: Mode;
+  section?: Section;
+  selections: DraftPortfolioStakePool[];
+  viewedStakePool?: Wallet.Cardano.StakePool;
```
przemyslaw-wlodek, line 70: I feel weird about having implicit payloads for some of the Actions/Commands. Here it's purely `payload` of a `BrowsePoolsCommand.ShowDetails` but on the other hand e.g. `selections` and `draftPoftfolio` are the payloads of other commands. Maybe it's okay but I wanted to say if out loud so we can brainstorm it.

I think staying with this approach but introducing change that @refi93 and I proposed [here](https://github.com/input-output-hk/lace/pull/539/files#r1329889177) we can have something like `viewedStakePool?: BrowsePoolsCommand.ShowDetails.payload`

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+
+type DelegationPortfolioStore = State & {
+  mutators: {
+    command: <C extends Command>(...params: DataOfCommand<C> extends undefined ? [C] : [C, DataOfCommand<C>]) => void;
+    setCurrentPortfolio: (params: {
+      delegationDistribution: DelegatedStake[];
+      cardanoCoin: Wallet.CoinId;
+      currentEpoch: Wallet.EpochInfo;
+      delegationRewardsHistory: Wallet.RewardsHistory;
+    }) => Promise<void>;
```
przemyslaw-wlodek, line 81: Can this be also a command with payload?

Then we can have instead of `mutators` -> `actions` and everything is cleaner.

If we still want to separate "flows" from this setter then let's rename `command` as @refi93 proposed.

packages/staking/src/features/store/useDelegationPortfolioStore.ts
szymonmaslowski: How about `selectedPortfolio` to have consistency with other two properties?

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+    handler(params);
+  };
+
+// Just a wrapper for a simple handler function but narrowing down the data type
+// so we are able to hint TS that for a given command that is the type of passed data
+const handler =
+  <C extends Command>(handlerBody: Handler<C>): Handler<C> =>
+  (params) =>
+    handlerBody(params);
```
przemyslaw-wlodek, line 120: A lot of these types will be removed/simplified if we introduce other proposed changes. So I'm not digging into it right now.

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,240 @@
...
+          [OverviewCommand.Manage]: ({ store }) => {
+            store.mode = Mode.CurrentPortfolioManagement;
+            store.section = DrawerManagementSection.PREFERENCES;
+            store.draftPortfolio = store.currentPortfolio;
+          },
+          [OverviewCommand.GoToBrowsePools]: ({ store }) => {
+            store.mode = Mode.BrowsePools;
+          },
+        },
+        params.command as OverviewCommand,
```
przemyslaw-wlodek, line 161: Can we please utilise some of the proposed libraries e.g. for the pattern matching or state machines? I know it's a lot of great work but in general I feel bad that we are reinventing the wheel and it's not fully type-safe etc.

I feel that our own abstraction may not be future-proof.

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,833 @@
...
+  | CommandPoolDetailsSelectPool
+  | CommandPoolDetailsUnselectPool
+  | CommandPoolDetailsBeginSingleStaking;
+
+type CurrentPortfolioManagementStepPreferencesCommand =
+  | CommandCommonCancelDrawer
+  | CommandCommonDrawerContinue
+  | CommandCommonPreferencesStepAddStakePools
+  | CommandCommonPreferencesStepRemoveStakePool
+  | CommandCommonPreferencesStepUpdateWeight;
```
refi93, line 158: did I understand correctly that this `CommandCommonPreferencesStepUpdateWeight` command is a preparation for the introduction of sliders and currently it isn't really used, or does it already have a purpose?

refi93:
"build staking center" job is failing in CI with some TS errors, needs to be fixed before merging

Screenshot 2023-09-21 at 23 57 58

packages/staking/src/features/store/types.ts
refi93, line 1: I'd consider keeping a separate file for types, at least the ones not directly related to useDelegationPortfolioStore, such as `AdaSymbol`, `DraftPortfolioStakePool` / `CurrentPortdolioStakePool` - decouples the hook from more general types and another benefit of a separate file for types is prevention of potential runtime/compilation issues with cyclic imports because `type` imports are excluded from the built js

packages/common/src/ui/lib/__tests__/format-number.test.ts
```diff
@@ -20,3 +20,40 @@ describe('Number formatters', () => {
...
 });
+
+describe('getNumberWithUnit', () => {
+  test('formats a number rounding up to 2 decimal places according to its unit', () => {
+    expect(getNumberWithUnit('10234')).toEqual({ number: '10.23', unit: 'K' });
+    expect(getNumberWithUnit('10235')).toEqual({ number: '10.24', unit: 'K' });
+    expect(getNumberWithUnit('10235000')).toEqual({ number: '10.24', unit: 'M' });
+    expect(getNumberWithUnit('10235000000')).toEqual({ number: '10.24', unit: 'B' });
+    expect(getNumberWithUnit('10235000000000')).toEqual({ number: '10.24', unit: 'T' });
+    expect(getNumberWithUnit('10235000000000000')).toEqual({ number: '10.24', unit: 'Q' });
```
mrcnk, line 31: I think this is precisely what Intl.NumberFormat is supposed to do 🤔 do we know why we created whole custom formatter (not i18n ready) for that matter? (what a rhyme lol)

packages/staking/src/features/browse-pools/BrowsePools.tsx
```diff
@@ -1,62 +1,52 @@
...
-    <>
-      <PortfolioBar onStake={onStake} />
-      <StakePoolsTable scrollableTargetId={LACE_APP_ID} onStake={onStake} />
-    </>
-  );
-};
+export const BrowsePools = () => (
+  // const { setIsDrawerVisible } = useStakePoolDetails();
+
+  // const {
```
mrcnk, line 9: Is it left intentionally or should it be removed?

packages/staking/src/features/browse-pools/PortfolioBar.tsx
```diff
@@ -27,13 +24,13 @@ export const PortfolioBar = ({ onStake }: PortfolioBarParams) => {
       <Flex className={styles.buttons}>
         <Button.Secondary
           label={t('portfolioBar.clear')}
-          onClick={portfolioMutators.clearSelections}
+          onClick={() => portfolioMutators.executeCommand({ type: 'ClearSelections' })}
```
mrcnk, line 27: ❤️

packages/staking/src/features/store/types.ts
szymonmaslowski: Yep, we need to have a discussion anyway about structuring this code. I would even have a few files just for types like having commands separately for example.

mrcnk (changes requested):
Great job, just please make sure all the commented out code sections are there to be left or cleaned up

packages/staking/src/features/store/useDelegationPortfolioStore.ts
szymonmaslowski: I removed it actually. It was not used currently. Let's introduce it back once we have a sliders ;)

packages/staking/src/features/browse-pools/BrowsePools.tsx
szymonmaslowski: We realized that whole of this commented code is related to the hardware wallets. We decided to comment it out but maybe we should just remove it for now and re-introduce once we are ready for hardware wallets. We ignored the HW implementation in the PR as anyway we are not able to check if it would work correctly.

packages/common/src/ui/lib/__tests__/format-number.test.ts
refi93: Intl.NumberFormat afaik works with native JS numbers which however lose precision beyond ~2^53 which is less than some token amounts may be, so I guess that's the main reason for such custom utility function

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

packages/staking/src/features/store/useDelegationPortfolioStore.ts
```diff
@@ -0,0 +1,811 @@
...
+      ),
+    },
+    params.store.activeFlow,
+    'root'
+  )(params);
+
+const defaultState: State = {
+  activeDrawerStep: undefined,
+  activeFlow: Flow.Overview,
+  cardanoCoinSymbol: 'ADA',
```
refi93, line 676: IMHO the cardano coins symbol shouldn't be part of the drawer state nor the "display data" of stakepools, I see that rather as a global context that should rather be retrieved only when actualy rendering ADA values in components. Not worth it changing this now, but if you agree, I'd suggest making it a follow-up refactoring task

refi93 (approved):

przemyslaw-wlodek (approved):
🚀 💜 🚀

mrcnk (approved):

PR #547, opened 2023-09-19
DRAFT Feat/lw 8422 move drawer state into portfolio store

xdzurman:
DRAFT


packages/staking/.eslintrc.cjs
```diff
@@ -48,5 +48,6 @@ module.exports = {
     'sort-keys/sort-keys-fix': ['error', 'asc', { natural: true }],
     'unicorn/no-null': 'off',
     'unicorn/prefer-query-selector': 'off',
+    'unicorn/no-object-as-default-parameter': 'off'
```
refi93, line 51: Is it really needed to turn off this rule for the whole workspace? I'd consider turning it off in the presumably few places/files as it seems IMHO beneficial to have it enforced by default

PR #581, opened 2023-09-28
fix(staking): lw-8572 broken portfolio bar on narrow screen

greatertomi:
# Checklist

- [x] JIRA - [LW-8572]
- [ ] Screenshots added.

---

## Proposed solution

Add `minWidth` to the portfolio bar

#### Limitations
- The page distorts at <= 360px. The issue comes from the design library. I don't know if there is value in resolving this.
- The horizontal scroll for the page does not work with the portfolio bar.

## Testing

## Screenshots

---
# Allure report

**smokeTests**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2367/6417169010/index.html) for [e1cde16b](https://github.com/input-output-hk/lace/pull/581/commits/e1cde16b00a6d15f4499ad25e5554774d950fbd0)

[LW-8572]: https://input-output.atlassian.net/browse/LW-8572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


refi93 (approved):

packages/staking/src/features/browse-pools/StakePoolsTable/StakePoolsTable.module.scss
```diff
@@ -1,28 +1,10 @@
 @import '../../../../../common/src/ui/styles/theme.scss';
 @import '../../../../../common/src/ui/styles/abstracts/typography';
 
-.table {
-  display: flex;
-  flex-direction: column;
-  padding: size_unit(4) 0;
-  min-height: 990px;
-  .header {
+.stakePoolTable {
```
refi93, line 4: naming nitpick for the sake of consistency
```suggestion
.stakePoolsTable {
```

packages/staking/src/features/browse-pools/PortfolioBar.css.ts
```diff
@@ -1,4 +1,5 @@
 import { style, sx } from '@lace/ui';
+const minimumScreenSize = '668px';
```
szymonmaslowski, line 2: How about exporting all the breakpoint thresholds from the `@lace/common` and importing the value here?
I would suggest to duplicate all of [those](https://github.com/input-output-hk/lace/blob/4bb9a74380553a73dea5958ea32e0109894a07a7/packages/common/src/ui/styles/theme.scss#L11) variables in a ts file and export it from the lace package. My suggestion is to create a `theme.css.ts` or `theme.ts` next to the `theme.scss`. It would be also good to place a comment in both files (ts and scss) hinting that those variables exist in two places so in case someone wants to change them he has to do that in both places.

szymonmaslowski: Actually, the best approach would be probably to have it defined in the UI package, but this would probably require further discussion.

@przemyslaw-wlodek are you aware of any approach of the toolkit to the breakpoints?

packages/staking/src/features/browse-pools/PortfolioBar.css.ts
```diff
@@ -8,6 +9,11 @@ export const barContainer = style([
     right: '$24',
   }),
   {
+    '@media': {
+      [`screen and (max-width: ${minimumScreenSize})`]: {
+        minWidth: '480px',
```
szymonmaslowski, line 14:
```suggestion
        minWidth: theme.spacing.$480,
```

szymonmaslowski (commented):
Looks good in general 💪 I just left some considerations there. Could you please also add a screenshot of the changed UI?

packages/staking/src/features/browse-pools/PortfolioBar.css.ts
przemyslaw-wlodek: Here are the breakpoints which we support* https://github.com/input-output-hk/lace/blob/main/packages/ui/src/design-tokens/sx.css.ts#L5

Please follow them in the staking package :)

*Please double-check with Allan what's recommended.

greatertomi: There are more breakpoints in the `common` package [here](https://github.com/input-output-hk/lace/blob/4bb9a74380553a73dea5958ea32e0109894a07a7/packages/common/src/ui/styles/theme.scss#L11-L18) that is not reflected in the `ui-toolkit`. I think @szymonmaslowski suggestion of adding the breakpoints to the ui-toolkit is a very good idea.

packages/staking/src/features/browse-pools/StakePoolsTable/StakePoolsTable.module.scss
greatertomi: Fixed on d5ccbf4

packages/staking/src/features/browse-pools/PortfolioBar.css.ts
greatertomi: Fixed on d5ccbf4

packages/staking/src/features/browse-pools/PortfolioBar.css.ts
szymonmaslowski: I am not sure which one is better. In order to use breakpoints from the toolkit we need the whole Lece switched to them. I doubt it is already done. Maybe we could just duplicate the sizes in the common package and in the future switch the whole lace to use the new breakpoints.

przemyslaw-wlodek: Please consult this with Allan Leone but ui-toolkit should be the single source of truth and up to date.

szymonmaslowski: I would personally delay that. It is not always easy to switch to the toolkit. It is easier when building new components and just using the building blocks from the toolkit. It is harder when it comes to broader topics like breakpoints which should be consistent across the whole app. I wouldn't also be discussing that with Alan in the scope of this task. I think that the discussion about breakpoints should happen independently of any of our tasks, and should be rather a cross-team initiative.

emiride (approved):

przemyslaw-wlodek (approved):

PR #590, opened 2023-09-29
feat(staking: [LW-8510] init draft sliders to on chain percentages

refi93:
# Checklist

- [x] JIRA - LW-8510
- [ ] Proper tests implemented
- [ ] Screenshots added.

---

## Proposed solution

As agreed offline, let's initialize the slider values to on-chain current portfolio percentages if managing existing portfolio

## Testing

Try managing existing portfolio and verify that the initial state of the slider matches on-chain percentages

## Screenshots

Attach screenshots here if implementation involves some UI changes

---
# Allure report
`allure-report-publisher` generated test report!

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2290/6354759147/index.html) for [1d0f47cd](https://github.com/input-output-hk/lace/pull/590/commits/1d0f47cdf2feb3b95ab786408da635ac342c7cd0)
|       | passed | failed | skipped | flaky | total | result |
|-------|--------|--------|---------|-------|-------|--------|
| Total | 30     | 2      | 0       | 0     | 32    | ❌     |

PR #593, opened 2023-10-02
refactor(staking): [LW-8510] refactor percentage slider

refi93:
# Checklist

- [x] JIRA - LW-8510
- [ ] Screenshots added.

---

## Proposed solution

amend hotfix from #590 that fixed the infinite loop of the percentage slider, however it was at the cost of detaching the components from the `onPercentageChange` callback changes which makes it fragile/violates eslint rule on `useEffect()` hook exhaustive dependencies. This PR fixes that flaw by keeping the ref to the current callback which is updated anytime the callback is modified from above, without triggering any undesired `onPercentageChange()` callback invocations

On top of that, this PR gives more specific names to percentage-related props in the same component, suffixing them all consistently with the `percentage` suffix, improving overall readability of the code

## Testing

This PR should bring no noticeable change to the sliders' behavior, it's just a refactor. In order to test, open management drawer and try changing the slider values, check that after closing and reopening it still works as expected

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2301/6379057980/index.html) for [0cf98d3e](https://github.com/input-output-hk/lace/pull/593/commits/0cf98d3e59f3776be83414d2fec604b577f39a1f)


packages/staking/src/features/drawer/preferences/PoolDetailsCard/PoolDetailsCard.tsx
```diff
@@ -37,24 +37,28 @@ export const PoolDetailsCard = ({
...
+  const firstRender = useRef(true);
   const { t } = useTranslation();
-  const [localValue, setLocalValue] = useState(targetRatio);
+  const [percentage, setPercentage] = useState(targetPercentage);
+  const onPercentageChangeRef = useRef(onPercentageChange);
 
   useEffect(() => {
-    if (!effectInitialized.current) {
-      effectInitialized.current = true;
+    onPercentageChangeRef.current = onPercentageChange;
```
refi93, line 52: the main fix in this PR, took inspiration from the slider implementation in the lib

[state management hook used in the slider](https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/slider/src/Slider.tsx#L102)
[the implementation of it](https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/use-controllable-state/src/useControllableState.tsx#L12)
[the ref to the callback](https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/use-callback-ref/src/useCallbackRef.tsx#L7)

packages/staking/src/features/drawer/preferences/PoolDetailsCard/PoolDetailsCard.tsx
```diff
@@ -37,24 +37,28 @@ export const PoolDetailsCard = ({
...
+  }, [onPercentageChange]);
+
+  useEffect(() => {
+    if (firstRender.current) {
+      firstRender.current = false;
       return;
     }
-    onPercentageChange(localValue);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [localValue]);
```
refi93, line 57: actually, even the previous solution should work properly in the sense that if the value of the slider changed, the latest `onPercentageChange` passed from above would be used, but it's not very clean/intuitive that the `onPercentageChange` callback is left from the hook dependencies

szymonmaslowski (changes requested):

packages/staking/src/features/drawer/preferences/PoolDetailsCard/PoolDetailsCard.tsx
szymonmaslowski: I just realized we don't need `useEffect` at all. We removed a debounce so now we can just assign the `onPercentageChange` directly to the  Slider's `onValueChange`.

refi93: > I just realized we don't need useEffect at all. We removed a debounce so now we can just assign the onPercentageChange directly to the Slider's onValueChange.

indeed 🤦  good catch 👍

refi93: done in https://github.com/input-output-hk/lace/pull/593/commits/6a208d05973b38626059d3c2921b42999ea606ca

szymonmaslowski (approved):

przemyslaw-wlodek (approved):

oldGreg5 (approved):

PR #601, opened 2023-10-03
fix(staking): [LW-8685] preserve pool onchain/saved percentages upon SM transitions

xdzurman:
# Checklist

- [x] [JIRA - \<link>](https://input-output.atlassian.net/browse/LW-8685?atlOrigin=eyJpIjoiMzIyNTY1NjM0YjljNDYwZTkyZTQ3NGFmZDYzOTMyYmMiLCJwIjoiaiJ9)
- [x] Screenshots added.

---

## Proposed solution

Remove notion of `basedOnCurrentPortfolio`. To preserve the current portfolio's percentages (onchain, saved), these percentages need to live on `DraftPortfolioPool` somehow as well. Therefore, I made these percentages required on `CurrentPortfolioPool` type, and optional on `DraftPortfolioPool`. When adding a new pool (from browsePools, or stake-all-on-this-pool), there's a check if this pool resides in the currentPortfolio and if so, saved and onchain percentages are copied from there to the new pool. This allows the user to see these percentages in all places (adding pool to current portfolio / creating new portfolio that might have some current pool) if they're defined.

## Testing

Make sure that saved/onchain percentages of current pools are never lost throughout the app, even when creating a new portfolio and the draft pools happen to also be in the current portfolio.

## Video
1. current portfolio > add pool > see that saved and onchain percentages are not lost
2. new portfolio > see that every "current portfolio pool" added to new portfolio retains saved/onchain percentages
3. stake-all-on-this-pool on current portfolio pool retains   saved/onchain percentages

https://github.com/input-output-hk/lace/assets/22454337/a034e31e-c791-450f-88f6-9e7a2cf49b02

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2352/6405410358/index.html) for [47851d1b](https://github.com/input-output-hk/lace/pull/601/commits/47851d1b59ca732522fae48e5c4d60997fa84a55)


packages/staking/src/features/drawer/StakePoolDetails.tsx
```diff
@@ -64,7 +64,11 @@ export const StakePoolDetails = ({
   } = useDelegationPortfolioStore((store) => ({
     activeDrawerStep: store.activeDrawerStep,
     activeFlow: store.activeFlow,
-    currentPortfolioDraftModified: store.draftPortfolio?.some((pool) => !pool.basedOnCurrentPortfolio) || false,
+    currentPortfolioDraftModified:
+      store.draftPortfolio?.some(
+        ({ sliderIntegerPercentage, savedIntegerPercentage }) =>
+          !!savedIntegerPercentage && sliderIntegerPercentage !== savedIntegerPercentage
+      ) || false,
```
refi93, line 71: my understanding is that this flag is supposed indicate that the portfolio draft has changed respective to the current portfolio (at least that was my understanding at the time of introducing this flag and that's how it's used e.g. for the sake of the dynamic button text). I think the current logic may not always work that way, namely in these two cases:

1. User removes a pool from the current portfolio -> all remaining pools would have the same percentages, making this check evaluate incorrectly to `false`
2. When working with selected pools not included in the current portfolio, these don't have any `sliderIntegerPercentage` making this condition evaluate to `false` as well, despite the portfolio being obviously different from the `current portfolio`

If I'm correct, I'd suggest for the sake of clarity and simplicity reversing the flag (renaming it to `portfolioDraftMatchesCurrentPortfolio`) and adding a few conditions which should ensure that the flag is checking 1:1 match with the current portfolio:

```suggestion
    portfolioDraftMatchesCurrentPortfolio:
      store.draftPortfolio.length === store.currentPortfolio?.length
      && (store.draftPortfolio || []).every(
        ({ sliderIntegerPercentage, savedIntegerPercentage }) =>
          !!savedIntegerPercentage && sliderIntegerPercentage !== savedIntegerPercentage
      ),
```

refi93 (changes requested):
see my comment regarding detection of portfolio modification, which I believe to not match the desired UX (e.g. confirm button (even if disabled) not visible if pool removed when managing current portfolio)

refi93 (approved):

packages/staking/src/features/drawer/StakePoolDetails.tsx
```diff
@@ -100,7 +105,11 @@ export const StakePoolDetails = ({
         return null;
       })(),
       [DrawerManagementStep.Preferences]: (() => {
-        if (activeFlow === Flow.PortfolioManagement && !currentPortfolioDraftModified && !currentPortfolioDrifted) {
+        if (
+          activeFlow === Flow.PortfolioManagement &&
+          portfolioDraftMatchesCurrentPortfolio &&
+          !currentPortfolioDrifted
+        ) {
```
szymonmaslowski, line 112: Can we move this whole condition to the selector?

packages/staking/src/features/drawer/StakePoolDetails.tsx
```diff
@@ -57,19 +57,24 @@ export const StakePoolDetails = ({
...
     draftPortfolioValidity: getDraftPortfolioValidity(store),
     openPoolIsSelected: store.selectedPortfolio.some(
       (pool) => store.viewedStakePool && pool.id === store.viewedStakePool.hexId
     ),
+    portfolioDraftMatchesCurrentPortfolio:
+      store.draftPortfolio?.length === store.currentPortfolio?.length &&
+      (store.draftPortfolio || []).every(
+        ({ sliderIntegerPercentage, savedIntegerPercentage }) =>
+          !!savedIntegerPercentage && sliderIntegerPercentage === savedIntegerPercentage
+      ),
```
szymonmaslowski, line 77: IMO it would be better to have a boolean `footerHiddenInPreferencesStep` and check if it is a management flow. Right now you always compare the draft against the current, but that check is invalid in the new portfolio flow. So by checking all directly here we eliminate this problem

packages/staking/src/features/drawer/StakePoolDetails.tsx
```diff
@@ -110,7 +119,9 @@ export const StakePoolDetails = ({
         return (
           <StepPreferencesFooter
             buttonTitle={
-              activeFlow === Flow.PortfolioManagement && !currentPortfolioDraftModified && currentPortfolioDrifted
+              activeFlow === Flow.PortfolioManagement &&
+              portfolioDraftMatchesCurrentPortfolio &&
+              currentPortfolioDrifted
```
szymonmaslowski, line 124: I guess this also could be moved to the selector

packages/staking/src/features/drawer/preferences/StepPreferencesContent.tsx
```diff
@@ -57,10 +57,9 @@ export const StepPreferencesContent = () => {
       color: PIE_CHART_DEFAULT_COLOR_SET[i] as PieChartColor,
       id,
       name: name || '-',
-      onChainPercentage: draftPool.basedOnCurrentPortfolio ? draftPool.onChainPercentage : undefined,
+      onChainPercentage: draftPool?.onChainPercentage,
```
szymonmaslowski, line 60: ❤️

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/initializeDraftPortfolioPool.ts
```diff
@@ -0,0 +1,23 @@
+import { Wallet } from '@lace/cardano';
+import { mapStakePoolToPortfolioPool } from './mapStakePoolToPortfolioPool';
```
szymonmaslowski, line 2: How about moving the `mapStakePoolToPortfolioPool` into this file as a local helper?

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/atomicStateMutators.ts
```diff
@@ -50,22 +45,15 @@ export const atomicStateMutators = {
...
-      {
-        ...mapStakePoolToPortfolioPool({
-          cardanoCoinSymbol: state.cardanoCoinSymbol,
-          sliderIntegerPercentage: 0,
-          stakePool,
-        }),
-      },
-      ...state.selectedPortfolio,
-    ]);
+    const newPool = initializeDraftPortfolioPool({ initialPercentage: 0, stakePool, state });
```
szymonmaslowski, line 54: This means values will be populated even if we make selections just going to browse pools manually... Is this a requirement? I lost track already...

szymonmaslowski (changes requested):
Great Improvement! Left a few comments

packages/staking/src/features/drawer/StakePoolDetails.tsx
refi93: IMHO I'd rather leave in the state/selectors UI-agnostic facts and the logic deciding whether something should be shown or not should be as close to the relevant place as possible, especially in components like this where multiple flows are mixed, so having UI logic at the top-level could make it pretty tricky for the reader to figure out which flow is it bound to, making the code harder to follow than it needs to be

So IMHO the `portfolioDraftMatchesCurrentPortfolio` variable (or some equivalent fact-based flag) should actually be left here (the fact whether the draft matches current portfolio makes sense regardless of the flow) and I'd keep the check whether a footer should be shown or not rather right within the footers map to not clutter the top-level scope of this "master" component

szymonmaslowski: Right, I agree with you :wink: I initially thought comparing the draft vs current not considering the kind of flow is problematic, however because the variable name explicitly explains that fact it rather won't be missused. :+1:

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/atomicStateMutators.ts
szymonmaslowski: Confirmed with Fernando, that is ok :+1:

PR #609, opened 2023-10-04
fix(staking): [LW-8714] handle percentages from sdk delegation distribution not adding up to 100%

refi93:
# Checklist

- [x] JIRA - LW-8714
- [x] Screenshots added.

---

## Proposed solution

Apparently sdk percentages not always add up to 100% which turned out to be caused by one of the wallet stake addresses not delegating anywhere. To prevent any such inconsistencies from causing a catastrophic failure (uncaught exception), this PR relaxes the `normalizePercentages` helper to fall back to naive rounding if the percentages happen to not add up to 100%

See screenshots to understand how this change manifests in UI

## Testing

Test impacted portfolio, check that management drawer/staking overview is properly shown

## Screenshots

Screenshot of the bug
CleanShot 2023-10-04 at 17 02 30@2x

After implementing this patch this is how the overview/management drawer look like:

Screenshot 2023-10-05 at 09 06 30
Screenshot 2023-10-05 at 09 06 35

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2363/6411235092/index.html) for [f8367d30](https://github.com/input-output-hk/lace/pull/609/commits/f8367d30bbdbbd336f52a0f63185c83ea6ea8fc6)

### DIFF
--- a/packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
+++ b/packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
@@ -7,10 +7,13 @@ import { PERCENTAGE_SCALE_MAX } from '../constants';
 export const normalizePercentages = <K extends string, T extends { [key in K]: number }>(items: T[], key: K) => {
   const currentSum = items.reduce((acc, item) => acc + item[key], 0);
   const epsilon = 0.1;
-  if (Math.abs(currentSum - PERCENTAGE_SCALE_MAX) > epsilon)
-    throw new Error(`Percentages must sum to ${PERCENTAGE_SCALE_MAX}`);
 
   const itemsWithRoundedNumbers: T[] = items.map((item) => ({ ...item, [key]: Math.round(item[key]) }));
+  if (Math.abs(currentSum - PERCENTAGE_SCALE_MAX) > epsilon) {
+    console.error(`Percentages must add up to ${PERCENTAGE_SCALE_MAX}, instead they add up to ${currentSum}`);
+    // fall back to naive rounding
+    return itemsWithRoundedNumbers;
+  }
 
   // Calculate the adjustment needed to make the sum exactly equal to 100 (PERCENTAGE_SCALE_MAX)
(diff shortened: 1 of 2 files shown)


packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
```diff
@@ -7,10 +7,13 @@ import { PERCENTAGE_SCALE_MAX } from '../constants';
...
   const currentSum = items.reduce((acc, item) => acc + item[key], 0);
   const epsilon = 0.1;
-  if (Math.abs(currentSum - PERCENTAGE_SCALE_MAX) > epsilon)
-    throw new Error(`Percentages must sum to ${PERCENTAGE_SCALE_MAX}`);
 
   const itemsWithRoundedNumbers: T[] = items.map((item) => ({ ...item, [key]: Math.round(item[key]) }));
+  if (Math.abs(currentSum - PERCENTAGE_SCALE_MAX) > epsilon) {
+    console.error(`Percentages must add up to ${PERCENTAGE_SCALE_MAX}, instead they add up to ${currentSum}`);
+    // fall back to naive rounding
+    return itemsWithRoundedNumbers;
```
szymonmaslowski, line 15: Alternatively, we could add the missing number temporarily for the calculations and remove it at the end, but for sure your solution is much simpler. Let's discuss it 😉

szymonmaslowski (approved):

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
refi93: I'd rather keep the simplest solution for now, I'm not really sure what's the desired rounding when dealing with percentages not adding up to 100% apart from ensuring the app doesn't crash, moreover, if the percentages happen to exceed 100%, the number added would be negative and I'm not sure how the custom algorithm would behave then, so I'd rather just use the simplest solution that works for us and tweak it later on, if ever needed

PR #611, opened 2023-10-05
feat(staking): [LW-8683] make normalization work with decimals, handle onchain percentages

xdzurman:
# Checklist

- [x] JIRA - \<link>
- [x] Screenshots added.

---

## Proposed solution

Make normalization work with decimals (to apply to onchain percentages). Still throw if sum is not 100, but check whether the sum is 100 before calling the function. To ensure a lossless sum of float percentages, a new function is implemented (sumPercentagesLossless). 

Now, instead of always calling normalization for onchain percentages, check if sum is 100 beforehand. If it isn't, still round the onchain percentages at 2 decimal places. No special gray pie slice is added, this is handled implicitly by DelegationCard.

## Testing

Check the app no longer crashes with the non-100% staked wallet. Check that onchain percentages are shown with 2 decimal places at most.

## Screenshots

image
image

image

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2439/6468135920/index.html) for [ffdd4eac](https://github.com/input-output-hk/lace/pull/611/commits/ffdd4eac7b67a0a96eb0cc24d7307a85b2d64832)
| Total | 27     | 5      | 0       | 0     | 32    | ❌     |


packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sumPercentagesLossless.ts
```diff
@@ -0,0 +1,14 @@
...
+  items,
+  key,
+}: {
+  items: T[];
+  key: K;
+}) => {
+  const sum = items.reduce((acc, item) => acc + item[key], 0);
+  // cut-off at 10 decimal places
+  // eslint-disable-next-line no-magic-numbers
+  return Number(sum.toFixed(10));
```
refi93, line 13: not sure how/if that can happen but `0.999999...` may get evaluated by this function to `0.9999999999` as `toFixed()` does no rounding, which I gues wouldn't be intended. Wouldn't it be safer to do a "precision round" like proposed in this SO answer? https://stackoverflow.com/a/49729715

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sumPercentagesLossless.ts
```diff
@@ -0,0 +1,14 @@
+// Sums numbers in an array of objects, avoids JS-precision issues
+// Intended for summing mainly onchain (float) percentages, without which there would be precision errors
+export const sumPercentagesLossless = <K extends string, T extends { [key in K]: number }>({
```
refi93, line 3: "lossless" is IMHO a bit too strong of a claim, I'd consider calling it rather "sanitized" in the sense that this function shouldn't output any "wild" values

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -26,6 +27,27 @@ const defaultState: DelegationPortfolioState = {
   viewedStakePool: undefined,
 };
 
+// If percentages add up to 100, normalize them. Otherwise, round them to N decimal places
+const adjustOnchainPercentages = <K extends string, T extends { [key in K]: number }>({
```
refi93, line 31: in line with my other [suggestion](https://github.com/input-output-hk/lace/pull/611/files#r1347649019)  I'd use the term "sanitize" 

```suggestion
const sanitizeOnchainPercentages = <K extends string, T extends { [key in K]: number }>({
```

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -26,6 +27,27 @@ const defaultState: DelegationPortfolioState = {
...
+  items,
+  key,
+  decimals = 0,
+}: {
+  items: T[];
+  key: K;
+  decimals?: number;
+}) =>
+  sumPercentagesLossless({ items, key }) === PERCENTAGE_SCALE_MAX
```
refi93, line 40: I'd leave here a comment clarifying that percentages may not add up to 100% especially if some of the stake keys bound to account is deregistered and has funds at the same time - this is IMHO something that a casual reader of this code has very low chance of figuring out on their own

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
```diff
@@ -1,19 +1,36 @@
...
-//                            =>  [{val: 34},   {val: 33},   {val: 33}   ]
-export const normalizePercentages = <K extends string, T extends { [key in K]: number }>(items: T[], key: K) => {
-  const currentSum = items.reduce((acc, item) => acc + item[key], 0);
+// Rounds values of properties of an input array of objects, while keeping the sum of the numbers equal to 100 (PERCENTAGE_SCALE_MAX)
+// Requires input array to sum to 100 +- epsilon
+// Works at arbitrary decimal precision
+// Examples: normalizePercentages({items: [{val: 33.33},{val: 33.33},{val: 33.33}], key: 'val', decimals: 0})
+//                            =>          [{val: 34},   {val: 33},   {val: 33}   ]
+//           normalizePercentages({items: [{val: 33.333},{val: 33.333},{val: 33.333}], key: 'val', decimals: 2})
+//                            =>          [{val: 33.34}, {val: 33.33}, {val: 33.33} ]
```
refi93, line 10: I'd make this a proper docstring for better development experience, so that it's shown by editor intellisense when hovering over the function invocation in the code
```suggestion
/**
 * Rounds values of properties of an input array of objects, while keeping the sum of the numbers equal to 100 (PERCENTAGE_SCALE_MAX)
 * Requires input array to sum to 100 +- epsilon
 * Works at arbitrary decimal precision
 * Examples:
 * ```
 *           normalizePercentages({items: [{val: 33.33},{val: 33.33},{val: 33.33}], key: 'val', decimals: 0})
 *                             =>          [{val: 34},   {val: 33},   {val: 33}   ]
 * ```
 * ```
 *           normalizePercentages({items: [{val: 33.333},{val: 33.333},{val: 33.333}], key: 'val', decimals: 2})
 *                            =>          [{val: 33.34}, {val: 33.33}, {val: 33.33} ]
 * ```
 */
```

refi93 (changes requested):
overall LGTM, left a few renaming suggestions

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sumPercentagesLossless.ts
xdzurman: toFixed does round the numbers at the specified decimal point. In the funciton, I changed this to bignumber implementation since their `toFixed` is less buggy and I added a JS-math edge case unit test

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sumPercentagesLossless.ts
```diff
@@ -0,0 +1,19 @@
...
+  decimals = 10,
+}: {
+  items: T[];
+  key: K;
+  decimals?: number;
+}) => {
+  // eslint-disable-next-line unicorn/no-array-reduce
+  const sum = items.reduce((acc, item) => acc.plus(new BigNumber(item[key])), new BigNumber(0));
+  // cut-off at N or 10 decimal places
+  return Number(sum.toFixed(decimals));
```
szymonmaslowski, line 18: I think cutting the tail of the sum is redundant. We don't use `decimals` argument anywhere. I would resign from this and just return the real sum without any rounding. If rounding would be needed somewhere then it could be handled in there explicitly

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
```diff
@@ -1,19 +1,43 @@
...
+  decimals = 0,
+}: {
+  items: T[];
+  key: K;
+  decimals?: number;
+}) => {
+  const currentSum = sumPercentagesLossless({ items, key });
   const epsilon = 0.1;
   if (Math.abs(currentSum - PERCENTAGE_SCALE_MAX) > epsilon)
     throw new Error(`Percentages must sum to ${PERCENTAGE_SCALE_MAX}`);
```
szymonmaslowski, line 30: If we stick to throwing then we need to catch it its usage places. Otherwise, it might happen we still haven't catch some edge case and the UI will be broken (e.g. staking won't show)

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/__tests__/normalizePercentages.test.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable no-magic-numbers */
+import { normalizePercentages } from '../normalizePercentages';
+
+const thirds = [{ value: 100 / 3 }, { value: 100 / 3 }, { value: 100 / 3 }];
+const sevenths = Array.from({ length: 7 }, () => ({ value: 100 / 7 }));
+const key = 'value';
+
+describe('Testing normalizePercentages function', () => {
+  test('Should correctly normalize numbers to 0 decimal places', () => {
```
szymonmaslowski, line 9: A nitpick, but the word `should` is redundant 😉 
```suggestion
  test('correctly normalizes numbers to 0 decimal places', () => {
```

szymonmaslowski (changes requested):
Good job!

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sumPercentagesLossless.ts
xdzurman: image
This is what it's needed for. I added "decimals" for modifiability, before I had just a magic number of 10 inline. I can remove that parameter, but cutting decimals somewhere is necessary.

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/normalizePercentages.ts
xdzurman: When operating on onchain percentages (sanitizeOnchainPercentages fn), before calling `normalizePercentages` there's an explicit check of `sumPercentagesSanitized` (previously sumPercentagesLossless), so this will never throw. 

The other calling place is
```
    const percentageValue = PERCENTAGE_SCALE_MAX / selections.length; // may be float
      state.draftPortfolio = selections.map((pool) => ({
        ...pool,
        sliderIntegerPercentage: percentageValue,
      }));
      state.draftPortfolio = normalizePercentages({ items: state.draftPortfolio, key: 'sliderIntegerPercentage' });
```
which deterministically assigns 100/N value N items, I can wrap that but I see it as an overkill. What do you think?

packages/staking/src/features/delegation-card/DelegationCard.tsx
```diff
@@ -3,11 +3,10 @@ import cn from 'classnames';
 import { Fragment, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { TranslationKey } from '../i18n';
-// import { PERCENTAGE_SCALE_MAX } from '../store';
+import { PERCENTAGE_SCALE_MAX } from '../store';
```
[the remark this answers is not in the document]
xdzurman: I imported it from ../store/constants, does that fix the issue? I'm afraid we would forget such TODO comments before merging.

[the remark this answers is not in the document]
szymonmaslowski: It is fine if it works 😄 Could you check if story for this component works fine? (`yarn dev`)

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sumPercentagesLossless.ts
szymonmaslowski: It cannot happen because you operate on a `BigNumber` object which solves the problem

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -12,6 +12,7 @@ import {
   processPopupViewCases,
 } from './stateMachine';
 import { normalizePercentages } from './stateMachine/normalizePercentages';
+import { sumPercentagesSanitized } from './stateMachine/sumPercentagesSanitized';
```
szymonmaslowski, line 15:
```suggestion
import { sumPercentagesSanitized } from './stateMachine';
```

szymonmaslowski (approved):

packages/staking/src/features/delegation-card/DelegationCard.tsx
```diff
@@ -3,11 +3,12 @@ import cn from 'classnames';
 import { Fragment, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { TranslationKey } from '../i18n';
-// import { PERCENTAGE_SCALE_MAX } from '../store';
+import { sumPercentagesSanitized } from '../store/delegationPortfolioStore/stateMachine/sumPercentagesSanitized';
 import * as styles from './DelegationCard.css';
-
-// TODO
+// import { PERCENTAGE_SCALE_MAX } from '../store';
+// re-declared due to ladle crash if importing from ../store
```
refi93, line 9: I'd consider making a ticket for this issue, and referencing it in the comment, so that we get to it hopefully sooner than later as keeping in mind such arbitrary constraint IMHO degrades development experience significantly

refi93 (approved):

PR #614, opened 2023-10-05
feat(staking): [LW-8315] remove withdrawals from activity, simplify tx transformers

refi93:
# Checklist

- [x] JIRA - LW-8315
- [x] Screenshots added.

---

## Proposed solution

Turns out "rewards" shown in activity so far were actually withdrawals which doesn't align with the intent behind this activity from a product perspective (should be actual pool rewards distributed at the end of the epoch). This PR removes the 'rewards' activity from the activity list shown to the user. Besides this removal, this PR simplifies the tx transformer logic, merging the duplicate code that has been there for pending and historical transactions

Note: proper rewards to be reintroduced with LW-8751

## Testing

Compare with main and check that 'rewards' have been dropped from activity (and nothing else). Check that when making a transaction/delegation, the pending transaction is shown properly.

## Screenshots
before

Screenshot 2023-10-10 at 10 45 58

after:

Screenshot 2023-10-10 at 10 45 13

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2424/6466411834/index.html) for [3202bbc7](https://github.com/input-output-hk/lace/pull/614/commits/3202bbc71234c2aebd1c118717c2409b9b272345)
| Total | 27     | 5      | 0       | 0     | 32    | ❌     |


apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/tx-history-transformer.ts
```diff
@@ -64,50 +59,53 @@ export const txHistoryTransformer = ({
...
-      {
-        ...transformedTx,
-        type: type === 'self-rewards' ? 'self' : 'outgoing'
-      },
-      {
-        type: 'rewards',
-        direction: 'Incoming',
-        amount: Wallet.util.getFormattedAmount({ amount: rewardsAmount, cardanoCoin }),
-        fiatAmount: getFormattedFiatAmount({
-          amount: new BigNumber(tx?.body?.withdrawals[0]?.quantity?.toString() || '0'),
```
refi93, line 96: this logic looks incorrect/inconsistent with the `amount` field as it takes into account only the first withdrawal, which obviously may not always be correct in case of multidelegation, I believe this should take the result of `getRewardsAmount()` - I fixed it

apps/browser-extension-wallet/src/views/browser-view/features/activity/components/TransactionDetail.tsx
```diff
@@ -101,7 +101,7 @@ export const TransactionDetail = withAddressBookContext<TransactionDetailProps>(
   );
 
   const currentTransactionStatus = useMemo(
-    () => getCurrentTransactionStatus(walletActivities, transactionDetail.tx.id) || transactionInfo?.status,
```
refi93, line 104: not 100% sure if this is OK, but due to the original logic the "parent" tx status was taken and therefore for any "rewards" done within a "non-rewards" tx with "success" status this resulted in the "rewards" activity detail being totally broken as can be seen in the screenshots

xdzurman: perhaps the underlying issue here is rather the rendering logic:
```ts
      {status === TransactionStatus.SPENDABLE ? (
        <RewardDetails {...rewardProps} />
      ) : (
        <Transaction {...transactionProps} />
      )}
``` 
we could perhaps render it based on `TransactionType` that's already available in the component

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/tx-history-transformer.ts
```diff
@@ -64,50 +59,53 @@ export const txHistoryTransformer = ({
...
-        status: Wallet.TransactionStatus.SPENDABLE,
-        date: transformedTx.date,
-        assets: [],
-        timestamp: transformedTx.timestamp
-      }
-    ] as Array<Omit<AssetActivityItemProps, 'onClick'>>;
+   * whenever the wallet have withdrawn rewards, we will need to create a new record Rewards and
+   * add it to the transaction history list given this, we will keep the original
+   * send transaction type and we will add this new Rewards record
+   */
```
xdzurman, line 88:
```suggestion
   * If the wallet has withdrawn rewards, create "Rewards" transaction history activity
   * alongside any tx activity that triggered it.
   */
```

apps/browser-extension-wallet/src/views/browser-view/features/activity/components/TransactionDetail.tsx
refi93: actually, I just checked and the rewards are shown properly on `main`, I'm puzzled TBH, probably I just forgot to properly refresh/rebuild the extension

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/__tests__/pending-tx-transformer.test.ts
```diff
@@ -93,29 +100,32 @@ describe('Testing tx transformers utils', () => {
         cardanoCoin,
         time
       });
-      expect(result).toStrictEqual({
-        type: 'incoming',
```
refi93, line 97: note that I changed the type from `incoming` to `outgoing` which is directly related to https://github.com/input-output-hk/lace/pull/614/files#diff-a8eeb7075dc2f548e0db1e6c371086d15c669c35eaeea39de965c20d39084361R43

Note that pending txs are in currently by design only outgoing: https://github.com/input-output-hk/lace/pull/614/files#diff-d3b69cb2d21707ff79d1247eb51ee935f7cb8d42981d7b91a2e2b9a43c3d7423L127 so the original test fixture, as I understood the code, had incorrect type which was caused by addresses missing in the transaction inputs fixture

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/tx-history-transformer.ts
refi93: this comment is outdated - the logic is completely dropped as the task description changed to dropping rewards completely (and reintroducing them in a separate task)

apps/browser-extension-wallet/src/stores/slices/transaction-detail-slice.ts
```diff
@@ -169,31 +168,14 @@ const getTransactionDetail =
         transaction = {
           ...transaction,
           pools: pools.map((pool) => ({
-            name: pool.metadata.name || '-',
-            ticker: pool.metadata.ticker || '-',
+            name: pool.metadata?.name || '-',
```
refi93, line 171: unrelated to this PR but pools without metadata were causing tx details to fail. This is actually pretty tricky to notice because we don't have strict null checks enabled in TS - do you perhaps know if there is/was some strong reason not to enable it?

apps/browser-extension-wallet/src/utils/__tests__/inspectTxType.test.ts
```diff
@@ -206,27 +205,7 @@ describe('testing tx-inspection utils', () => {
         ] as Wallet.KeyManagement.GroupedAddress[]
       });
 
-      expect(result).toBe('rewards');
-    });
-
-    test('is incoming rewards', () => {
```
refi93, line 212: I removed this test because it involves a transaction withdrawing from the account owned by the wallet whereas inputs come from a foreign wallet which is a pretty weird combination that I can't imagine in real life and I'm not sure if it makes sense to classify it as "outgoing" or "incoming". A few testcases below is almost the same testcase but with both the input and withdrawal coming from a foreign wallet so I just dropped this test instead of changing it

apps/browser-extension-wallet/src/stores/slices/transaction-detail-slice.ts
xdzurman: I agree to enable it, if it wouldn't take too much time. I'm not sure however how many places would blow up and how long it'd take to fix, considering product priorities.

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/common-tx-transformer.ts
```diff
@@ -0,0 +1,154 @@
...
+  cardanoCoin,
+  time,
+  date,
+  direction,
+  status
+}: TxTransformerInput): TransformedTx[] => {
+  // note that TxInfliht at type level does not expose the inputs with address
+  // which would prevent `inspectTxType` from determining whether tx is incoming or outgoing
+  // however at runtime the property is present (ATM) thanks to which the call below works
+  // ticket LW-8767 open with SDK team to fix the type of Input in TxInFlight to contain address
```
xdzurman, line 104:
```suggestion
  // Note that TxInFlight at type level does not expose the inputs with address,
  // which would prevent `inspectTxType` from determining whether tx is incoming or outgoing.
  // However at runtime, the property is present (ATM) and the call below works.
  // SDK Ticket LW-8767 should fix the type of Input in TxInFlight to contain the address
```

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/common-tx-transformer.ts
```diff
@@ -0,0 +1,154 @@
...
+  time,
+  date,
+  direction,
+  status
+}: TxTransformerInput): TransformedTx[] => {
+  // note that TxInfliht at type level does not expose the inputs with address
+  // which would prevent `inspectTxType` from determining whether tx is incoming or outgoing
+  // however at runtime the property is present (ATM) thanks to which the call below works
+  // ticket open with SDK team to fix the type of Input in TxInFlight to contain address
+  const type = inspectTxType({ walletAddresses, tx: tx as unknown as Wallet.Cardano.HydratedTx });
```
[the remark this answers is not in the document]
xdzurman: I'd also add a warning comment about the run-time property presence to the code, this cast looks quite dangerous
```
Type 'TxInFlight | HydratedTx' is not assignable to type 'HydratedTx'.
  Type 'TxInFlight' is missing the following properties from type 'HydratedTx': index, blockHeader, txSize, inputSource, witness
```

xdzurman (approved):
left minor comments

szymonmaslowski (approved):
Looks good :+1:

przemyslaw-wlodek (approved):
👍🏻 Great work @refi93 :) Please remember to remove those comments as soon as you merge the new PRs/tickets.

apps/browser-extension-wallet/src/views/browser-view/features/activity/helpers/common-tx-transformer.ts
```diff
@@ -0,0 +1,155 @@
...
+        // and the other transaction show fee to avoid duplicity
+        fee: '0'
+      }
+    ];
+  } else if (tx.depositReclaim) {
+    return [
+      {
+        ...tx,
+        type: 'delegation',
+        // Reclaimed deposit already shown in the delegationDeregistration
```
danielmain, line 58: very small: add a space between delegation and Deregistration.

danielmain (approved):
Awesome job. Also tested successfully 

CleanShot 2023-10-17 at 11 17 26@2x

PR #635, opened 2023-10-11
fix(extension): handle not visible in tx summary

shawnbusuttil:
# Checklist

- [ ] JIRA - \<LW-8492>
- [ ] Screenshots added.

---

## Proposed solution

Constructing the handle resolution from the handle in build transaction props.

## Testing

## Screenshots

image

---
# Allure report

**smokeTests**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2540/6506976309/index.html) for [17fd0764](https://github.com/input-output-hk/lace/pull/635/commits/17fd076401b8d8c3c67af5fe66dd29860f060d31)


VanessaPC (approved):
Looks good!

packages/cardano/src/wallet/lib/build-transaction-props.ts
```diff
@@ -47,7 +48,12 @@ export const buildTransactionProps = (props: {
           coins: BigInt(output.value.coins),
           ...(output?.value?.assets && { assets: convertAssetsToBigInt(output.value.assets, props.assetsInfo) })
         },
-        handle: output.handle?.startsWith('$') ? output.handle.slice(1) : output.handle
+        handleResolution: {
+          handle: output.handle,
+          cardanoAddress: output.address,
+          hasDatum: !!output.datum,
+          policyId: Wallet.Cardano.PolicyId('f0ff48bbb7bbe9d59a40f1ce90e9e9d0ff5002ec48f232b49ca0fb9a')
```
pczeglik-iohk, line 55: @shawnbusuttil Should we import exported PolicyId from `packages/cardano/src/wallet/lib/config.ts` instead of hard coding it?

pczeglik-iohk: or actually from `packages/cardano/src/wallet/index.ts` :)

shawnbusuttil: Nice catch.

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransactionSummary.tsx
```diff
@@ -138,8 +138,8 @@ export const SendTransactionSummary = withAddressBookContext(
 
     const rows = [...(outputs?.values() ?? [])].map((item) => ({
       list: formatRow({ output: item, assetInfo: assetsInfo, cardanoCoin, fiatCurrency, prices: priceResult }),
-      recipientAddress: item.address.toString(),
-      recipientName: addressToNameMap?.get(item.address.toString()) || item.handle
+      recipientAddress: item.address,
+      recipientName: addressToNameMap?.get(item.handleResolution.handle || item.address) || item.handleResolution.handle
```
pczeglik-iohk, line 142: I'm trying to understand this line and I may have impression that this should be something like this:

if there is a `handle` use it, otherwise get name by address. If there is no name, default to empty string? 

```
recipientName: item.handleResultion.handle ?? addressToNameMap?.get(item.address) ?? ''
```

If I follow the code this `addressToNameMap?.get(item.handleResolution.handle || item.address)` will always returns `undefined` in case `handle` has value. We map address to name, `handle` is not an address right?

What do you think?

shawnbusuttil: If there is a saved address use it. To get the value of the saved address, use the key (which is a handle or a public key in that order). If there is no saved address, use a handle. If there is neither, display nothing.

PR #640, opened 2023-10-11
feat(staking): [LW-8628] read portfolio persistence data

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-8628)
- [ ] Screenshots added.

---

## Proposed solution

1. Reading the saved weights of the current portfolio
2. Transforming saved weights to percentages
3. Storing saved percentages as `savedIntegerPercentage` on the current portfolio data
   - setting the value to 0 if there is no persistence data
5. copying over the `savedIntegerPercentage` to `sliderIntegerPercentage`
   - falling back to normalized `onChainPercentage` in case the `savedIntegerPercentage` is 0 (no persistence data)
6. detecting the case of no persistence data in the drift warning banner
   - if on chain does not sum up to 100 - warning visible
   - if on chain sums up to 100 + no saved percentage data - no warning
## Testing

Please check cases of no persistence data stored yet
Please check the case if there are unallocated funds (incoming transaction) + no persistence data
to be checked: drift warning, drawer step with sliders

## Screenshots

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2649/6614724853/index.html) for [ba665376](https://github.com/input-output-hk/lace/pull/640/commits/ba665376906130de219d53fe3b9bb03575280e5f)
| Total | 29     | 3      | 0       | 0     | 32    | ❌     |


package.json
```diff
@@ -24,7 +24,7 @@
     "packages/ui"
   ],
   "scripts": {
-    "base-command-for-build-scripts": "wsrun --prefix -tram ${APP_NAME:+-p \"@lace/$APP_NAME\"} ${EXCLUDE_APP:+-x \"@lace/$APP_NAME\"} ${EXCLUDE_BROWSER:+-x \"@lace/browser-extension-wallet\"}",
+    "base-command-for-build-scripts": "wsrun --prefix -trame ${APP_NAME:+-p \"@lace/$APP_NAME\"} ${EXCLUDE_APP:+-x \"@lace/$APP_NAME\"} ${EXCLUDE_BROWSER:+-x \"@lace/browser-extension-wallet\"}",
```
szymonmaslowski, line 27: Additional `e` parameter makes the command failing if an error happens to any of the executed scripts
The `yarn browser build` will fail fast if e.g. `common` package fails to build

packages/staking/src/features/Drawer/preferences/PoolDetailsCard/PoolDetailsCardData.tsx
```diff
@@ -31,7 +31,7 @@ export const PoolDetailsCardData = ({
           </Tooltip>
         </Flex>
         <Text.Body.Large weight="$semibold">
-          {savedPercentage || '-'} {savedPercentage && <Text.Body.Small weight="$medium">%</Text.Body.Small>}
+          {savedPercentage || '-'} {!!savedPercentage && <Text.Body.Small weight="$medium">%</Text.Body.Small>}
```
szymonmaslowski, line 34: Casting to boolean to avoid displaying unnecessary `0`

packages/staking/src/features/overview/helpers/isPortfolioDrifted.ts
```diff
@@ -22,6 +22,13 @@ const getPortfolioTotalPercentageDrift = (portfolio: CurrentPortfolioStakePool[]
 
 // TODO: move this file to store. It gets imported also outside the overview feature so the store seems better place.
 export const isPortfolioDrifted = (currentPortfolio: CurrentPortfolioStakePool[]) => {
+  const savedPercentageMissing = currentPortfolio.some(({ savedIntegerPercentage }) => !savedIntegerPercentage);
+  const onChainPercentageSum = sumPercentagesSanitized({
+    items: currentPortfolio,
+    key: 'onChainPercentage',
+  });
+  if (onChainPercentageSum !== PERCENTAGE_SCALE_MAX) return true;
```
szymonmaslowski, line 30: A case when there are more funds in the wallet and overview shows them as unallocated.

packages/staking/src/features/overview/helpers/isPortfolioDrifted.ts
```diff
@@ -22,6 +22,13 @@ const getPortfolioTotalPercentageDrift = (portfolio: CurrentPortfolioStakePool[]
...
+  if (savedPercentageMissing) return false;
```
szymonmaslowski, line 31: In case there is `onChainPercentages` sum up to 100% BUT there is no `savedPercentage` (portfolio built before persistence) then there is no point in comparing the on-chain to saved. We cannot detect if there was a drift - ignoring this case

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/processExpandedViewCases.ts
```diff
@@ -57,10 +58,14 @@ import {
...
 export const currentPortfolioToDraft = (pools: CurrentPortfolioStakePool[]): DraftPortfolioStakePool[] =>
-  pools.map((cp) => ({
-    ...cp,
-    sliderIntegerPercentage: cp.savedIntegerPercentage,
-  }));
+  sanitizePercentages({
+    decimals: 0,
+    items: pools.map((cp) => ({
+      ...cp,
+      sliderIntegerPercentage: cp.savedIntegerPercentage || cp.onChainPercentage,
```
szymonmaslowski, line 65: Falling back slider values to on-chain values if there is no saved values

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/sanitizePercentages.ts
```diff
@@ -0,0 +1,24 @@
+// If percentages add up to 100, normalize them. Otherwise, round them to N decimal places.
```
szymonmaslowski, line 1: Moved from https://github.com/input-output-hk/lace/pull/640/files#diff-8c2da5fb662c55f1e206a0dc2c425e019a4aaa051ad51af64e5dace701b9615cL30

packages/staking/src/features/store/delegationPortfolioStore/stateMachine/processExpandedViewCases.ts
```diff
@@ -57,10 +58,14 @@ import {
 } from './types';
 
 export const currentPortfolioToDraft = (pools: CurrentPortfolioStakePool[]): DraftPortfolioStakePool[] =>
-  pools.map((cp) => ({
-    ...cp,
-    sliderIntegerPercentage: cp.savedIntegerPercentage,
-  }));
+  sanitizePercentages({
```
szymonmaslowski, line 61: Sanitizing in case `sliderIntegerPercentage` received `onChainPercentage` values

packages/staking/src/features/overview/helpers/isPortfolioDrifted.ts
refi93: I'd find it more resilient to modify the `getPortfolioTotalPercentageDrift()` function to deal with partially delegated portfolios which then removes the need for this check which may result in false positives, if just a small fraction of funds is undelegated (imagine the extreme case sending 1 ADA to an undelegated address to a wallet with 1M ADA, which would trigger this condition which is IMHO not desired)

I'd propose something like this:

```
const getPortfolioTotalPercentageDrift = (portfolio: CurrentPortfolioStakePool[]): number => {
  const onChainPercentageSum = sumPercentagesSanitized({
    items: portfolio,
    key: 'onChainPercentage',
  });
  // note that "on chain" percentages of the portfolio may not add up if part of the funds
  // reside on an undelegated stake key and the code below deals with that edge-case
  const totalDelegatedValue = Wallet.BigIntMath.sum(portfolio.map(({ value }) => value));
  const totalValue = new BigNumber(totalDelegatedValue.toString()).times(PERCENTAGE_SCALE_MAX / onChainPercentageSum);
  return sum(
    portfolio.map(({ value, savedIntegerPercentage }) => {
      const currentPercentage = new BigNumber(value.toString()).div(totalValue).times(PERCENTAGE_SCALE_MAX).toNumber();
      return Math.abs(savedIntegerPercentage - currentPercentage);
    })
  );
};
```

With the proposed version of `getPortfolioTotalPercentageDrift()` the `if (onChainPercentageSum !== PERCENTAGE_SCALE_MAX) return true;` check will become redundant

packages/staking/src/features/overview/helpers/isPortfolioDrifted.ts
```diff
@@ -22,6 +22,13 @@ const getPortfolioTotalPercentageDrift = (portfolio: CurrentPortfolioStakePool[]
 
 // TODO: move this file to store. It gets imported also outside the overview feature so the store seems better place.
```
refi93, line 23: as part of this PR I'd consider already moving this helper to store as suggested, should be a small change

packages/staking/src/features/overview/helpers/isPortfolioDrifted.ts
refi93: I'd leave a comment in the code clarifying this. In the future, when portfolios would be persisted by default for a long time already, it may no longer be obvious why this check is here

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -109,53 +87,69 @@ export const useDelegationPortfolioStore = create(
...
-          };
+            return {
+              displayData: {
+                ...mapStakePoolToDisplayData({ cardanoCoinSymbol: get().cardanoCoinSymbol, stakePool }),
+                lastReward: confirmedPoolRewards[confirmedPoolRewards.length - 1] || BigInt(0),
+                totalRewards: Wallet.BigIntMath.sum(confirmedPoolRewards),
+              },
+              id: stakePool.hexId,
+              onChainPercentage: percentage * PERCENTAGE_SCALE_MAX,
+              savedIntegerPercentage: mapOfPoolIdBySavedPercentages[stakePool.hexId] || 0,
```
refi93, line 132: I think it would be better for readability if we explicitly captured with `null` that the given saved percentage is just missing:

```suggestion
              savedIntegerPercentage: mapOfPoolIdBySavedPercentages[stakePool.hexId] ?? null,
```

e.g. `isPortfolioDrifted()` helper and the check for missing percentage would become more readable as checking for `zero` (which one may argue is a valid value) is IMHO unnecessarily "hacky"

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -109,53 +87,69 @@ export const useDelegationPortfolioStore = create(
...
-            return acc;
-          }, {} as Record<Wallet.Cardano.PoolIdHex, number>);
+          mapOfPoolIdBySavedPercentages = portfolioWithPercentages.reduce<Record<Wallet.Cardano.PoolIdHex, number>>(
+            (acc, { percentage, id }) => {
+              acc[id] = percentage;
+              return acc;
+            },
+            {}
+          );
+        }
```
refi93, line 116: I'd consider simplifying/extracting this logic into a helper `mapPoolsToSavedPercentages`:

```typescript
const getPoolsSavedPercentages = (pools: Wallet.Cardano.Cip17Pool[]) => {
  const weightsSum = Number(Wallet.BigIntMath.sum(pools.map(({ weight }) => BigInt(weight))));
  // eslint-disable-next-line unicorn/no-array-reduce
  return pools.reduce<Record<Wallet.Cardano.PoolIdHex, number>>((acc, { id, weight }) => {
    acc[id] = (weight / weightsSum) * PERCENTAGE_SCALE_MAX;
    return acc;
  }, {});
};
```

refi93: or I guess it would be even more convenient if the helper already sanitized the percentages:

```typescript
const mapPoolsToSavedIntegerPercentages = (pools: Wallet.Cardano.Cip17Pool[]) => {
  const weightsSum = Number(Wallet.BigIntMath.sum(pools.map(({ weight }) => BigInt(weight))));
  const rawPoolPercentages = pools.map(({ id, weight }) => ({
    id,
    percentage: (weight / weightsSum) * PERCENTAGE_SCALE_MAX,
  }));
  const sanitizedPoolPercentages = sanitizePercentages({
    decimals: 2,
    items: rawPoolPercentages,
    key: 'percentage',
  });

  // eslint-disable-next-line unicorn/no-array-reduce
  return sanitizedPoolPercentages.reduce<Record<Wallet.Cardano.PoolIdHex, number>>((acc, { id, percentage }) => {
    acc[id] = percentage;
    return acc;
  }, {});
};
```

packages/staking/src/features/overview/helpers/isPortfolioDrifted.ts
szymonmaslowski: IMO it should not be placed in the store if it does not use it. I can see usage in the overview and the drawer, so it doesn't belong to any of the `features`' directories, so it should get its own directory 😉 I will think about it.

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
szymonmaslowski: I considered `null` at first, but I realized it requires more changes in places, where we read the value. In those places, we would need to fall back it to 0 anyway. Not in all the places, though. I am tentative about this, but I support your argument about readability.

packages/staking/src/features/store/delegationPortfolioStore/useDelegationPortfolioStore.ts
```diff
@@ -98,22 +99,7 @@ export const useDelegationPortfolioStore = create(
...
-          // eslint-disable-next-line unicorn/no-array-reduce
-          mapOfPoolIdBySavedPercentages = portfolioWithPercentages.reduce<Record<Wallet.Cardano.PoolIdHex, number>>(
-            (acc, { percentage, id }) => {
-              acc[id] = percentage;
-              return acc;
-            },
-            {}
-          );
-        }
+        const mapOfPoolIdBySavedPercentages = makeMapOfSavedPercentages(delegationPortfolio?.pools);
```
refi93, line 102: naming nitpick: `mapOfPoolIdBySavedPercentages` makes it sound as if the percentages were the keys which obviously isn't the case, I'd consider naming it `mapOfPoolIdToSavedPercentages` instead

packages/staking/src/features/portfolio-drift/isPortfolioDrifted.ts
```diff
@@ -20,8 +22,19 @@ const getPortfolioTotalPercentageDrift = (portfolio: CurrentPortfolioStakePool[]
   );
 };
 
+const isSavedPercentagePresent = (
+  currentPortfolio: CurrentPortfolioStakePool[]
+): currentPortfolio is (CurrentPortfolioStakePool & { savedIntegerPercentage: number })[] =>
+  currentPortfolio.some(({ savedIntegerPercentage }) => !savedIntegerPercentage);
```
refi93, line 28: I guess this should actually be the inverse: 

```suggestion
  currentPortfolio.every(({ savedIntegerPercentage }) => savedIntegerPercentage);
```

also, I'd consider calling the function `hisPortfolioWithSavedPercentages` which IMHO aligns better with its purpose/logic

refi93 (changes requested):
looks good 👍 just last few nitpicks

packages/staking/src/features/portfolio-drift/isPortfolioDrifted.ts
szymonmaslowski: Nice catch! Thanks for spotting this! I might have missed this case while testing 😅

refi93 (approved):

PR #650, opened 2023-10-17
feat(staking): [LW-8110] add analytics

przemyslaw-wlodek:
# Checklist

- [x] JIRA - LW-8110
- [ ] Screenshots added.

---

## Proposed solution

Track events referenced in the task in the new multistaking interface. As there were no analytics in the `staking` packagge so far, as part of this PR we also expose the analytics from the `browser-extension` package into `staking` by extending the `OutsideHandlesProvider`.

## Testing

Built an extension with enabled analytics and check that events added in this PR are properly logged into posthog - e.g. typing into the "browse pools" search bar should trigger an event. For other events check the google doc linked in the task, it contains also high level descriptions of how/when the events should be logged.

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2640/6593288334/index.html) for [d63dd04b](https://github.com/input-output-hk/lace/pull/650/commits/d63dd04bff654790cafd31c3c13c16f294d9b743)


packages/staking/src/features/outside-handles-provider/types.ts
```diff
@@ -48,12 +48,60 @@ export interface IBlockchainProvider {
   rewardsProvider: Wallet.RewardsProvider;
 }
 
-export type StakingRewards = {
-  totalRewards: BigInt | number;
-  lastReward: BigInt | number;
+export enum PostHogAction {
+  StakingClick = 'staking | staking | click',
```
refi93, line 52: @przemyslaw-wlodek I thought about somehow getting rid of the need to declare the allowed posthog events twice - once in the browser-extension repo and once here and I didn't really find an alternative that I would consider worth implementing right now.

One possible approach I thought of was exposing a generic version of the `sendEventToPostHog()` method which would accept a string, so the staking package could pass there any event it declares, but it would IMHO make the tracking just more complex to reason about and in the end, it's IMHO actually beneficial to have all the events tracked by the app gathered in one single place for the sake of visibility, identifying inconsistencies, duplicates etc. so I chose not to go forward with that. On the other hand, if we wanted, the current implementation is covered by types and should be as refactorable now as it would be in the future, if we ever find any issues

Another approach I considered was moving tracking to a separate package, but that would IMHO just make the posthog client management more complex, especially if we wanted to maintain a single instance of it which seems sensible. I guess we would need then to also extract some kind of global "service locator" as a package that would inject the shared dependencies and in the end, we'd just save the need of declaring this enum which IMHO doesn't seem to justify the effort right now - again, should be as doable in the future as now, the "tech debt" as we extend this should boil down just to more items in this list so I don't see any particular gain in doing such refactor right now, and I'd rather tackle it methodically as we rething the `UseOutsideHandles` approach as such

marcin-mazurek (commented):
Great to see PostHog in action 🚀

packages/staking/src/features/outside-handles-provider/types.ts
refi93: > you should be able to apply generic types to PostHogClient and AnalyticsTracker, and declare your own StakingPostHogActions enum without duplicating code. I would definitely wouldn't swap PostHogAction for a string. Type checking is there for a reason.

That's actually the approach I tried initially, just forgot to mention it. What stopped me from pursuing it further was that it seemed quite fragile/dirty to me to re-cast with `as` the instance of analytics instantated with the `PostHogActions` type earlier [passed to the OutsideHandlesProvider ](https://github.com/input-output-hk/lace/pull/650/files#diff-22082fe1bbf5006106fe8852f6832c9122280bf0eb60837ba5072c33094e058dL82) and I couldn't think of a reasonable way around that. Nevertheless I agree that it would be cleaner than exposing a method which accepts raw strings in the sense that at least whoever uses it would be motivated to pass some proper type to the generic.

> Maybe, even simpler approach would be to move the event list to one of shared packages such as @lace/common or @lace/core? I never touched either of them and I'm not sure which one would be more suitable, but it feels like it's basically their purpose.

I like that, will do 👍

refi93: agreed offline to revert the export of Matomo stuff to common, will do

marcin-mazurek (approved):

emiride (approved):
I tested this and it works fine.

greatertomi (approved):

PR #659, opened 2023-10-23
feat: [LW-7837] add warning banner for retiring, retired and saturated stake pools

greatertomi:
# Checklist

- [x] JIRA - LW-7837
- [x] Screenshots added.

---

## Proposed solution

Add a warning banner when any pool in the user portfolio is saturated, retiring or retired.

## Screenshots

Screenshot 2023-10-23 at 20 29 43

---
# Allure report

**smokeTests**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2715/6653412193/index.html) for [42341835](https://github.com/input-output-hk/lace/pull/659/commits/42341835db317cdee502706c0e0deab865bd0aaa)


packages/staking/src/features/overview/StakingNotificationBanner/StakingNotificationBanner.tsx
```diff
@@ -50,6 +51,17 @@ export const StakingNotificationBanner = ({
...
       );
+    case 'poolSaturatedOrRetired':
+      return (
+        <Banner
+          popupView={popupView}
+          withIcon
+          customIcon={<ExclamationIcon />}
+          message={t('overview.banners.saturatedOrRetiredPool.title')}
+          description={t('overview.banners.saturatedOrRetiredPool.message')}
+          onBannerClick={onPortfolioDriftedNotificationClick}
```
refi93, line 62: the `onPortfolioDriftedNotificationClick` name no longer makes sense as it's being used also for the case of saturated/retired pools which is not related to portfolio drift. To tackle this, I'd consider generalizing the name of the prop to `clickableBannerAction`

refi93: alternatively, we can just encapsulate this logic into `StakingNotificationBanner` and bind it to `isPopupView` flag (i.e. open management drawer if expanded view, otherwise open expanded view) - I think that would indeed be cleaner

packages/staking/src/features/overview/StakingNotificationBanner/getCurrentStakingNotification.ts
```diff
@@ -22,5 +22,9 @@ export const getCurrentStakingNotification = ({
     return 'portfolioDrifted';
   }
 
+  if (hasSaturatedOrRetiredPools(currentPortfolio)) {
```
refi93, line 25: I don't see in the task the priority of the "saturated or retired" notification vs the "drift" one and I can see arguments in favor of either ordering. I'd suggest clarifying the precence of notification in such a case with Fernando, if you already didn't

packages/staking/src/features/overview/helpers/hasSaturatedOrRetiredPools.ts
```diff
@@ -0,0 +1,11 @@
+import { CurrentPortfolioStakePool } from '../../store';
+
+export const SATURATION_UPPER_BOUND = 100;
```
refi93, line 3: I'd consider extracting this constant which is already used in the staking package to a `constants.ts` file within the `helpers` folder

packages/staking/src/assets/icons/triangle-exclamation.component.svg
refi93, line 1: to prevent further duplication/inconsistencies of assets I'd consider moving this (together with other icons in this folder, for the sake of consistency) to `packages/ui/assets/icons` folder from where we could import them as we already do e.g. for `chevron-up` icon: https://github.com/input-output-hk/lace/blob/0ce732940b0aea274e9036c491863c21e409b607/packages/staking/src/features/drawer/preferences/PoolDetailsCard/PoolDetailsCard.tsx#L3C1-L3C1

refi93 (changes requested):
see comments

packages/staking/src/features/overview/StakingNotificationBanner/getCurrentStakingNotification.ts
greatertomi: I asked Fernando and he said we should show both banners. 8a1ba0d does that.

packages/staking/src/features/overview/helpers/hasSaturatedOrRetiredPools.ts
greatertomi: Fixed on 82593bd

mrcnk (approved):

packages/staking/src/features/overview/Overview.tsx
```diff
@@ -111,6 +118,17 @@ export const Overview = () => {
           />
         </Box>
       )}
+      {isPoolRetiredOrSaturated && (
+        <Box mb="$40">
+          <Banner
```
refi93, line 123: I'd consider keeping this Banner within the `StakingNotificationBanner` component to keep the notification logic in a single place, otherwise it may become hard to follow what notifications are shown when which may result in undesired combinations of the notifications being shown

I'd suggest refactoring `getCurrentStakingNotification()` to return an array of notifications (most of the time it would be empty or single item, with the exception of saturated/retired pool and drift at the same time) and then this Banner component should be movable into `StakingNotificationBanner` component (which I'd also refactor to take an array of notifications and rename it to `StakingNotificationBanners`

What do you think?

refi93 (changes requested):
see comment

packages/staking/src/features/overview/Overview.tsx
greatertomi: I agree. Created an array of notifications here: 97d92bb

packages/staking/src/features/overview/StakingNotificationBanner/StakingNotificationBanner.tsx
refi93: agreed offline to go with the first suggestion, but naming the flag `onClickableBannerClick` which is simpler while it should make clear enough to the reader that not all banners are clickable

packages/staking/src/features/overview/StakingNotificationBanner/StakingNotificationBanners.tsx
```diff
@@ -0,0 +1,70 @@
+import { Banner } from '@lace/common';
+import ExclamationIcon from '@lace/ui/dist/assets/icons/warning-icon-triangle.component.svg';
+import { useTranslation } from 'react-i18next';
+import BellIcon from '../../../assets/icons/bell-icon.component.svg';
+import InfoIcon from '../../../assets/icons/info-icon.svg';
+import * as styles from './StakingNotificationBanner.css';
+import { StakingNotificationType } from './types';
+
+type StakingNotificationBannerProps = {
```
refi93, line 9:
```suggestion
type StakingNotificationBannersProps = {
```

refi93 (approved):

oldGreg5 (approved):

PR #661, opened 2023-10-24
fix: lace crashing on send due to missing handle resolution

shawnbusuttil:
# Checklist

- [ ] JIRA - \<LW-8873>
- [ ] Screenshots added.

---

## Proposed solution

Lace crashing during send transaction due to altered network response and no optional guards on the frontend. 

## Testing

Go through send flow with/without a handle.

## Screenshots

### DIFF
--- a/apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransactionSummary.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransactionSummary.tsx
@@ -102,7 +102,7 @@ interface SendTransactionSummaryProps {
 export const SendTransactionSummary = withAddressBookContext(
   ({ isPopupView = false }: SendTransactionSummaryProps): React.ReactElement => {
     const { t } = useTranslation();
-    const { builtTxData: { uiTx: { fee, outputs } = {} } = {} } = useBuiltTxState();
+    const { builtTxData: { uiTx: { fee, outputs, handleResolutions } = {} } = {} } = useBuiltTxState();
     const [metadata] = useMetadata();
     const { inMemoryWallet } = useWalletStore();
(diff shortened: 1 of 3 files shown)


bslabiak (approved):

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransactionSummary.tsx
```diff
@@ -136,10 +136,11 @@ export const SendTransactionSummary = withAddressBookContext(
       [addressList]
     );
 
-    const rows = [...(outputs?.values() ?? [])].map((item) => ({
+    const rows = [...(outputs?.values() ?? [])].map((item, idx) => ({
       list: formatRow({ output: item, assetInfo: assetsInfo, cardanoCoin, fiatCurrency, prices: priceResult }),
       recipientAddress: item.address,
-      recipientName: addressToNameMap?.get(item.handleResolution.handle || item.address) || item.handleResolution.handle
+      recipientName:
+        addressToNameMap?.get(handleResolutions[idx]?.handle || item.address) || handleResolutions[idx]?.handle
```
VanessaPC, line 143: `addressToNameMap` works with `useMemo` a few lines above, this is unnecessary in this scenario and wasting memory usage, but it's also making this `get` slower. A small improvement would be to get rid of it.

VanessaPC (approved):
Nice work, it works well!

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/SendTransactionSummary.tsx
shawnbusuttil: Thanks for bringing it up. Investigating separately as removing it triggers the error this PR sought to originally fix (time is of the essence in this case).

PR #671, opened 2023-10-26
feat(staking): [LW-8837] add portfolio persistence one-time message

greatertomi:
# Checklist

- [x] JIRA - LW-8837
- [x] Screenshots added.

---

## Proposed solution

Add a one-time announcement modal that shows on the staking view when the user has an active multi-delegation portfolio.

## Testing

## Screenshots

Screenshot 2023-10-26 at 16 11 48

---
# Allure report

**smokeTests**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/linux/chrome/2793/6706378497/index.html) for [19dc1871](https://github.com/input-output-hk/lace/pull/671/commits/19dc187146df27175eef44ee9529840d1299e9d7)
| Total | 31     | 0      | 0       | 0     | 31    | ✅     |

### DIFF
--- a/packages/staking/src/features/staking/StakingView.tsx
+++ b/packages/staking/src/features/staking/StakingView.tsx
@@ -3,7 +3,7 @@ import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { BrowsePools } from '../BrowsePools';
 import { Drawer } from '../Drawer';
-import { ChangingPreferencesModal, MultidelegationBetaModal } from '../modals';
+import { ChangingPreferencesModal, MultidelegationBetaModal, PortfolioPersistenceModal } from '../modals';
 import { useOutsideHandles } from '../outside-handles-provider';
 import { Overview } from '../overview';
 import { DrawerManagementStep, DrawerStep, useDelegationPortfolioStore } from '../store';
@@ -13,7 +13,8 @@ const stepsWithBackBtn = new Set<DrawerStep>([DrawerManagementStep.Confirmation,
 
 export const StakingView = () => {
   const { t } = useTranslation();
-  const { portfolioMutators } = useDelegationPortfolioStore((store) => ({
+  const { portfolioMutators, currentPortfolio } = useDelegationPortfolioStore((store) => ({
+    currentPortfolio: store.currentPortfolio,
     portfolioMutators: store.mutators,
   }));
   const {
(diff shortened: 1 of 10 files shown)


packages/staking/src/features/staking/StakingView.tsx
```diff
@@ -48,7 +49,11 @@ export const StakingView = () => {
       </Navigation>
       <StakePoolDetails showCloseIcon showBackIcon={(step: DrawerStep): boolean => stepsWithBackBtn.has(step)} />
       <ChangingPreferencesModal />
-      <MultidelegationBetaModal visible={multidelegationFirstVisit} onConfirm={triggerMultidelegationFirstVisit} />
+      {currentPortfolio.length > 1 ? (
+        <PortfolioPersistenceModal visible={multidelegationFirstVisit} onConfirm={triggerMultidelegationFirstVisit} />
```
refi93, line 53: `multidelegationFirstVisit` will in practice already be `false` for users that are multidelegating because they obviously visited this section in the past. Therefore this one-time message wouldn't be visible to them.

If I understood that correctly, I think we need to introduce a new flag into local storage besides `multidelegationFirstVisit` (e.g. `multidelegationMigrationWarningSeen`) where we would capture if the user has already seen this banner or not

refi93 (changes requested):
see comment

packages/staking/src/features/staking/StakingView.tsx
refi93: clarified offline - the important case where the `multidelegationFirstVisit` flag alone would fail is that users that have seen the original "multidelegationFirstVisit" modal in earlier releases won't see the new modal at all because the `multidelegationFirstVisit` flag is already false for them. We can't just reset the `multidelegationFirstVisit` flag to `false` either because it would obviously result in the original "first visit" modal to appear for users without an active portfolio.

So the only viable solution I see is introducing a new separate flag for this modal and update the logic to handle it appropriately together with the pre-existing `multidelegationFirstVisit` ensuring the correct modal is shown (or not) to the user

greatertomi: Fixed on f794d58

przemyslaw-wlodek (approved):
There's a bit of duplication here but that's fine for now. Let's ship it :) Thanks for addressing @refi93 comments

emiride (approved):
Works as expected

PR #674, opened 2023-10-26
Fix(extension): [LW-8723] make deletion of wallet more robust

refi93:
# Checklist

- [ ] JIRA - LW-8723
- [ ] Screenshots added.

---

## Proposed solution

There are two problems with the current implementation of wallet deletion:

1. items from local/background storage were explicitly cleaned meaning there was a chance that important items could be left out
2. zustand stores with wallet info (staking history, ...) remained untouched after deleting the wallet, meaning that if the newly loaded wallet didn't overwrite them, data of the previous wallet were shown

The solution for problem 1 was to refactor the functions to clean local/background storage to expose the option to pass a whitelist of keys to keep and delete the rest, ensuring we keep in storage only what we really want

The solution for problem 2 is forcing a full reload of the current Lace tab and to prevent the setup page from flickering a new `deletingWallet` flag has been added to the wallet store to signal that a wallet is being deleted.

Note that the other open Lace tabs already reloaded due to listening to storage changes of key agent data triggered by other documents - see docs: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

## Testing

Try deleting a wallet and check that the Lace tab reloads and SetupPage appears. Check that Lace behaves as expected also when wiping the wallet if user clicks "forgot password" on locked wallet.

## Screenshots

### DIFF
--- a/apps/browser-extension-wallet/src/features/unlock-wallet/components/UnlockWalletContainer.tsx
+++ b/apps/browser-extension-wallet/src/features/unlock-wallet/components/UnlockWalletContainer.tsx
@@ -19,7 +19,7 @@ export interface UnlockWalletContainerProps {
 export const UnlockWalletContainer = ({ validateMnemonic }: UnlockWalletContainerProps): React.ReactElement => {
   const analytics = useAnalyticsContext();
   const { unlockWallet, lockWallet, deleteWallet } = useWalletManager();
-  const { environmentName, setKeyAgentData } = useWalletStore();
+  const { environmentName, setKeyAgentData, setDeletingWallet } = useWalletStore();
   const backgroundService = useBackgroundServiceAPIContext();
 
   const [isVerifyingPassword, setIsVerifyingPassword] = useState(false);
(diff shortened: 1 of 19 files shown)


apps/browser-extension-wallet/src/features/unlock-wallet/components/UnlockWalletContainer.tsx
```diff
@@ -75,8 +75,10 @@ export const UnlockWalletContainer = ({ validateMnemonic }: UnlockWalletContaine
   const onForgotPasswordClick = async (): Promise<void> => {
     await analytics.sendEventToPostHog(PostHogAction.UnlockWalletForgotPasswordProceedClick);
     saveValueInLocalStorage({ key: 'isForgotPasswordFlow', value: true });
+    setDeletingWallet(true);
     await deleteWallet(true);
     await backgroundService.handleOpenBrowser({ section: BrowserViewSections.FORGOT_PASSWORD });
+    setDeletingWallet(false);
```
refi93, line 81: in practice, when user clicks "forgot password" in the popup view lock screen, the popup view closes after opening the browser window with terms and conditions, but TBH I'm not sure why and it's not obvious from the code, so I just defensively set "deletingWallet" flag to `false` here

apps/browser-extension-wallet/src/utils/local-storage.ts
```diff
@@ -40,6 +40,16 @@ export const saveValueInLocalStorage = <T = ILocalStorage, K extends keyof T = k
 
 export const deleteFromLocalStorage = (key: keyof ILocalStorage): void => window.localStorage.removeItem(key);
 
+type ClearLocalStorageOptions = { except: (keyof ILocalStorage)[] };
+export const clearLocalStorage = (params?: ClearLocalStorageOptions): void => {
+  const except = params?.except || [];
+  for (const key in window.localStorage) {
```
przemyslaw-wlodek, line 46: This requires `hasOwnProperty` check. I will update it.

przemyslaw-wlodek (approved):
🚀

apps/browser-extension-wallet/src/utils/local-storage.ts
przemyslaw-wlodek: fixed

emiride (approved):
This now works as expected

emiride (approved):

PR #706, opened 2023-11-07
feat(staking): [LW-8929, LW-8777] apply pool search also to selected pools

xdzurman:
# Checklist

- [x] JIRA - LW-8929, LW-8777
- [ ] Screenshots added.

---

## Proposed solution

Not ideal, but the best for now imho:
```
/**
 * If there's a search value, filter selected pools if they are in the search results.
 * A disadvantage of this approach is that if a current pool is not yet in the search results (it's further down in results),
 *   it will not be shown in the list. The confusing part about this is that if the user scrolls down long enough,
 *   the pool will appear in the upper selected list.
 * Alternative solutions include increasing the search limit / not show selected pools when searching at all,
 *   but this seems like the best solution for now.
 */
```

The only consistent solution IMHO would be to hide selected pools upon search, and when adding a pool, re-set the search and show all selected pools. Other solutions, such as increasing search limit from 10 to N would create problems on their own. Other solution would be to apply the same "search filters", as `stake-pools-search-slice.ts`, but that would cause duplicity and would be fragile

## Testing

Play around with selected pools & searching, see that search is applied also to selectedPools.

### DIFF
--- a/packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
+++ b/packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
@@ -23,6 +23,17 @@ const DEFAULT_SORT_OPTIONS: StakePoolSortOptions = {
 const searchDebounce = 300;
 const defaultFetchLimit = 10;
 
+// imitates apps/browser-extension-wallet/src/stores/slices/stake-pool-search-slice.ts
+const naiveSelectedPoolsSearch = (searchString: string, pools: Wallet.Cardano.StakePool[]) => {
+  const lowerCaseSearchString = searchString.toLowerCase();
+  return pools.filter(
+    (pool) =>
+      pool.metadata?.name.toLowerCase().includes(lowerCaseSearchString) ||
+      pool.metadata?.ticker.toLowerCase().includes(lowerCaseSearchString) ||
+      pool.id.toLowerCase() === lowerCaseSearchString
+  );
+};
+
 export const StakePoolsTable = ({ scrollableTargetId }: StakePoolsTableProps) => {
   const { t } = useTranslation();
   const [searchValue, setSearchValue] = useState<string>('');
@@ -31,9 +42,7 @@ export const StakePoolsTable = ({ scrollableTargetId }: StakePoolsTableProps) =>
   const [sort, setSort] = useState<StakePoolSortOptions>(DEFAULT_SORT_OPTIONS);
   const [stakePools, setStakePools] = useState<Wallet.StakePoolSearchResults['pageResults']>([]);
   const [skip, setSkip] = useState<number>(0);
-  const selectedPortfolioStakePools = useDelegationPortfolioStore((store) =>
-    store.selectedPortfolio.map(({ stakePool }) => stakePool)
-  );
+  const selectedPortfolio = useDelegationPortfolioStore((store) => store.selectedPortfolio);
   const {
     walletStoreWalletUICardanoCoin: cardanoCoin,
     currentChain,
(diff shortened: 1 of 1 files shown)


mrcnk (approved):

packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
```diff
@@ -93,12 +93,27 @@ export const StakePoolsTable = ({ scrollableTargetId }: StakePoolsTableProps) =>
   };
 
   const combinedUnique = useMemo(() => {
-    const combinedStakePools = [...selectedPortfolioStakePools, ...stakePools];
+    const combinedStakePools = [
+      /**
+       * If there's a search value, filter selected pools if they are in the search results.
+       * A disadvantage of this approach is that if a current pool is not yet in the search results (it's further down in results),
```
refi93, line 99: > if a current pool is not yet in the search results (it's further down in results)

I think this currently can't happen as we get the full result and the pagination is done just by infinite scroll at the UI level, or am I missing something?

packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
```diff
@@ -93,12 +93,27 @@ export const StakePoolsTable = ({ scrollableTargetId }: StakePoolsTableProps) =>
...
+       *   it will not be shown in the list. The confusing part about this is that if the user scrolls down long enough,
+       *   the pool will appear in the upper selected list.
+       * Alternative solutions include increasing the search limit / not show selected pools when searching at all,
+       *   but this seems like the best solution for now.
+       */
+      ...(searchValue
+        ? selectedPortfolioStakePools.filter((pool) =>
+            stakePools.find((foundStakePool) => foundStakePool.id === pool.id)
+          )
```
refi93, line 108: I think this can be written more idiomatically using lodash:

```suggestion
        ? intersectionBy(selectedPortfolioStakePools, stakePools, (p) => p.id)
```

https://www.geeksforgeeks.org/lodash-_-intersectionby-method/?ref=lbp

refi93 (approved):
lgtm, just a minor question/improvement suggestion

packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
xdzurman: You actually get just `limit` number of results, so imagine this scenario:
1. you stake to XSEC
2. you search for string `S`
3. you get back 10 results, where none of which is XSEC. So you see no `selected` pools at the top, just the 10 results
4. you scroll down a bit, another page of search results is fetched, XSEC is the 11th overall result
5. XSEC appears at the top among selected pools

packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
```diff
@@ -93,12 +94,25 @@ export const StakePoolsTable = ({ scrollableTargetId }: StakePoolsTableProps) =>
...
 
   const combinedUnique = useMemo(() => {
-    const combinedStakePools = [...selectedPortfolioStakePools, ...stakePools];
+    const combinedStakePools = [
+      /**
+       * If there's a search value, filter selected pools if they are in the search results.
+       * A disadvantage of this approach is that if a current pool is not yet in the search results (it's further down in results),
+       *   it will not be shown in the list. The confusing part about this is that if the user scrolls down long enough,
+       *   the pool will appear in the upper selected list.
+       * Alternative solutions include increasing the search limit / not show selected pools when searching at all,
```
refi93, line 103:
```suggestion
       *   Alternative solutions include increasing the search limit / not show selected pools when searching at all,
```

packages/staking/src/features/BrowsePools/StakePoolsTable/StakePoolsTable.tsx
```diff
@@ -93,12 +94,25 @@ export const StakePoolsTable = ({ scrollableTargetId }: StakePoolsTableProps) =>
...
+       */
+      ...(searchValue
+        ? intersectionBy(selectedPortfolioStakePools, stakePools, (p) => p.id)
+        : selectedPortfolioStakePools),
+      ...stakePools,
+    ];
     const combinedUniqueIds = [...new Set(combinedStakePools.map((pool) => pool.id))];
     return combinedUniqueIds.map((id) =>
       combinedStakePools.find((pool) => pool.id === id)
     ) as Wallet.Cardano.StakePool[];
```
refi93, line 114: I think this can be simplified:
```suggestion
    return combinedStakePools.filter((p) => combinedUniqueIds.includes(p.id))
```

PR #714, opened 2023-11-08
fix(staking): [LW-8689] change wording/name of pool migration banner

refi93:
# Checklist

- [x] JIRA - LW-8689
- [x] Screenshots added.

---

## Proposed solution

This PR just changes the wording of the message but the main bug reported unfortunately remains, explanation below

When changing the delegation portfolio, the `distribution$` observable is apparently updates as soon as the delegation tx is submitted, not waiting for its confirmation (interestingly, only re-delegations seem to be impacted, not first delegations). This results in the bug reported in the ticket, i.e. that user sees the new portfolio even while the delegation tx is pending.

I made a follow-up ticket LW-9025 to fix the `distribution$` observable in the SDK

I heavily considered making a workaround in the frontend - basically "caching" the old distribution while a delegation transaction is pending but I concluded this can do more harm than good. First, it would "work" only for the window from which the tx was sent, and moreover, in case of switching the network to one with a pending delegation tx the distribution from the previous network would remain (as we don't re-initialize the store, AFAIK) which would be also wrong. Future changes in the frontend/sdk may further break this workaround as it would have to make assumptions on otherwise loosely coupled structures (wallet activities vs portfolio distribution)

## Testing

* Make a delegation and check that the update message is shown

## Screenshots

Screenshot 2023-11-08 at 18 27 53

### DIFF
--- a/packages/staking/src/features/i18n/translations/en.ts
+++ b/packages/staking/src/features/i18n/translations/en.ts
@@ -110,9 +110,9 @@ export const en: Translations = {
   'overview.banners.pendingFirstDelegation.message':
     'You will see your staking portfolio here once the transaction has been validated',
   'overview.banners.pendingFirstDelegation.title': 'Your staking transaction has been submitted',
-  'overview.banners.pendingPoolMigration.message':
-    'You will continue to receive rewards from your former stake pool(s) for two epochs',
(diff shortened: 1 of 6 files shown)


przemyslaw-wlodek (approved):
💪🏻

PR #717, opened 2023-11-09
[LW-9000] Add confirmed transaction metric

lucas-barros:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-9000) 
- [ ] Screenshots added.

---

## Proposed solution

Added new analytics event "Transaction confirmed".

New local storage property `unconfirmedTransactions` added to store transactions that have been submitted and have not been confirmed yet. Once new `ObservableWallet.transactions.outgoing.onChain$` event is emitted the entry from local storage is removed. 

For the edge case where the user closes the application before the transaction is confirmed, `ObservableWallet.transactions.history$` is used on start up and matched against `unconfirmedTransactions`.

For the edge case where the user has more than one window opened (browser view, popup or dapp connector) the `Lock API` is used to allow only one event to be processed at the same time.

Lock API support: https://caniuse.com/mdn-api_lock

## Testing
Test submitting transaction from browser view, popup and dapp connector.
Test submitting transaction then close the browser before transaction is confirmed, open the application again and check that the event in posthog is displayed.
Test submitting transaction and having browser view and popup or dapp connector simultaneously then check that the event in posthog is displayed only once.
Test that unconfirmed transactions older than one week are removed from local storage.

## Screenshots


mkazlauskas (approved):
Looks good! Great to see unit tests :heart_eyes:

PR #723, opened 2023-11-13
test(extension): test automation maintenance 13 nov

bslabiak:


ljagiela (approved):

packages/e2e-tests/src/steps/onboardingSteps.ts
```diff
@@ -103,6 +104,7 @@ When(
   /^I click "(Cancel|OK)" button on "(Limited support for DApp|Restoring a multi-address wallet\?|Are you sure you want to start again\?)" modal$/,
   // eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
   async (button: 'Cancel' | 'OK', _modalType: string) => {
+    await browser.pause(500);
```
wklos-iohk, line 107: I'd suggest to use `waitForClickable()` method on particular button we want to click instead of hardcoded pause.
E.g.
```
      case 'Cancel':
        await Modal.cancelButton.waitForClickable();
        await Modal.cancelButton.click();
        break;

wklos-iohk (changes requested):

packages/e2e-tests/src/steps/onboardingSteps.ts
bslabiak: This is not solving the problem because step with clicking are passing correctly, but somehow modal does not disappear.
Screenshot from regression:
image

bslabiak: I was trying to recreate this defect manually, but it was not possible. I assumed that somehow test is being executed to fast and added a delay.

wklos-iohk (approved):

PR #728, opened 2023-11-15
feat(ui): [LW 9028] Multi wallet UI elements

renanvalentin:
# Checklist

- [x] JIRA - https://input-output.atlassian.net/browse/LW-9028
- [x] Screenshots added.

---

## Proposed solution

This is a partial implementation of the personal dropdown from Lace UI's toolkit. Below is a possible API for the final implementation, which will allow users to compose the component as they see fit:

```ts
<PersonalDropdown.Root>
   <PersonalDropdown.Trigger />
   <PersonalDropdown.Popup>
        <PersonalDropdown.WalletOption />
        <PersonalDropdown.Divider />
        <PersonalDropdown.Option />
   </PersonalDropdown.Popup>
</PersonalDropdown.Root>
```

Currently, only these components are implemented and can be used on the main application:

```
<PersonalDropdown.Trigger />
<PersonalDropdown.WalletOption />
```

## Screenshots

capture-41701a94
capture-bd4afb55
capture-c8d03858


packages/ui/src/design-system/personal-dropdown/personal-dropdown-wallet-option.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+    fallback: string;
+    alt?: string;
+    delayMs?: number;
+  };
+};
+
+export const WalletOption = ({
+  id,
+  disabled,
+  open = false,
```
lucas-barros, line 31: open is unused.

lucas-barros (approved):

packages/ui/src/design-system/personal-dropdown/personal-dropdown-wallet-option.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+  subtitle: string;
+  profile: {
+    imageSrc: string;
+    fallback: string;
+    alt?: string;
+    delayMs?: number;
+  };
+};
+
+export const WalletOption = ({
```
DominikGuzei, line 28: this components seem to be very specific for a single use case in Lace, maybe lets move it into the Lace app repository and setup Storybook there? And also it seems to be a near duplicate to the `Trigger` component?

renanvalentin: The items mentioned actually belong to the UI Toolkit. The UX team is currently updating the `Navigation & Toolbars` section to include the latest specifications. 

From these specs, you can create a `LinkOption` or `ToggleOption`. It might be a good idea to rename `WalletOption` to `ProfileDetailOption` to make it more general. However, I personally believe that we should stick to something that conveys the true purpose, which is a wallet.

Screenshot 2023-11-16 at 08 53 17

DominikGuzei: Got it 👍 Yeah, i still think that a ui framework shouldn't suggest the final app component being built from framework parts. We had this discussion many times in relation to `react-polymorph` in Daedalus and that's why we kept app specific components which could not be generalized (or didn't have to be re-used) in Daedalus Storybook.

I'm OK with picking more general purpose names (`LinkOption` or `ToggleOption`) like you suggested. Otherwise it becomes confusing because we could start moving any ui component from Lace into `@lace/ui` which would defeat the purpose of creating a foundational ui library used by many apps (like Dex, DApps store etc. potential more in the future)

DominikGuzei:
@renanvalentin can you explain the difference between the `Trigger` component and the `WalletOption` as they look identical on the screenshots?

renanvalentin:
@lucas-barros @DominikGuzei ready for review

packages/ui/src/design-system/search-box/search-box.stories.tsx
```diff
@@ -6,7 +6,7 @@ import { within, userEvent } from '@storybook/testing-library';
 
 import { ThemeColorScheme, ThemeProvider } from '../../design-tokens';
```
lucas-barros, line 7: `ThemeProvider` and `ThemeColorScheme` are not used.

DominikGuzei (approved):
great work 🎉

PR #764, opened 2023-11-29
[LW-9186]Fix: take in account multidelegation transactions with inputs that share same txID

VanessaPC:
# Checklist

- [x] JIRA - LW-9186
- [x] Screenshots added.

---

## Proposed solution
On `gettingTxInputsValueAndAddress` we were only working under the idea that a transaction could have inputs with different txID's, whereas in multidelegation transactions the txID's are all equal, and the addresses are not. Therefore we  take this consideration into account and resolve the values and addresses for the inputs to match the same data in CExplorer.

## Testing
- Open any transaction and check that the inputs match to CExplorer
- Create a delegation transaction and check the inputs match 
- Update your delegation portfolio and check the inputs again 

## Screenshots
Screenshot 2023-11-29 at 16 39 52
Screenshot 2023-11-29 at 16 39 33
Screenshot 2023-11-29 at 16 39 24


DominikGuzei (approved):
great work 🎉

bslabiak (approved):

PR #775, opened 2023-12-06
refactor(extension): use isFeatureEnabled

pczeglik-iohk:
## Proposed solution

I'm proposing a new (slightly updated) way of using feature flags in the browser extension code base.

Instead of reaching out directly in code to `process.env`, I've created a helper method `isFeatureEnabled` which is an abstraction  layer on top of `process.env`. It does the same thing, but in more maintainable way. In the future if we decide to use other solution for feature flags it will be easier to refactor.

Apart of easier usage and better semantic there is one more benefit of this approach. `isFeatureEnabled` throws Error in case feature flag is not defined in `.env`.

Ex. Code is using `USE_HIDE_MY_BALANCE` feature flag, but it is not defined in jest's config file `set-env-vars.js`

```
FAIL  src/features/address-book/hooks/__tests__/useGetFilteredAddressBook.test.tsx
  ● Test suite failed to run

    'HIDE_MY_BALANCE' is not set

      29 | export const isFeatureEnabled = (name: FeatureFlag): boolean => {
      30 |   if (!process.env[`USE_${name}`]) {
    > 31 |     throw new Error(`'USE_${name}' is not set`);
         |           ^
      32 |   }
      33 |   return process.env[`USE_${feature}`] === 'true';
      34 | };
```

Additionally I configured jest to use the same `.env` file as engineer working with the code. In case there is no `.env` then `.env.defaults` will be used.

How to use this helper?

1. Define feature flag in `.env.defaults` and `.env.example` (please keep flags sorted) ex. `USE_MY_FEATURE=true`
2. Extend `FeatureFlag` type in `utils/feature-flags.ts` (please keep flags sorted) ex. `MY_FEATURE`
3. In code instead of using `process.env.USE_MY_FEATURE === 'true'` use `isFeatureEnabled('MY_FEATURE')`


apps/browser-extension-wallet/.env.defaults
```diff
@@ -10,25 +10,24 @@ TOKEN_PRICE_POLLING_IN_SEC=300
 SAVED_PRICE_DURATION_IN_MINUTES=720
 
 # Feature Flags
-USE_PASSWORD_VERIFICATION=false
+USE_ADA_HANDLE=true
```
pczeglik-iohk, line 13: This is only sorting. It would be good to have such things sorted so it's easier to manage

apps/browser-extension-wallet/test/__mocks__/set-env-vars.js
```diff
@@ -1,14 +1,17 @@
-process.env.CARDANO_SERVICES_URL_PREVIEW = 'https://preview-prod.com';
-process.env.CARDANO_SERVICES_URL_PREPROD = 'https://preprod-prod.com';
-process.env.CARDANO_SERVICES_URL_MAINNET = 'https://mainnet-url.com';
+require('dotenv-defaults').config({
```
pczeglik-iohk, line 1: Use `.env` file with fallback to `.env.defaults` instead of manually maintained list of feature flags and env vars

PR #781, opened 2023-12-11
[LW-9157] Cancel adding new wallet

lucas-barros:
# Checklist

- [x] JIRA - https://input-output.atlassian.net/browse/LW-9157
- [x] Screenshots added.

---

## Proposed solution

This PR add a cancel new wallet creation dialog confirmation.
Before the user has entered any data, the cancellation should happen without any confirmation.
After the user has entered some data, the confirmation modal should appear.

## Testing

## Screenshots

https://github.com/input-output-hk/lace/assets/11845759/da96c5b4-0dc8-4feb-8fa4-4bf5846b4077


apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/MultiWallet.tsx
```diff
@@ -71,33 +74,36 @@ export const MultiWallet = (): JSX.Element => {
...
-      <Modal
-        centered
-        closable={false}
-        // eslint-disable-next-line unicorn/no-null
-        footer={null}
-        open
-        width="100%"
-        className={styles.modal}
-      >
+      <Dialog.Root open={isDialogOpen} setOpen={setIsDialogOpen} zIndex={1000}>
```
VanessaPC, line 87: Is it possible to move the Dialog into its own component so the Multiwallet.tsx file mostly only cares about the actions about it?

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/useCancelDialog.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+
+export const useCancelDialog = (
```
VanessaPC, line 3: `useCancelDialog` seems very specific, it's a corner or a small dialogue, that can be used for a cancel action, but also any other action. We can just pass any strings into it. 
Making it this way would also make more sense to move it into the `hooks` folder so it can be reused somewhere else in the project, otherwise, we would have to create a hooks folder inside multi-wallet, which could be less necessary. WDYT?

VanessaPC (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/useCancelDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  setRef: (instance: HTMLElement) => void;
+  closeWithDialog: () => void;
+  withReset: (WrappedComponent: React.FC) => () => JSX.Element;
+} => {
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+  const shouldOpenDialogRef = useRef(false);
+  const containerRef = useRef<HTMLElement>(null);
+
+  const handleInputChange = useCallback(() => {
+    shouldOpenDialogRef.current = true;
```
renanvalentin, line 17: What if we create a Prompt component for handling this situation in a more declarative manner? For instance, when the user inputs a value but then clears it out, we should set `shouldOpenDialogRef.current` as false. Additionally, with regards to a hardware wallet, shouldn't we display the dialog box after the user connects the wallet but then closes it?

```tsx
<PromptProvider>
      <Prompt title={t('multiWallet.cancelDialog.title')} subtitle={{t('multiWallet.cancelDialog.description')}} />
      <WalletSetupFlowProvider flow={WalletSetupFlow.ADD_WALLET}>
      <Modal centered closable={false} footer={null} open={!isDialogOpen} width="100%" className={styles.modal}>
        <div className={styles.closeButton}>
          <NavigationButton icon="cross" onClick={closeWithDialog} />
        </div>
        <div ref={setRef}>
          <Switch>
            <Route path={newWallet.create.root} component={SetupCreateWallet} />
            <Route path={newWallet.hardware.root} component={SetupHardwareWallet} />
            <Route path={newWallet.restore.root} component={SetupRestoreWallet} />
            <Route exact path={`${path}/`} component={withReset(Home)} />
          </Switch>
        </div>
    </WalletSetupFlowProvider>

</PromptProvider>
```

Each flow could supply an observable `state$` to the providers property, which would allow you to display the `prompt`. e.g:

```tsx
export const SetupRestoreWallet = (): JSX.Element => {
   const [prompt, setPrompt]=  usePrompt()
   const state$ = useMemo(() => new Observable(), []);

 useEffect(() => {
    const subscription = state$.subscribe((state) => {
      if (state is Dirty) {
        setPrompt(true)
      }
    });

    return () => {
      subscription.unsubscribe();
    };
  }, [state$]);


 return (
  <RestoreWallet
    providers={{
      createWallet
    }}
  />
)
};
```

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/useCancelDialog.tsx
lucas-barros: Are you suggestion a name change? Or to make it the logic more generic?

VanessaPC: I suggest considering both. As well as moving the hook into a folder for hooks.

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/useCancelDialog.tsx
renanvalentin: You could trigger from the context: https://github.com/input-output-hk/lace/blob/feat/lw-9157-cancel-adding-new-wallet/apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/create-wallet/context.tsx#L38

The core logic is centralized inside the context.ts file, so it should be easier to emit a new state change whenever something changes.

lucas-barros: If I'm not mistaken, those state changes only get triggered once the user goes from one step to the other, not on input event. I think we'll need a combination of both approaches.

lucas-barros: @renanvalentin I'm still going to make some improvements, but I updated the PR to incorporate some of your suggestions. Let me know what you think.

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/useCancelDialog.tsx
lucas-barros: I incoporated some of your and Renan's suggestion and moved the Dialog to a provider along with the hook logic.

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/create-wallet/steps/Setup.tsx
```diff
@@ -7,10 +7,11 @@ import { useCreateWallet } from '../context';
 export const Setup = (): JSX.Element => {
   const history = useHistory();
   const { setName, setPassword } = useCreateWallet();
+  const { withConfirmationDialog } = useWalletSetupConfirmationDialog();
```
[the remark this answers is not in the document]
renanvalentin: EDIT: It may be possible to remove the `WalletSetupConfirmationDialogProvider` component altogether. 

```tsx
interface Props {
  onProgressChange: () => 'dirty' | 'none' 
}

export const SetupCreateWallet = ({onProgressChange}: {onChang}): JSX.Element => {
   const state$ = useMemo(() => new Observable(), []);

    useEffect(() => {
    const subscription = state$.subscribe((state) => {
      if (state is Dirty) {
        onProgressChange('dirty')
      }
    });

    return () => {
      subscription.unsubscribe();
    };
  }, [state$]);

  return (
  <CreateWallet
    providers={{
      createWallet,
      generateMnemonicWords: Wallet.KeyManagement.util.generateMnemonicWords,
     state$
    }}
  />
  );
};


export const MultiWallet = (): JSX.Element => {
  /*...*/
const onProgressChange = (status) => setState({progress: status })

const leave = () => {
     setBackgroundPage();
              history.push(page);
}

  return (
<>
    /*.... */
       <div className={styles.closeButton}>
          <NavigationButton
            icon="cross"
            onClick={() => {
            if state.progress is 'none'
              leave()
            else
                 setState((s) => ({
                ...s,
                showLeavingDialog: true
              }))
            }}
          />
        </div>
          <Route path={newWallet.create.root} component={SetupCreateWallet} onProgressChange={}/>
   /* ... */
   {state. showLeavingDialog && <Dialog onConfirm={() => {
              leave()
} } />
<>
  );
};
```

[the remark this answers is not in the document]
lucas-barros: I think we still need the provider to cover the OnBack dialog scenario. Without it we would have to keep a local state for the dialog for each step and it would not be possible to hide the flow modal. We would display the dialog on top of the modal.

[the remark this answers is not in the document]
renanvalentin: > I think we still need the provider to cover the OnBack dialog scenario. Without it we would have to keep a local state for the dialog for each step and it would not be possible to hide the flow modal. We would display the dialog on top of the modal.

See this comment: https://github.com/input-output-hk/lace/pull/781#discussion_r1426754511

We only need to apply this logic to the initial screen.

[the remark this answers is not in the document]
lucas-barros: @renanvalentin updated.

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/create-wallet/CreateWallet.test.tsx
```diff
@@ -64,4 +73,30 @@ describe('Multi Wallet Setup/Create Wallet', () => {
     await keepWalletSecureStep();
     await recoveryPhraseStep();
   });
+
+  test.only('should emit correct value for shouldShowDialog', (done) => {
```
renanvalentin, line 77: remove `test.only`

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/create-wallet/CreateWallet.test.tsx
```diff
@@ -64,4 +73,30 @@ describe('Multi Wallet Setup/Create Wallet', () => {
...
+    const nameInput = screen.getByTestId('wallet-name-input');
+
+    const firstSub = providers.confirmationDialog.shouldShowDialog$.subscribe((value) => {
+      expect(value).toBe(true);
+      firstSub.unsubscribe();
+    });
+
+    fireEvent.change(nameInput, { target: { value: 'My X Wallet' } });
+
+    const secondSub = providers.confirmationDialog.shouldShowDialog$.subscribe((value) => {
```
renanvalentin, line 94: how about using `firstValueFrom` to keep a flat control flow?

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/create-wallet/CreateWallet.test.tsx
lucas-barros: Thanks, updated.

apps/browser-extension-wallet/src/views/browser-view/features/multi-wallet/create-wallet/CreateWallet.test.tsx
lucas-barros: Thanks, updated.

renanvalentin (approved):
great job @lucas-barros

pczeglik-iohk (approved):
Looks good. Thank you @lucas-barros

VanessaPC (approved):
Great work!

PR #792, opened 2023-12-19
feat: add developer preview build tooling

mchappell:
# Checklist

- [x] JIRA - [LW-9180](https://input-output.atlassian.net/browse/LW-9180)
- [ ] Screenshots added.

---

## Proposed solution

Add a dedicated GitHub action to build the developer preview build of Lace

## Testing

- Existing functionality should not be broken when [running existing e2e tests](https://github.com/input-output-hk/lace/actions/runs/7260753199)
- Run the GitHub action against a specific branch, and assert that the `.env.devpreview` parameters have been used

[LW-9180]: https://input-output.atlassian.net/browse/LW-9180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/shared/build/action.yml
```diff
@@ -20,6 +20,10 @@ inputs:
     description: 'Enable analytics tracking in production'
     required: false
     default: 'false'
+  BUILD_DEV_PREVIEW:
+    description: 'Should build developer preview version?'
+    required: false
+    default: 'false'
```
przemyslaw-wlodek, line 26: I have a couple of concerns.
1. This variable description looks like it intends to be used as a flag to decide whether to build a dev preview or not but in the `build-dev-preview.yml` I see it's always running the build (there are no `if` statements for workflow).
2. I feel like we should have different `inputs` defined and used in the environment variables during the build instead of replacing the defaults file for `dotenv-defaults`.
3. Can we please remove decrypt/lint/test steps from the workflow and leave just build/upload? We are doing it in Build and Test workflow. Maybe we could just make dev preview build dependant on the success of that job?

Apologies but I'm not an expert with github actions (and best practices) so my concerns may be unfounded.

mchappell: 1. this means it'll only set the env variable to build the developer preview version if this is set, in this case that is only enabled in the `build-dev-preview.yml` so this is what i'd expect
2. This would only work for CI, and it's useful to have these changes explicitly set, rather than declaring overrides in local `.env` files
3. Unfortunately not, we'd still want to lint even the dev preview version and test it. and the decrypt step is required and isn't covered in the `action.yml` file. and is a requirement of the linting, I'd be happy to cover this workflow refactoring in a secondary change to streamline all the workflows

przemyslaw-wlodek: 1. Ok. I just noticed that the description is different than for Analytics. It could be i.e. `Enable developer preview build`
2. 👍🏻 
3. It's still running before the dev preview build (I mean we always run regular `CI / Build & Test` before)

przemyslaw-wlodek (approved):
Just one comment. LGTM 🚀

bernokl (approved):

.github/shared/build/action.yml
```diff
@@ -20,6 +20,10 @@ inputs:
     description: 'Enable analytics tracking in production'
     required: false
     default: 'false'
+  BUILD_DEV_PREVIEW:
```
pczeglik-iohk, line 23: If this is not breaking anything I would keep those inputs sorted alphabetically :)

.github/workflows/build-dev-preview.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          BUILD_DEV_PREVIEW: 'true'
+          LACE_EXTENSION_KEY: ${{ secrets.DEV_PREVIEW_MANIFEST_PUBLIC_KEY  }}
+          PRODUCTION_MODE_TRACKING: 'false'
+      - name: Check for linter issues
+        run: yarn lint
+      - name: Run unit tests, generate test coverage report
+        env:
+          AVAILABLE_CHAINS: 'Preprod,Preview,Mainnet'
+          DEFAULT_CHAIN: 'Preprod'
```
pczeglik-iohk, line 29: Is `Preprod` correct here? Should it be `Preview` maybe?

apps/browser-extension-wallet/webpack-utils.js
```diff
@@ -1,11 +1,18 @@
...
-  path: './.env',
-  encoding: 'utf8',
-  defaults: './.env.defaults'
-});
-
 const transformManifest = (content, mode) => {
+  require('dotenv-defaults').config({
+    path: './.env',
+    encoding: 'utf8',
+    defaults: process.env.BUILD_DEV_PREVIEW === 'true' ? './.env.devpreview' : './.env.defaults'
```
pczeglik-iohk, line 5: Can we rename this `.env.devpreview` file to `.env.dev.preview`? It improves slightly readability imo

apps/browser-extension-wallet/webpack-utils.js
```diff
@@ -1,11 +1,18 @@
...
+  });
+
+  const yy = new Date();
+  yy.getTime();
```
pczeglik-iohk, line 9: do we need `yy`?

apps/browser-extension-wallet/webpack.common.dev.js
```diff
@@ -43,6 +39,14 @@ module.exports =
         devtool: 'inline-source-map',
         cache: { type: 'filesystem' },
         plugins: [
+          new Dotenv({
+            path: '.env',
+            safe: false,
+            silent: false,
+            defaults: process.env.BUILD_DEV_PREVIEW === 'true' ? '.env.devpreview' : true,
```
pczeglik-iohk, line 46: If you decide to rename this file, please update it in places like this one

pczeglik-iohk (commented):
Have a look at my comments please @mchappell

apps/browser-extension-wallet/webpack-utils.js
```diff
@@ -1,11 +1,18 @@
...
   const manifest = JSON.parse(content.toString());
```
danielmain, line 10: try catch for JSON.parse?

apps/browser-extension-wallet/webpack-utils.js
```diff
@@ -1,11 +1,18 @@
...
+  manifest.name = manifest.name.replace('$WALLET_MANIFEST_NAME', process.env.WALLET_MANIFEST_NAME);
```
danielmain, line 11: chaining operator for `manifest?.name?.replace`

apps/browser-extension-wallet/webpack-utils.js
```diff
@@ -1,11 +1,18 @@
...
+  if (process.env.BUILD_DEV_PREVIEW === 'true') {
+    const date = new Date();
+    manifest.version = `${manifest.version}.${date.getTime()}`;
```
danielmain, line 14: I would write it like this: 
```
manifest.version = `${manifest.version}.${Date.now()}`;
```

no need for like 13

danielmain (changes requested):
small changes, great job!

.github/workflows/build-dev-preview.yml
mchappell: Preprod is correct here, the unit tests run in an identical way to current `ci.yml` step

apps/browser-extension-wallet/webpack-utils.js
mchappell: how about `.env.developerpreview` as I want to distinguish that these aren't defaults for the `preview` network?

apps/browser-extension-wallet/.env.devpreview
```diff
@@ -0,0 +1,81 @@
...
+USE_POSTHOG_ANALYTICS_FOR_OPTED_OUT=false
+USE_MATOMO_ANALYTICS_FOR_OPTED_OUT=false
+
+# In App URLs
+CATALYST_GOOGLE_PLAY_URL=https://play.google.com/store/apps/details?id=io.iohk.vitvoting
+CATALYST_APP_STORE_URL=https://apps.apple.com/fr/app/catalyst-voting/id1517473397?l=en
+TWITTER_URL=https://twitter.com/lace_io
+YOUTUBE_URL=https://www.youtube.com/c/Laceio
+MEDIUM_URL=https://medium.com/@lace_wallet
+GITHUB_URL=https://github.com/input-output-hk
```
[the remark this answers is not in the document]
mchappell: it's identical to the current defaults, we can raise another ticket to change this if required

apps/browser-extension-wallet/webpack-utils.js
pczeglik-iohk: It makes sense. Thank you

apps/browser-extension-wallet/.env.devpreview
```diff
@@ -0,0 +1,81 @@
...
+DISCORD_URL=https://discord.gg/inputoutput
```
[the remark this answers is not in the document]
mchappell: updated with latest merge from `main` in 12327ac94840cf583d176fb122d0b7b1bc447f2e

apps/browser-extension-wallet/webpack-utils.js
mchappell: done in 12327ac94840cf583d176fb122d0b7b1bc447f2e

apps/browser-extension-wallet/webpack-utils.js
mchappell: as this is a required manifest field, I don't think this is necessary, do you?

apps/browser-extension-wallet/webpack-utils.js
mchappell: done in 12327ac94840cf583d176fb122d0b7b1bc447f2e

apps/browser-extension-wallet/webpack.common.dev.js
mchappell: updated in 12327ac94840cf583d176fb122d0b7b1bc447f2e

apps/browser-extension-wallet/webpack-utils.js
mchappell: no, thanks for spotting this, removed in 12327ac94840cf583d176fb122d0b7b1bc447f2e

apps/browser-extension-wallet/webpack-utils.js
danielmain: you have a try catch, not so necessary ;)

apps/browser-extension-wallet/webpack-utils.js
danielmain: but you are still defining one line before :D 
```
const date = new Date();
```

danielmain: CleanShot 2024-01-04 at 10 17 08@2x

pczeglik-iohk (approved):
Looks good. Thank you @mchappell

apps/browser-extension-wallet/webpack-utils.js
mchappell: 🤦 fixed in 144a7b0783c03aa0dde5a05e231936d45e3e6b4f

danielmain (approved):

PR #800, opened 2023-12-21
chore(staking): migrate from ladle to storybook [LW-9454]

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-9454)

---

## Proposed solution
This PR enables storybook development in Staking. Previously we used Ladle but due to the testing approach (using Chromatic) we are bringing Storybook back in place as a standard tool.

## Testing
- QA: Not required.
- DEV: Please run the storybook locally.


packages/staking/.storybook/StakingStorybookProvider.tsx
```diff
@@ -0,0 +1,60 @@
+// @ts-ignore
+import { SetupBase } from '../src/features/staking/Setup/SetupBase';
+// @ts-ignore
```
przemyslaw-wlodek, line 3: Not sure how to solve it elegantly. `baseUrl` is set to `src` dir and this file is just a mock provider for storybook (outside of `src` dir). I think we could probably add `include` to `tsconfig.json` but this may require explicit mention on src files again. I see no harm in that for now.

packages/staking/.storybook/main.js
```diff
@@ -0,0 +1,113 @@
...
+    'storybook-addon-pseudo-states',
+  ],
+  framework: '@storybook/react',
+  webpackFinal: (config) => {
+    config.plugins = [
+      ...config.plugins,
+      new VanillaExtractPlugin(),
+      new MiniCssExtractPlugin(),
+      // TODO make it work. Then we can bring back aliases if needed `../../../outside-handles-provider` -> `features/outside-handles-provider`
+      // new TsconfigPathsPlugin({ baseUrl: path.resolve(__dirname, '../src') }),
```
przemyslaw-wlodek, line 21: There were ongoing discussions if we even wanted to have the aliases/baseUrl (how much value vs tooling problems it gives). If we want then we need to fix this, otherwise I can create a separate task to migrate from paths `features/x` to `../../x` for example and remove `baseUrl` config.

vetalcore:
running `yarn story:build` fails for me with:
```
ERR! Module build failed (from ./node_modules/@vanilla-extract/webpack-plugin/loader/dist/vanilla-extract-webpack-plugin-loader.cjs.js):
ERR! Error: No module factory available for dependency type: CssDependency
```

packages/staking/package.json
```diff
@@ -33,18 +32,16 @@
...
     "eslint:check": "yarn run -T eslint --cache --cache-location ../../.cache/eslintcache --cache-strategy metadata --ignore-path '../../.eslintignore' '**/*.{cjs,mjs,js,ts,tsx}'",
     "eslint:fix": "yarn eslint:check --fix",
     "lint": "echo 'Staking package runs lint in its own CI'",
     "lint:all": "yarn eslint:check && yarn prettier:check",
     "lint:fix": "yarn eslint:fix && yarn prettier:fix",
     "prettier:check": "yarn run -T prettier --check '**/*.{js,ts,jsx,tsx,html,json,md}'",
     "prettier:fix": "yarn run -T prettier --write --loglevel warn '**/*.{js,ts,jsx,tsx,html,json,md}'",
-    "story:build": "ladle build",
-    "story:preview": "ladle preview",
+    "story:build": "build-storybook",
```
vetalcore, line 44: would i make sence to have same steps as in the browser-extension app?
```suggestion
        "storybook": "ENV=development start-storybook -p 6006",
    "prestorybook:build": "yarn build",
    "storybook:build": "ENV=development build-storybook --log-level error",
```

vetalcore (changes requested):

packages/staking/.storybook/StakingStorybookProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { OutsideHandlesContextValue, OutsideHandlesProvider } from '../src/features/outside-handles-provider';
+import React, { ReactNode } from 'react';
+import { ThemeColorScheme } from '@lace/ui';
+
+type StakingStorybookProviderProps = {
+  children: ReactNode;
+};
+
+// Extend mocks if needed
+const outsideHandlesMocks: OutsideHandlesContextValue = {
```
vetalcore, line 13:
```suggestion
const outsideHandlesMocks = {
```

packages/staking/.storybook/StakingStorybookProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+  walletStoreBlockchainProvider: undefined,
+  currencyStoreFiatCurrency: undefined,
+  compactNumber: undefined,
+  multidelegationFirstVisit: undefined,
+  triggerMultidelegationFirstVisit: undefined,
+  multidelegationFirstVisitSincePortfolioPersistence: undefined,
+  triggerMultidelegationFirstVisitSincePortfolioPersistence: undefined,
+  walletAddress: undefined,
+  currentChain: undefined,
+};
```
vetalcore, line 51:
```suggestion
} as unknown as OutsideHandlesContextValue;
```

packages/staking/.storybook/StakingStorybookProvider.tsx
vetalcore: somehow it works fine if i remove `// @ts-ignore` comments here

vetalcore (changes requested):

tommayeliog (approved):

packages/staking/package.json
przemyslaw-wlodek: I've updated scripts now we have it `build` and `build:storybook`. I hope it makes more sense now. `pre*` step in this case would be a killer. I cannot imagine building all packages each time I run storybook. Everyone has to do it manually if needed 🙏🏻

packages/staking/.storybook/StakingStorybookProvider.tsx
przemyslaw-wlodek: Outside handles should be removed in the future when we migrate these functions/hooks/etc to `staking`. Let's leave it as it is. Now we explicitly see which functionalities aren't mocked as opposed to the type-casting you proposed.

vetalcore (approved):

mchappell (approved):

packages/staking/.storybook/main.js
przemyslaw-wlodek: Fixed 🎉

wklos-iohk (approved):

PR #805, opened 2023-12-27
feat(staking): add feature flag for new grid view

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-9266)

---

## Proposed solution
Add a new feature flag for staking that will enable switching between Grid and List view when browsing pools.

## Testing
To be covered in the other PR (Grid View).


apps/browser-extension-wallet/.env.example
```diff
@@ -28,6 +28,7 @@ USE_POSTHOG_ANALYTICS_FOR_OPTED_OUT=false
 USE_MATOMO_ANALYTICS_FOR_OPTED_OUT=false
 USE_COMBINED_PASSWORD_NAME_STEP_COMPONENT=false
 USE_MULTI_DELEGATION_STAKING_ACTIVITY=true
+USE_MULTI_DELEGATION_STAKING_GRID_VIEW=false
```
vetalcore, line 31:
```suggestion
USE_MULTI_DELEGATION_GRID_VIEW=false
```
wdyt?

vetalcore (changes requested):

apps/browser-extension-wallet/.env.example
przemyslaw-wlodek: I'd say yes but I want to be consistent with all variables we have. 😄 I think it'd be worth renaming after we drop the old single-staking impl.
```
USE_MULTI_DELEGATION_STAKING_LEDGER=false
USE_MULTI_DELEGATION_STAKING_TREZOR=false
...
USE_MULTI_DELEGATION_STAKING_ACTIVITY=true
```

vetalcore: could you please leave a todo then?

vetalcore (approved):

apps/browser-extension-wallet/.env.example
przemyslaw-wlodek: Done :)

mchappell (approved):

PR #807, opened 2024-01-02
[LW-9482] Fix/cip30 sign data

### LINKED ISSUE #803: Cannot sign message to get signature
I used signMessage in useCardano function to get the wallet's signature. but it reported missing password error
image

mchappell:
# Checklist

- [ ] JIRA - \<link>
- [x] Screenshots added.

---

## Proposed solution

`signBlob` method from keyAgent is now also wrapped, so that the auto-redirect to sign success state can be removed, this was prematurely removing password from memory which caused the blob not to sign

## Testing

Execute at example.com in console

```
const api = await window.cardano.lace.enable();
const ownAddress = await api.getUsedAddresses();
const signedData = await api.signData(
        ownAddress[0],
        '66697865642074686520627567'
      );
```
## Screenshots

image
image

Fixes https://github.com/input-output-hk/lace/issues/803

### DIFF
--- a/packages/cardano/src/wallet/lib/cardano-wallet.ts
+++ b/packages/cardano/src/wallet/lib/cardano-wallet.ts
@@ -127,7 +127,7 @@ export const createCardanoWalletsByChain = async (
   return { wallet, keyAgent: activeKeyAgent, keyAgentsByChain };
 };
 
-const keyagentSigningCallback = (signCallback?: (result: boolean) => void, response?: boolean) => {
+const executeSigningCallbackIfExists = (signCallback?: (result: boolean) => void, response?: boolean) => {
   if (signCallback)
     setTimeout(() => {
       signCallback(response);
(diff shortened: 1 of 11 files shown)


lucas-barros (approved):

packages/cardano/src/wallet/lib/cardano-wallet.ts
```diff
@@ -127,31 +127,45 @@ export const createCardanoWalletsByChain = async (
   return { wallet, keyAgent: activeKeyAgent, keyAgentsByChain };
 };
 
+const keyagentSigningCallback = (signCallback?: (result: boolean) => void, response?: boolean) => {
```
pczeglik-iohk, line 130: Can we slightly improve the name of this function:`keyAgentSigningCallback`?

pczeglik-iohk (approved):
I spot one small think, but it's not a deal breaker :)

packages/cardano/src/wallet/lib/cardano-wallet.ts
mchappell: `executeSigningCallbackIfExists` ?

ljagiela (approved):

danielmain (approved):

packages/cardano/src/wallet/lib/cardano-wallet.ts
pczeglik-iohk: Sounds good to me :)

wklos-iohk (approved):

PR #816, opened 2024-01-05
refactor(extension): [LW-9094] use new wallet manager

mkazlauskas:
In order to support multi-wallet/account & shared wallets, walletManager was completely redesigned in SDK.

There are a couple new components:
- WalletRepository stores key agent data, used accounts, as well as any additional metadata such as 'name'. Lace should use it to get an array of all users wallets.
- SignerManager is an abstraction over KeyAgent. Lace no longer needs to maintain the lifecycle of a KeyAgent. Instead, it should use this component to sign transactions.

For more info, see [cardano-js-sdk documentation](https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/web-extension/README.md)

There are also some other changes due to SDK update:
- Transaction inspectors are now async

# Checklist

- [x] JIRA LW-9094 LW-9533
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


rhyslbw (commented):
Looks good from a high level, although coveralls is reporting a drop of 3.5% test coverage which is out of the allowed threshold. We'll also need to run full regression testing on this PR @ljagiela to have full confidence, given it's a huge refactor.

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -192,17 +164,34 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
...
-        setTx(backgroundTx);
-      })
-      .catch((error) => {
-        setErrorMessage(error);
-        console.error(error);
-      });
-  }, []);
+    console.warn('e2e-debug', 'subscribe');
+    const subscription = signingCoordinator.transactionWitnessRequest$.subscribe(async (r) => {
+      console.warn('e2e-debug', 'got req', r);
```
DominikGuzei, line 169: do we want to keep these `console.log` statements?

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -101,10 +92,10 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
...
 
-  const [tx, setTx] = useState<Wallet.Cardano.Tx>();
   const assets = useObservable<TokenInfo | null>(inMemoryWallet.assetInfo$);
-  const [errorMessage, setErrorMessage] = useState<string>();
+  // const [errorMessage, setErrorMessage] = useState<string>();
   const redirectToSignFailure = useRedirection(dAppRoutePaths.dappTxSignFailure);
+  const redirectToSignSuccess = useRedirection(dAppRoutePaths.dappTxSignSuccess);
   const [isConfirmingTx, setIsConfirmingTx] = useState<boolean>();
   const keyAgentType = getKeyAgentType();
   const isUsingHardwareWallet = useMemo(
```
DominikGuzei, line 101: Are we still using key agents? I thought this was replaced by the new `SigningCoordinator` – or is this something else?

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -249,57 +238,66 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
...
+      });
 
-    // eslint-disable-next-line consistent-return
-    return {
-      fee: Wallet.util.lovelacesToAdaString(tx.body.fee.toString()),
-      outputs: txSummaryOutputs,
-      type: txType,
-      mintedAssets: createMintedList(minted),
-      burnedAssets: createMintedList(burned)
+      // eslint-disable-next-line unicorn/no-array-reduce
```
DominikGuzei, line 267: Maybe we should disable `unicorn/no-array-reduce` if we still want to use it?

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  if (txType === TxType.VotingProcedures) {
+    return 'core.votingProcedures.title';
+  }
+
+  return sectionTitle[DAPP_VIEWS.CONFIRM_TX];
+};
+
+export const disallowSignTx = (close = false): void => {
```
DominikGuzei, line 45: It would be great to document this function as I don't understand what it does exactly

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+      }),
+      baseChannel: DAPP_CHANNELS.userPrompt,
+      properties: { allowSignTx: RemoteApiPropertyType.MethodReturningPromise }
+    },
+    { logger: console, runtime }
+  );
+  close && setTimeout(() => window.close(), DAPP_TOAST_DURATION);
+};
+
+export const allowSignTx = (): void => {
```
DominikGuzei, line 61: It would be great to document this function with a comment as I don't understand what it does exactly

apps/browser-extension-wallet/src/features/unlock-wallet/components/UnlockWalletContainer.tsx
```diff
@@ -84,13 +68,23 @@ export const UnlockWalletContainer = ({ validateMnemonic }: UnlockWalletContaine
...
   useKeyboardShortcut(['Enter'], onUnlock);
 
-  return validateMnemonic && unlocked ? (
-    <MnemonicValidation
-      plainPassword={password}
-      walletKeyAgent={unlocked[environmentName]?.keyAgentData}
-      onValidationSuccess={loadWallet}
-    />
-  ) : (
+  // LW-9500
```
DominikGuzei, line 71: Do we want to keep this commented-out code?

apps/browser-extension-wallet/src/lib/scripts/background/services/expose.ts
```diff
@@ -0,0 +1,19 @@
+import { exposeApi } from '@cardano-sdk/web-extension';
+import { userIdServiceProperties } from '../config';
+import * as wallet from '../wallet';
+import { UserIdService } from './userIdService';
+import { USER_ID_SERVICE_BASE_CHANNEL, UserIdService as UserIdServiceInterface } from '@lib/scripts/types';
+import { of } from 'rxjs';
+import { runtime } from 'webextension-polyfill';
+
+// This was hoisted from userIdService.ts so that it's not being exposed while running the unit tests of the class itself.
```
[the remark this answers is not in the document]
mkazlauskas: LW-9667

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
mkazlauskas: Nice catch. I'll leave the console.error ones unless there's a better way to log in lace?

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
```diff
@@ -154,162 +299,182 @@ export const useWalletManager = (): UseWalletManager => {
...
-  const createWallet = useCallback(
-    async ({ mnemonic, name, password, chainId }: CreateWallet): Promise<CreateWalletData> => {
-      const { keyAgentsByChain, ...wallet } = await executeWithPassword(password, () =>
-        cardanoWalletManager.createCardanoWallet(walletManagerUi, name, mnemonic, getPassword, chainId)
+      const lockValue =
+        lock && keyAgentData.__typename === Wallet.KeyManagement.KeyAgentType.InMemory
+          ? HexBlob.fromBytes(lock)
+          : undefined;
+      // lock behavior changed a little bit, it is only in local storage while actually locked
+      // REVIEW: it is likely that Lace update/migration would unlock a locked wallet. Is that ok?
```
lucas-barros, line 355: You could check if the lock is active and update the lock value?

apps/browser-extension-wallet/src/lib/wallet-api-ui.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+    } catch {
+      // nothing to do here, the error
+      // bubbles up when action rejects
+    }
+  });
+  try {
+    return await action();
+  } finally {
+    subscription.unsubscribe();
```
lucas-barros, line 77: Could we clear the password here?

apps/browser-extension-wallet/src/views/browser-view/features/staking/hooks.ts
```diff
@@ -1,14 +1,26 @@
...
+  const { inMemoryWallet, cardanoWallet } = useWalletStore();
   const { delegationTxBuilder } = useDelegationStore();
   const signAndSubmitTransaction = useCallback(async () => {
-    const signedTx = await delegationTxBuilder.build().sign();
-    await inMemoryWallet.submitTx(signedTx.tx);
-  }, [delegationTxBuilder, inMemoryWallet]);
+    const tx = delegationTxBuilder.build();
+    const { hash } = await tx.inspect();
+    void firstValueFrom(
+      cardanoWallet.signingCoordinator.transactionWitnessRequest$.pipe(
```
lucas-barros, line 15: How does this work? Could we use withSignTxConfirmation here?

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
```diff
@@ -59,79 +60,69 @@ export const DappConfirmData = (): React.ReactElement => {
...
-        };
-        setDappInfo(backgroundDappInfo);
-        setFormattedData(jsonStructureOrHexString);
-      })
-      .catch((error) => {
-        console.error(error);
-      });
-  }, []);
+    const dataFromHex = fromHex(req.blob);
+    // TODO: address is currently not available in sign request,
```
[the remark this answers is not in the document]
mkazlauskas: Yes, address is displayed in sign data confirmation page. It's available now, fixed in 6c1a1b013b51f585a50023fdebe4a3c03a1692ca

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
mkazlauskas: Removed debug logs in 02acf86e4f3ffeb311fdf43518fad755fd055e5d

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
mkazlauskas: Nice catch, I remember wanting to refactor this but then forgot about it. Done in e72edfebcbab3cd3d6daf271d7b7852724cd61c6

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
mkazlauskas: Agree 6e71227b31d86581e55e022eafd4f1ed37c9e3f1

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
mkazlauskas: This one wasn't used, removed in 688280f8b9b6277c399e310042e33a903486d040

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
mkazlauskas: Renamed for better clarity and added a doc comment in 688280f8b9b6277c399e310042e33a903486d040

apps/browser-extension-wallet/src/features/unlock-wallet/components/UnlockWalletContainer.tsx
mkazlauskas: Agree, removed in d31c1fd508d2ad4464bf660bf0e25e84e865f616 and added a note in ticket description that it can be found in git before new wallet manager

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
```diff
@@ -59,79 +60,69 @@ export const DappConfirmData = (): React.ReactElement => {
...
+  useEffect(() => {
+    const subscription = signingCoordinator.signDataRequest$.subscribe(async (r) => {
+      setDappInfo(await senderToDappInfo(r.signContext.sender));
+      setSignDataRequest(r);
+    });
+
+    const api = exposeApi<Pick<UserPromptService, 'allowSignTx'>>(
       {
         api$: of({
-          async allowSignData(): Promise<boolean> {
```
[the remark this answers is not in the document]
mkazlauskas: It is, nice catch. Fixed this in the rename commit here 688280f8b9b6277c399e310042e33a903486d040

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "esModuleInterop": true,
     "declaration": true,
     "declarationMap": true,
-    "noEmitOnError": true,
+    "noEmitOnError": false,
```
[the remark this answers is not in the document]
mkazlauskas: Sounds right, removed this change in 24c957686f7e3c10f2e144d37c67d48f4a7ec21c

packages/cardano/src/wallet/lib/cardano-wallet.ts
```diff
@@ -269,63 +123,35 @@ export const validateWalletPassword = async (
...
-    getPassword
-  );
   // To verify password
-  const originalChallenge = await Challenge.sign(originalKeyAgent);
   const validatingKeyAgent = await KeyManagement.InMemoryKeyAgent.fromBip39MnemonicWords(
     {
       mnemonicWords,
-      chainId: originalKeyAgent.chainId,
-      getPassphrase: getPassword
+      chainId: Cardano.ChainIds.Mainnet, // does not matter
```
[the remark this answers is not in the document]
mkazlauskas: This function compares extended account public key: the one we have with the one we derive from mnemonic. It is the same for all networks, but we're deriving it by creating a key agent which has `chainId` as a required parameter. Ideally we would have a util that just derives the xpub without the redundant parameters, but working with what we have I think it's better if it we set it to a valid value instead of omitting and explicit type cast. Do you think otherwise?

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
mkazlauskas: Done in 90c33fa05136778db3617270a6bec27754b58d1e
Let's make sure it's tested in QA

apps/browser-extension-wallet/src/lib/wallet-api-ui.ts
mkazlauskas: Good idea, done in 741b3994c6f8008a80823da01089764f68683bd8
Although it would be better if it came as byte array already as that we can actually overwrite in memory

apps/browser-extension-wallet/src/views/browser-view/features/staking/hooks.ts
mkazlauskas: Nice catch, this is just some code duplication I left. No need to actually check tx hash here. Done in 38cf7bf51d59d01f5fc6e500dbc64aeaab1d5be7

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
```diff
@@ -40,98 +40,84 @@ const hasJsonStructure = (str: string): boolean => {
...
 
   const [formattedData, setFormattedData] = useState<{
     address: string;
     dataToSign: string;
   }>();
 
-  const cancelTransaction = useCallback(() => {
-    exposeApi<Pick<UserPromptService, 'allowSignData'>>(
+  const cancelTransaction = useCallback(async () => {
+    await req.reject('User rejected to sign');
```
mchappell, line 62: Same comment here, should we be rejecting with `ApiError` and `APIErrorCode.Refused` or does the sdk spit that out?

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -145,64 +134,55 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
...
-        baseChannel: DAPP_CHANNELS.userPrompt,
-        properties: { allowSignTx: RemoteApiPropertyType.MethodReturningPromise }
-      },
-      { logger: console, runtime }
-    );
-    close && setTimeout(() => window.close(), DAPP_TOAST_DURATION);
-  }, []);
+  const cancelTransaction = useCallback(
+    async (close = false) => {
+      await req.reject('User rejected to sign');
```
mchappell, line 139: Same comment, if we reject here, it'll drop into the catch statement of `cip30.ts` won't it, which will return unable to sign, rather than an explicit rejection code.

apps/browser-extension-wallet/src/stores/slices/wallet-info-slice.ts
```diff
@@ -11,23 +10,21 @@ export const walletInfoSlice: SliceCreator<WalletInfoSlice & BlockchainProviderS
...
   hdDiscoveryStatus: null,
   setHdDiscoveryStatus: (hdDiscoveryStatus) => set({ hdDiscoveryStatus }),
   setCardanoWallet: (wallet?: Wallet.CardanoWallet) => set({ inMemoryWallet: wallet?.wallet, cardanoWallet: wallet }),
-  setWalletManagerUi: (walletManagerUi: WalletManagerUi) => set({ walletManagerUi }),
   setCurrentChain: (chain: Wallet.ChainName) => {
     set({ currentChain: Wallet.Cardano.ChainIds[chain], environmentName: chain });
     get().setBlockchainProvider(chain);
   },
-  getKeyAgentType: () => get()?.cardanoWallet?.keyAgent.serializableData.__typename,
+  getWalletType: () => get()?.cardanoWallet?.source.wallet.type,
```
mchappell, line 28: if we also added a variable for this rather than a function, we'd negate the need to perform this check in so many places and just use the type directly

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
```diff
@@ -0,0 +1,146 @@
...
+ * 2. Sign
+ *
+ * This function notifies service worker that UI is ready to sign.
+ */
+export const readyToSign = (): void => {
+  exposeApi<Pick<UserPromptService, 'readyToSignTx'>>(
+    {
+      api$: of({
+        async readyToSignTx(): Promise<boolean> {
+          return Promise.resolve(true);
```
DominikGuzei, line 55: I guess I don't understand how just resolving a boolean value here can be of any use? Ok there is one scenario i can think of: the service worker is polling this `readyToSignTx` endpoint and when the JS in the UI has initialized it returns true?

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
```diff
@@ -1,44 +1,87 @@
...
 
-const walletFactory: WalletFactory = {
-  create: async (
-    props: WalletManagerActivateProps,
-    dependencies: { keyAgent: Wallet.KeyManagement.AsyncKeyAgent; stores: storage.WalletStores }
-  ) => {
+// It is important that this file is not exported from index,
+// because creating wallet repository with store creates an actual pouchdb database
+// which results in some trash files when running the tests (leveldb directory)
+export const walletRepository = new WalletRepository({
```
DominikGuzei, line 33: hm, should we then create the wallet repository instance in the root of the file?

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
mkazlauskas: Did you see 688280f8b9b6277c399e310042e33a903486d040 ? I added a doc comment for `readyToSignTx`.
It's basically this:

> the service worker is polling this readyToSignTx endpoint and when the JS in the UI has initialized it returns true

Except that it doesn't have to do any polling, it just waits for extension messaging connection to establish, then this is called and resolves, notifying the service worker that UI is ready to sign.

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
mkazlauskas: Could you please elaborate on this suggestion? :thinking:

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
DominikGuzei: Ok, after looking at the `exposeApi` + `api$` + `of` in the SDK, is it true that this creates an observable from the `allowSignTx` api method which can then be observed by the remote (service worker)?

mkazlauskas: `api$` can be used to switch the exposed object (like the ObservableWallet that is exposed from service worker). In this case it uses `of` which just emits single value and completes. So this is not utilizing this capability, just exposing a specific api object. This is not new a new setup btw, just renamed the method from `allowSignTx` and changed it's purpose a little bit so that it's doing less - only used for coordination to notify service worker when UI is initialized.

DominikGuzei: thanks for the explanation 👍 yes I have read the doc comment and for me it was still not obvious what's going on, so after your explanation maybe let's change it to:
```
// Exposes an `allowSignTx` observable, which emits a single `true` value to the service worker to indicate that the UI has loaded and is ready to sign
```

what do you think?

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
DominikGuzei: @mkazlauskas your comment suggests that creating the pouchdb database directly in the root of the file complicates the handling of this file, so I wondered if it might be possible to defer the creation with a setup function and call it separately?

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  return sectionTitle[DAPP_VIEWS.CONFIRM_TX];
+};
+
+export const disallowSignTx = (close = false): void => {
+  exposeApi<Pick<UserPromptService, 'allowSignTx'>>(
+    {
+      api$: of({
+        async allowSignTx(): Promise<boolean> {
+          return Promise.reject(new Error('Transaction rejected'));
```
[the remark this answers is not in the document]
mkazlauskas: This was removed

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
```diff
@@ -40,98 +40,84 @@ const hasJsonStructure = (str: string): boolean => {
...
-            },
-            { logger: console, runtime }
-          );
-        })
-        .catch((error) => {
-          throw error;
-        });
-    } catch {
+      if (req.walletType !== WalletType.Ledger && req.walletType !== WalletType.Trezor) {
+        throw new Error('Invalid state: expected hw wallet');
```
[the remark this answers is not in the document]
mkazlauskas: I'm not sure: this is an assertion for state that should never happen. It doesn't mean that user rejected a transaction, but that there's a bug in Lace

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
mkazlauskas: SDK wraps it in `AuthenticationError` (from key-management), which should then be wrapped in `TxSignError` in cip30.ts :thinking:

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
mkazlauskas: It could be deferred by converting this into `getWalletRepository` (as well as any other exports that depends on this one), but I personally prefer this for cleaner code. Ideally this would be created in index script (e.g. background.ts) and passed as dependency everywhere else, but that's a bigger refactor that I don't think we should do in this PR.

.github/shared/build/action.yml
```diff
@@ -8,6 +8,9 @@ inputs:
   LACE_EXTENSION_KEY:
     description: 'Public extended manifest key'
     required: true
+  CARDANO_SERVICES_URL_MAINNET:
```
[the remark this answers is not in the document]
mkazlauskas: Reverted in 7cd2b62b0662443ee8b21117ab76b2076e518f9c

.github/workflows/smoke-tests.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         uses: ./.github/shared/build
         with:
           LACE_EXTENSION_KEY: ${{ secrets.MANIFEST_PUBLIC_KEY }}
+          CARDANO_SERVICES_URL_MAINNET: https://backend.dev-perf-mainnet.eks.lw.iog.io
```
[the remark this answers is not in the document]
mkazlauskas: Reverted in 7cd2b62b0662443ee8b21117ab76b2076e518f9c

apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/HardwareWalletFlow.tsx
```diff
@@ -175,14 +175,7 @@ export const HardwareWalletFlow = ({
...
       );
-      const addressDiscoverySubscriber = wallet.wallet.addresses$.subscribe((addresses) => {
-        if (addresses.length === 0) return;
-        const hdWalletDiscovered = addresses.some((addr) => addr.index > 0);
-        if (hdWalletDiscovered) {
-          analytics.sendEventToPostHog(PostHogAction.OnboardingRestoreHdWallet);
-        }
-        addressDiscoverySubscriber.unsubscribe();
-      });
+      analytics.sendEventToPostHog(PostHogAction.OnboardingRestoreHdWallet);
```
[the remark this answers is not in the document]
mkazlauskas: Nice catch! This was a bug, fixed in c68ccbe97560198f238d6b716e04bbba3e13c91d

Note I hoisted it one level up to WalletSetupWizard, because wallet has to be activated in order for `addresses$` to emit.

packages/staking/src/features/outside-handles-provider/types.ts
```diff
@@ -52,6 +52,7 @@ export interface IBlockchainProvider {
 
 export type OutsideHandlesContextValue = {
   analytics: IAnalyticsTracker;
+  walletManagerExecuteWithPassword: (action: () => Promise<unknown>, password?: string) => Promise<unknown>;
```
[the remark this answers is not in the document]
mkazlauskas: No reason, I deleted it during development and then re-added with unknown as (I believe) return type wasn't used anywhere. Fixed in ad868dccb2736c690645d1a95d1e5598816891df

packages/cardano/src/wallet/lib/hardware-wallet.ts
```diff
@@ -1,14 +1,9 @@
 /* eslint-disable unicorn/no-null */
-import { Cardano } from '@cardano-sdk/core';
-import * as KeyManagement from '../../../../../node_modules/@cardano-sdk/key-management/dist/cjs';
-import { ChainName, DeviceConnection, CreateHardwareWalletArgs, HardwareWallets } from '../types';
-import { CardanoWalletByChain, KeyAgentsByChain } from '@src/wallet';
-import { WalletManagerUi } from '@cardano-sdk/web-extension';
-import * as Crypto from '@cardano-sdk/crypto';
+import * as KeyManagement from '@cardano-sdk/key-management';
+import { DeviceConnection, HardwareWallets } from '../types';
 import * as HardwareLedger from '../../../../../node_modules/@cardano-sdk/hardware-ledger/dist/cjs';
```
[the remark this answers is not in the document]
mkazlauskas: I tried, but unfortunately it's a rabbit hole. I can try to spend a couple more hours on this separately, after this is merged.

apps/browser-extension-wallet/src/components/MainMenu/DropdownMenuOverlay/components/LockWallet.tsx
```diff
@@ -1,17 +1,22 @@
...
 import { PostHogAction } from '@providers/AnalyticsProvider/analyticsTracker';
+import { Wallet } from '@lace/cardano';
 
 export const LockWallet = (): React.ReactElement => {
   const { t } = useTranslation();
   const { lockWallet } = useWalletManager();
-  const { walletLock } = useWalletStore();
+  const { getKeyAgentType } = useWalletStore();
   const analytics = useAnalyticsContext();
+  const isInMemoryWallet = useMemo(
```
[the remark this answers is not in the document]
mkazlauskas: Done in 9f36f17a21f5a090465eb66a616f34d1348a1064

Note: didn't add isInMemoryWallet and isHardwareWallet props for `staking` package state, as it's only used in a couple places.

apps/browser-extension-wallet/src/stores/slices/wallet-info-slice.ts
mkazlauskas: Done in 9f36f17a21f5a090465eb66a616f34d1348a1064

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/utils.ts
mkazlauskas: Updated comment in cd41d9e5706f6455558f2e5bd27df9a179938277

DominikGuzei (approved):
great work 🎉

lucas-barros (approved):

apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/HardwareWalletFlow.tsx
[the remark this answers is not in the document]
mchappell: yeh that was likely my fault when we did the review on the other SDK pr, I spotted it then put it back but I think you'd already done the merge at that point

mchappell (approved):

rhyslbw (approved):

VanessaPC (approved):

packages/cardano/src/wallet/test/mocks/TestKeyAgent.ts
```diff
@@ -23,6 +21,6 @@ export const testKeyAgent = ({
     },
     {
       logger: console,
-      bip32Ed25519: new Crypto.CmlBip32Ed25519(CML)
+      bip32Ed25519: new Crypto.SodiumBip32Ed25519()
```
mchappell, line 24: 🚀

mchappell (commented):
LGTM

apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/components/Form/Form.tsx
```diff
@@ -159,13 +152,6 @@ export const Form = ({
           adaAmount={totalCost.ada}
           fiatAmount={totalCost.fiat}
           tooltipContent={t('send.theAmountYoullBeChargedToProcessYourTransaction')}
-          onTooltipHover={() =>
-            analytics.sendEventToMatomo({
```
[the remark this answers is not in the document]
mkazlauskas: I don't know, but it wasn't, so it should probably be out of scope for this PR

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/StakePoolDetails/StakePoolDetailsDrawer.tsx
```diff
@@ -39,22 +37,19 @@ export const StakePoolDetailsDrawer = ({
   } = useStakePoolDetails();
   const { t } = useTranslation();
   const { setIsRestaking } = useSubmitingState();
-  const { getKeyAgentType } = useWalletStore();
+  const { isInMemoryWallet } = useWalletStore();
   const { password, removePassword } = usePassword();
-  const isInMemory = useMemo(() => getKeyAgentType() === Wallet.KeyManagement.KeyAgentType.InMemory, [getKeyAgentType]);
   // const isSuccessSection = useMemo(
```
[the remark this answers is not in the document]
mkazlauskas: https://github.com/input-output-hk/lace/pull/816/commits/cc4fa8e6d7b9967bc1a44ec8dbd6f733c09de4d1

ljagiela (approved):

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
```diff
@@ -40,105 +40,90 @@ const hasJsonStructure = (str: string): boolean => {
...
   }>();
 
-  const cancelTransaction = useCallback(() => {
-    exposeApi<Pick<UserPromptService, 'allowSignData'>>(
+  const cancelTransaction = useCallback(async () => {
+    await req.reject('User rejected to sign');
+    setTimeout(() => window.close(), DAPP_TOAST_DURATION);
+  }, [req]);
+
+  window.addEventListener('beforeunload', cancelTransaction);
```
VanessaPC, line 64: generally adding add event listeners but not removing them is discouraged, I would suggest: 
```
  useEffect(() => {
    window.addEventListener('beforeunload', cancelTransaction);

    return () => {
      window.removeEventListener('beforeunload', cancelTransaction);
    };
  }, [cancelTransaction]);
  ```

File: apps/browser-extension-wallet/src/features/dapp/components/collateral/CreateCollateral.tsx [hunk 32]
```diff
@@ -65,39 +62,36 @@ export const CreateCollateral = ({
...
-    if (isInMemory) {
-      try {
-        await validatePassword(password);
-        await executeWithPassword(password, submitTx, true);
-      } catch {
+
+    try {
+      await withSignTxConfirmation(submitTx, password);
+    } catch (error) {
+      if (error instanceof Wallet.KeyManagement.errors.AuthenticationError) {
```
VanessaPC, line 81: nice

VanessaPC (approved):
I left a small comment, it looks good to me!

apps/browser-extension-wallet/src/features/dapp/components/ConfirmData.tsx
mkazlauskas: Agree, but I tried to avoid making changes unless something got in the way of wallet manager refactor. This was already like that: https://github.com/input-output-hk/lace/pull/816/files#diff-36b0722aea432cfca18257c06db391c57e9d8a0d8fbd9dd9281237b45fe91936L78 Let me know if you want me to resolve this issue in this PR or if it's ok to keep as-is.

VanessaPC: I thought it was a change from the PR, let's keep it as is.

PR #826, opened 2024-01-09
test(extension): [lw-8530] handle reload after network switching

bslabiak:
# Checklist

- [x] JIRA - \<[link](https://input-output.atlassian.net/browse/LW-8530)>

### DIFF
--- a/packages/e2e-tests/src/features/NavigationTopPopup.feature
+++ b/packages/e2e-tests/src/features/NavigationTopPopup.feature
@@ -82,7 +82,7 @@ Feature: Top Navigation - Popup view
     And I click on the network option
     When I click on "Preview" radio button
     Then Lace is loaded properly
-    Then I click the menu button
+    When I click the menu button
     And I click on the settings option
     Then I see current network: "Preview" name in network setting
     And I see network id: "Preview"
--- a/packages/e2e-tests/src/features/NavigationTopPopup.feature
+++ b/packages/e2e-tests/src/features/NavigationTopPopup.feature
@@ -67,23 +67,22 @@ Feature: Top Navigation - Popup view
     When I click on then network sub-menu back button
     Then the dropdown menu is visible
 
-  @LW-6078 @Testnet @Mainnet @Pending
-  # Bug LW-8530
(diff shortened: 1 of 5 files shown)


packages/e2e-tests/src/features/NavigationTopPopup.feature
```diff
@@ -67,23 +67,22 @@ Feature: Top Navigation - Popup view
...
-  # Bug LW-8530
+  @LW-6079 @Testnet @Mainnet
   Scenario: Popup View - Network switched after choosing Preview network
     Given I click the menu button
     And I see current network in user menu
     And I click on the network option
     When I click on "Preview" radio button
-    And I click on then network sub-menu back button
+    Then Lace is loaded properly
+    Then I click the menu button
```
wklos-iohk, line 85: How about replacing `Then` with `When` keyword here?

packages/e2e-tests/src/steps/commonSteps.ts
```diff
@@ -41,7 +41,12 @@ import consoleAssert from '../assert/consoleAssert';
 
 Given(/^Lace is ready for test$/, async () => {
   await settingsExtendedPageObject.waitUntilSyncingModalDisappears();
+  await tokensPageObject.waitUntilCardanoTokenLoaded();
   await settingsExtendedPageObject.closeWalletSyncedToast();
+});
+
+Then(/^Lace is loaded properly$/, async () => {
```
wklos-iohk, line 48: This step looks pretty similar to the one above `Lace is ready for test`, do we need them both?

wklos-iohk (changes requested):

packages/e2e-tests/src/steps/commonSteps.ts
bslabiak: Difference is in line `await settingsExtendedPageObject.closeWalletSyncedToast();` which is not needed. 
Name of the step `Lace is ready for test` is not suitable for this test scenario so added new one.

wklos-iohk (approved):

PR #830, opened 2024-01-10
[LW-9407] Quorum step storybook setup set

VanessaPC:
# Checklist

- [x] JIRA - LW-9407
- [ ] Proper tests implemented
- [x] Screenshots added.

---

## Proposed solution
- Add the design system component for radio button 
- Add the design system component for dropdown selectors 
- Add the view for the Quorum set up step

## Testing
- Run the storybook package in `packages/UI` and check Radio button and Select group. 
- Run storybook in `packages/core` and check the UI for the quorum step

## Screenshots
- Select component: 
Screenshot 2024-01-18 at 20 16 01
Screenshot 2024-01-18 at 20 16 10

- Radio button component:
Screenshot 2024-01-11 at 11 28 58
Screenshot 2024-01-11 at 09 16 12
Screenshot 2024-01-18 at 19 45 50

- Quorum screen:
Screenshot 2024-01-10 at 11 15 50
Screenshot 2024-01-11 at 11 39 18
Screenshot 2024-01-11 at 11 39 38
Screenshot 2024-01-11 at 11 39 30
Screenshot 2024-01-11 at 11 39 23

PR #829, opened 2024-01-09
feat(core): [LW-9409] create Find Your Shared Wallet component

renanvalentin:
# Checklist

- [x] JIRA tickets: https://input-output.atlassian.net/browse/LW-9409 , https://input-output.atlassian.net/browse/LW-9521
- [x] Screenshots added.

---

## File upload component

Added a new UI component for file upload:

```tsx
 <FileUpload
          label={[
            { text: 'Drag & drop or', highlight: false },
            { text: 'choose file', highlight: true },
            { text: 'to upload', highlight: false },
          ]}
          removeButtonLabel='Remove'
          supportedFormats='Supported formats: JSON'
          accept="application/json"
          onChange={handleFileChange}
          files={file ? [file.name] : undefined}
        />
```

## Screenshots

capture-4b66205b

capture-fa81fbd4

capture-b8ae2f97

PR #814, opened 2024-01-04
feat(extension): add shared wallet LW-9403 & LW-9413

DominikGuzei:
# Checklist

- [x] JIRA tickets: https://input-output.atlassian.net/browse/LW-9413 https://input-output.atlassian.net/browse/LW-9403
- [x] Screenshots added.

---

## Proposed solution

This PR implements the foundation for the shared wallets flow.

## Screenshots

Share wallet account type in the wallet menu:

image

Shared wallet setup step components in Storybook in the `core` package:

Bildschirmfoto 2024-01-04 um 13 42 37

Bildschirmfoto 2024-01-04 um 13 42 28

Recording of the UX in Lace:

https://github.com/input-output-hk/lace/assets/172414/127f3a07-994a-4b91-a2f5-b47545d3d1ea

## How to test

Set `USE_SHARED_WALLET=true` in your `.env` and then open the modal via the "add shared wallet" link in the account menu

PR #835, opened 2024-01-12
chore: LW-9555 create icon packages

renanvalentin:
JIRA - https://input-output.atlassian.net/browse/LW-9555

## Problem

While building `packages/core`, these branches crashed due to a mysterious rollup error:

```sh
src/index.ts → dist/index.js, dist/index.esm.js...
[!] Error: 'ReactComponent' is not exported by ../ui/dist/assets/icons/book-gradient.component.svg, imported by src/ui/components/SharedWallet/AllSet/AllSet.tsx
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
src/ui/components/SharedWallet/AllSet/AllSet.tsx (3:9)
```

https://github.com/input-output-hk/lace/pull/814/
https://github.com/input-output-hk/lace/pull/830/

Whereas this branch was working fine:
https://github.com/input-output-hk/lace/pull/829/

The storybook, which uses `webpack` is working in all listed branches, so it might be some tooling issue.

## Proposed solution

Instead of using the raw file, let other packages consume the component version of the icons, removing the tooling step and enabling seamless integration.

## Usage

Import the icon component using the path `@lace/icons/dist/*`. e.g:

```tsx
import { ReactComponent as LightBulbGradient } from '@lace/icons/dist/LightBulbGradientComponent';

render(
<LightBulbGradient />
)
```

## Notes

Use the `Commits` tab to navigate the changes.


renanvalentin:
Looks like there's no regression. See visual regression testing:
https://www.chromatic.com/build?appId=659e90c5edb661a3f34725d8&number=21
https://www.chromatic.com/build?appId=6422dc516f0543c5fd174d4c&number=584

VanessaPC (approved):
Looks great to me !

packages/icons/README.md
```diff
@@ -0,0 +1,6 @@
+# Icons
+
+## How to
+
+- Add the icon to `raw` folder using this naming convention: `[name].component.svg`
```
DominikGuzei, line 5: maybe lets change the convention to `[name].icon.svg` so that the generated names include `Icon` at the end instead of `Component`?

renanvalentin: HM. hasn't done this to avoid significant changes.

DominikGuzei: Let's do this in a later iteration 👍

DominikGuzei (approved):

packages/core/src/ui/components/SharedWallet/AllSet/AllSet.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { sx, Box, Flex, Text, Button, EducationalCard } from '@lace/ui';
-import { ReactComponent as BookGradient } from '@lace/ui/dist/assets/icons/book-gradient.component.svg';
-import { ReactComponent as LightBulbGradient } from '@lace/ui/dist/assets/icons/light-bulb-gradient.component.svg';
+import { ReactComponent as BookGradient } from '@lace/icons/dist/BookGradientComponent';
```
szymonmaslowski, line 3: Could you please make the icons public by making them available to be accessed with
`import {} from '@lace/icons'` or `import BookGradient from '@lace/icons/BookGradient'`

This will require two things:
1. generating an index file exporting all the icons
2. setting up the `exports` property on the package json of the `icons` package.

packages/icons/src/ArrowLeftComponent.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
```
szymonmaslowski, line 1: Can we avoid storing duplicated code? The src gets generated so I guess we don't need to keep them in the repo.

szymonmaslowski (commented):
Greate initiative. I left a few suggescrions to improve the dev exp slightly.

packages/core/src/ui/components/SharedWallet/AllSet/AllSet.tsx
renanvalentin: The `index` file is omitted to facilitate tree shaking and reduce the bundle size. 

Regarding `import BookGradient from '@lace/icons/BookGradient'`, it can be considered as an improvement for future tasks, as it would require additional effort to ensure that rollup/webpack is able to import it without the 'dist' folder. I faced some issues last week when trying to implement this, as it involved upgrading TypeScript and changing module resolution.

Given the urgency of this PR, most of the feedback can be addressed in separate tasks. The main focus was to extract the icons from `packages/ui` with as minimal changes as possible.

renanvalentin:
staking looks good! 
Screenshot 2024-01-17 at 11 49 23

pczeglik-iohk (approved):

renanvalentin:
Forced pushed due to an unverified commit that was preventing auto-merge

PR #883, opened 2024-02-08
[LW-9145] TxSummary refactor

VanessaPC:
# Checklist

- [x] JIRA - LW-9145 LW-9710
- [x] Screenshots added.

---

## Proposed solution
- Refactor the transaction summary for dapps 
- add to and from sections 
- txType and origin 

## Testing
- Run the app 
- Go to wingriders and try to get some wingrider token from ada, there shouldn't be tokens 
- Do the reverse transactions, you should see the token 
- Go to handle.me , mint a token, you should see the transaction 

**Tx notes:** 
- A transaction doesn't have to have to and from addresses. Sometimes it may have one sections with various, and sometimes both. But not always. 
- Transactions from swapping places such as Wingrider or SundaeSwap work with a batcher which means that you might see an amount in their interface, and a slightly different amount of ADA or tADA in the dapp, because you will see the amount of coins that don't get sent to a batcher to process the following transaction to actually create the swap. 

## Screenshots

Screenshot 2024-02-21 at 08 28 02
Screenshot 2024-02-21 at 08 29 17
Screenshot 2024-02-21 at 08 33 55

// This image is here only to show the display of to address / from address sections, the styles apply from the previous snapshots. 
Screenshot 2024-02-14 at 08 35 05


DominikGuzei:
@VanessaPC it seems like you need to update this PR to the latest `main` as the changed file count jumped from around 50 to over 200 and i can see many unrelated changes on Github 🤔

packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
```diff
@@ -1,91 +1,177 @@
...
+};
+
+type TransactionType = keyof typeof TransactionTypes;
+
+const tokenName = (assetWithAmount: AssetInfoWithAmount) =>
+  assetWithAmount.assetInfo.nftMetadata !== null
+    ? assetWithAmount.assetInfo.nftMetadata.name
+    : assetWithAmount.assetInfo.tokenMetadata.name;
+
+const hash = (assetWithAmount: AssetInfoWithAmount) =>
```
pczeglik-iohk, line 75: This is only suggestion from my side. Maybe we can rename `hash` and `tokenName` to `getHash` and `getTokenName` so we are consistent with `getStringFromLovelace` and `getTxType`.

What do you think?

packages/core/src/ui/components/DappInfo/DappInfo.module.scss
```diff
@@ -49,4 +49,4 @@
       word-break: break-all;
     }
   }
-}
+}
```
pczeglik-iohk, line 52: Can you revert this change please? :)

packages/core/src/ui/components/ActivityDetail/TransactionFooterDetails.tsx
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable no-magic-numbers */
```
pczeglik-iohk, line 1: where is this magic number?

packages/common/src/ui/lib/add-ellipsis.ts
```diff
@@ -1,6 +1,13 @@
 /* eslint-disable no-magic-numbers */
 const addedLength = 4;
 
+export const truncate = (text: string, partOnelength: number, partTwoLength: number): string => {
```
pczeglik-iohk, line 4: `partOneLength`

apps/browser-extension-wallet/src/views/browser-view/features/staking/components/StakePoolDetails/StakePoolDetailsDrawer.tsx
```diff
@@ -39,7 +39,6 @@ export const StakePoolDetailsDrawer = ({
   const { setIsRestaking } = useSubmitingState();
   const { isInMemoryWallet } = useWalletStore();
   const { password, removePassword } = usePassword();
-
```
pczeglik-iohk, line 42: Can you revert this change, please :)

apps/browser-extension-wallet/src/utils/combined-input-resolvers.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { combineInputResolvers, createBackendInputResolver, createInputResolver } from '@cardano-sdk/wallet';
+import { Cardano } from '@cardano-sdk/core';
+
+interface CombinedInputResolverArgs {
+  utxo: Wallet.ObservableWallet['utxo'];
+  chainHistoryProvider?: any;
+}
+
+export const combinedInputResolver = ({
```
pczeglik-iohk, line 11: `combine`? if you like this suggestion, than we need to rename the file as well :)

apps/browser-extension-wallet/src/hooks/useChainHistoryProvider.ts
```diff
@@ -0,0 +1,15 @@
+import { getBaseUrlForChain } from '@src/utils/chain';
+import { useMemo } from 'react';
+import { chainHistoryHttpProvider } from '@cardano-sdk/cardano-services-client';
+import { logger } from '@lib/wallet-api-ui';
+
+type useChainHistoryProviderArgs = {
```
pczeglik-iohk, line 6: Should type start from the upper case?

As this type is local, maybe it's ok to have `ChainHistoryProviderArgs`?

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -86,52 +47,37 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
...
-    for (const asset of assetMaps) {
-      if (asset) {
-        for (const id of asset.keys()) {
-          !uniqueAssetIds.has(id) && uniqueAssetIds.add(id);
-        }
-      }
-    }
-    return [...uniqueAssetIds.values()];
-  }, [req]);
+  const newChainHistoryProvider = useChainHistoryProvider({ chainName });
```
pczeglik-iohk, line 71: Why new chain history provider?

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -295,14 +204,16 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
   };
 
   return (
-    <Layout pageClassname={styles.spaceBetween} title={t(sectionTitle[DAPP_VIEWS.CONFIRM_TX])}>
-      {req && txSummary ? (
+    <div data-testid="layout-title" className={styles.transactionContainer}>
```
pczeglik-iohk, line 207: What do you think about using `FlexComponent` from ui toolkit, instead of div with custom defined styles?

pczeglik-iohk: https://github.com/input-output-hk/lace/blob/main/packages/ui/src/design-system/flex/flex.component.tsx

packages/ui/src/design-tokens/theme/hooks/use-theme-variant.tsx
```diff
@@ -0,0 +1,11 @@
+import { useContext } from 'react';
+
+import { ThemeContext } from '../theme.context';
+
+import type { ThemeColorScheme } from '../theme.context';
+
+export const useThemeVariant = (): { theme: ThemeColorScheme } => {
+  const themeContext = useContext(ThemeContext);
```
pczeglik-iohk, line 8: should we throw if no `themeContext`? like for example in `useAddressBookContext`

packages/ui/src/design-system/grid/grid.component.tsx
```diff
@@ -9,13 +9,15 @@ export type Props = PropsWithChildren<{
...
   rows?: GridVariants['rows'];
   gutters?: GridVariants['gutters'];
+  alignItems?: GridVariants['alignItems'];
 }>;
 
 export const Grid = ({
   columns = '$none',
   children,
   rows = '$none',
+  alignItems,
```
pczeglik-iohk, line 19: does it make sense to provide default value for `alignItems`?

pczeglik-iohk (commented):
I reviewed the code and posted a few comments.

Nothing crucial, but there are a few improvements which can lead to increased consistency and maintenance of the code.

Let me know what do you think :)

and thank you @VanessaPC for great work.

packages/core/src/ui/components/DappAddressSections/DappAddressSections.tsx
```diff
@@ -0,0 +1,147 @@
...
+      metadataHash={truncate(token.assetInfo.assetId, charBeforeEllMetadata, charAfterEllMetadata)}
+    />
+  ));
+
+const { Title, Text } = Typography;
+
+const charBeforeEllipsisName = 8;
+const charAfterEllipsisName = 8;
+
+export const DappAddressSections = ({
```
DominikGuzei, line 63: is this really only `DApp` related or could this address representation potentially be re-used in other scenarios?

VanessaPC: I don't think this part could be reused, the UI component to display assets could. I think that the TX history does something different, or does nothing at the moment. If we re-do that part and need to use this, then we can think about abstracting this further.

apps/browser-extension-wallet/src/hooks/useChainHistoryProvider.ts
DominikGuzei: or maybe even inlining the type into the function definition in this case?

apps/browser-extension-wallet/src/utils/combined-input-resolvers.ts
DominikGuzei: @pczeglik-iohk there is already a `combineInputResolvers` which this function wraps in order to provide a fixed, already combined set of resolvers, so i think the past tense makes sense since you cannot "combine" arbitrary input resolvers with it.

Maybe instead of the generic `combined` part we could say a bit more about the combination in the name? (i don't have any good suggestion atm)

packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
DominikGuzei: and maybe make it more descriptive like `getAssetHash`?

packages/ui/src/design-system/grid/grid.css.ts
```diff
@@ -83,6 +83,11 @@ export const grid = recipe({
       $20: sx({ gap: '$20' }),
       $32: sx({ gap: '$32' }),
     },
+    alignItems: {
+      $center: {
+        alignItems: 'center',
```
DominikGuzei, line 88: does this mean we can only center the grid for now?

apps/browser-extension-wallet/src/utils/combined-input-resolvers.ts
pczeglik-iohk: yes there is, but this one created by @VanessaPC does not contain `s` on the end :)

packages/icons/src/AdaComponent.tsx
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import * as React from 'react';
+import type { SVGProps } from 'react';
+const SvgAdacomponent = (props: SVGProps<SVGSVGElement>) => (
```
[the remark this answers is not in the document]
VanessaPC: is the convention used in this folder, `SvgLaceLogoComponent`, `SvgUploadComponent`...

packages/ui/src/design-system/grid/grid.css.ts
VanessaPC: I added the basic properties now.

packages/ui/src/design-system/grid/grid.component.tsx
VanessaPC: I wouldn't because the Grid is used in other places without using this property, and this property is also not always used in a grid, if I set a default now, I will be causing issues with the places where its used.

packages/ui/src/design-tokens/theme/hooks/use-theme-variant.tsx
VanessaPC: Here isn't needed because we have a default set.

VanessaPC:
@DominikGuzei @pczeglik-iohk this is ready for re-review.

packages/icons/src/AdaComponent.tsx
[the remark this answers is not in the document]
pczeglik-iohk: sorry, I wasn't clear enough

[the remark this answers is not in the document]
VanessaPC: oh!... pushed now.

pczeglik-iohk (approved):
Code looks great. Thank you @VanessaPC

apps/browser-extension-wallet/src/hooks/useChainHistoryProvider.ts
```diff
@@ -0,0 +1,15 @@
+import { getBaseUrlForChain } from '@src/utils/chain';
+import { useMemo } from 'react';
+import { chainHistoryHttpProvider } from '@cardano-sdk/cardano-services-client';
+import { logger } from '@lib/wallet-api-ui';
+
+type UseChainHistoryProviderArgs = {
```
DominikGuzei, line 6: I can see that the names of the networks are all over the place in our code base, maybe let's move them into an enum somewhere like it's already one for the e2e here https://github.com/input-output-hk/lace/blob/6516722024494584fd3bb81196a345d5895f43db/packages/e2e-tests/src/types/network.ts

DominikGuzei (approved):
Works great now 🎉 I just left one small suggestion that you can optionally look into (not really the scope of this PR)

apps/browser-extension-wallet/src/hooks/useChainHistoryProvider.ts
VanessaPC: Your suggestions is really good. This has come up in conversation already somewhere else, but is there a generic constants place for an enum of this type? and have it been considered a constants file/folder in the app?

DominikGuzei: @VanessaPC I'm not sure, it's definitely something we could discuss with the team 👍

VanessaPC: I have created: https://input-output.atlassian.net/browse/LW-9883 to address the topic in a general setting. I think that doing a cleanup like this shouldn't be too complicated and it'd be better done at once.

bslabiak (approved):

apps/browser-extension-wallet/src/features/dapp/components/confirm-transaction/DappTransactionContainer.tsx
```diff
@@ -104,10 +104,7 @@ export const DappTransactionContainer = withAddressBookContext(
 
     const userAddresses = useMemo(() => walletInfo.addresses.map((v) => v.address), [walletInfo.addresses]);
     const userRewardAccounts = useObservable(inMemoryWallet.delegation.rewardAccounts$);
-    const rewardAccountsAddresses = useMemo(
-      () => userRewardAccounts && userRewardAccounts.map((key) => key.address),
-      [userRewardAccounts]
-    );
+    const rewardAccountsAddresses = useMemo(() => userRewardAccounts?.map((key) => key.address), [userRewardAccounts]);
```
pczeglik-iohk, line 107: thank you @VanessaPC

PR #902, opened 2024-02-20
feat: move pin extension notification to main screen

greatertomi:
# Checklist

- [x] JIRA - LW-9747
- [x] Screenshots added.

---

## Proposed solution

move the pin extension notification from the final onboarding step to the main screen

## Testing

## Screenshots

Screenshot 2024-02-20 at 13 52 47

### DIFF
--- a/apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
@@ -32,6 +32,7 @@ export const Layout = ({ children, drawerUIDefaultContent, isFullWidth }: Layout
   const backgroundServices = useBackgroundServiceAPIContext();
 
   const [showPinExtension, { updateLocalStorage: setShowPinExtension }] = useLocalStorage('showPinExtension', true);
+  const [showDappBetaModal] = useLocalStorage('showDappBetaModal', true);
 
   useEffect(() => {
     const openDrawer = async () => {
@@ -56,11 +57,19 @@ export const Layout = ({ children, drawerUIDefaultContent, isFullWidth }: Layout
   }, [backgroundServices, setTheme]);
 
   useEffect(() => {
-    const timer = setTimeout(() => {
-      setShowPinExtension(false);
-    }, PIN_EXTENSION_TIMEOUT);
-    return () => clearTimeout(timer);
-  }, [setShowPinExtension]);
+    let timer: NodeJS.Timeout;
+    if (!showDappBetaModal) {
+      timer = setTimeout(() => {
+        setShowPinExtension(false);
+      }, PIN_EXTENSION_TIMEOUT);
+    }
+
+    return () => {
+      if (timer) {
+        clearTimeout(timer);
+      }
+    };
+  }, [setShowPinExtension, showDappBetaModal]);
 
   const debouncedToast = useMemo(() => debounce(toast.notify, toastThrottle), []);
   const showNetworkError = useCallback(
--- a/apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
@@ -57,18 +57,13 @@ export const Layout = ({ children, drawerUIDefaultContent, isFullWidth }: Layout
   }, [backgroundServices, setTheme]);
 
   useEffect(() => {
-    let timer: NodeJS.Timeout;
-    if (!showDappBetaModal) {
-      timer = setTimeout(() => {
-        setShowPinExtension(false);
-      }, PIN_EXTENSION_TIMEOUT);
-    }
+    if (showDappBetaModal) return;
+    const timer = window.setTimeout(() => {
+      setShowPinExtension(false);
+    }, PIN_EXTENSION_TIMEOUT);
 
-    return () => {
-      if (timer) {
-        clearTimeout(timer);
-      }
-    };
+    // eslint-disable-next-line consistent-return
+    return () => window.clearTimeout(timer);
   }, [setShowPinExtension, showDappBetaModal]);
 
   const debouncedToast = useMemo(() => debounce(toast.notify, toastThrottle), []);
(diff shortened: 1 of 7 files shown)


szymonmaslowski (approved):
💪

szymonmaslowski:
I think we should start counting 5 seconds after the user closes the dapp connector popup. Now the pin extension element is barely visible and probably will disappear before the user finishes reading the modal content. It would be good to have a generic mechanism checking if the screen is not covered otherwise there might be a different modal in the future covering this pin extension thing.
image

The second thing I noticed is that this element covers entirely the account dropdown so the user has no possibility to interact with it. Maybe it could have a close button, or an indicator for when it is going to be hidden, similar to our toasts elements (see below).
image

cc: @vhulchenko-iohk

apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/PinExtension.module.scss
```diff
@@ -25,12 +25,12 @@
 
     h5 {
       @include text-bodySmall-bold;
-      color: var(--text-color-primary);
+      color: var(--text-color-white, #ffffff) !important;
```
szymonmaslowski, line 28: My heart bleeds seeing that 😭 But we cannot do anything about it because of [this piece of sh**](https://github.com/input-output-hk/lace/blob/0ccc47f6e55c9692d5509cf9d5f8f41cadee61be/apps/browser-extension-wallet/src/views/browser-view/features/settings/components/SettingsLayout.module.scss#L330) which impacts whole lace 😭 😭 😭

shawnbusuttil (approved):

apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
```diff
@@ -21,13 +23,16 @@ interface LayoutProps {
 
 const toastThrottle = 500;
 const isFlexible = process.env.USE_DESKTOP_LAYOUT === 'true';
+const PIN_EXTENSION_TIMEOUT = 5000;
```
danielmain, line 26: can we move `PIN_EXTENSION_TIMEOUT` it same as `USE_DESKTOP_LAYOUT` in the envs?

apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/PinExtension.module.scss
danielmain: Any possibility to address it on #lace-dev ?

danielmain (approved):
I left a question about the constant

apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
greatertomi: I don't think it's the same as `USE_DESKTOP_LAYOUT`, besides the `PIN_EXTENSION_TIMEOUT` is only used in this component, I don't see the need to move it to an env file.

greatertomi:
> The second thing I noticed is that this element covers entirely the account dropdown so the user has no possibility to interact with it. Maybe it could have a close button, or an indicator for when it is going to be hidden, similar to our toasts elements (see below).

Is this important since the notification will close in 5s?

apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/PinExtension.module.scss
szymonmaslowski: I asked Michael for a place where I could report that and waiting for the suggestion. Do you think it is worth raising it on our channel?

greatertomi: I don't know, but changing that could have app-wide effects. 😅

vhulchenko-iohk:
> I think we should start counting 5 seconds after the user closes the dapp connector popup. Now the pin extension element is barely visible and probably will disappear before the user finishes reading the modal content. It would be good to have a generic mechanism checking if the screen is not covered otherwise there might be a different modal in the future covering this pin extension thing. image
> 
> The second thing I noticed is that this element covers entirely the account dropdown so the user has no possibility to interact with it. Maybe it could have a close button, or an indicator for when it is going to be hidden, similar to our toasts elements (see below). image
> 
> cc: @vhulchenko-iohk

Hey @szymonmaslowski, great points. Chatting with the UI/UX designer about them. If OK and agreed, I will have a separate Jira ticket to add these suggested changes. OK?

cc @greatertomi

apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
```diff
@@ -56,11 +57,19 @@ export const Layout = ({ children, drawerUIDefaultContent, isFullWidth }: Layout
...
+      timer = setTimeout(() => {
+        setShowPinExtension(false);
+      }, PIN_EXTENSION_TIMEOUT);
+    }
+
+    return () => {
+      if (timer) {
+        clearTimeout(timer);
+      }
+    };
```
szymonmaslowski, line 71: We should not use `NodeJS.Timeout` type, as it is not a nodejs runtime. Please use `number` and `window.setTimeout` or consider this change:
```suggestion
    if (showDappBetaModal) return undefined;
    const timer = window.setTimeout(() => {
      setShowPinExtension(false);
    }, PIN_EXTENSION_TIMEOUT);
    return () => window.clearTimeout(timer);
```

szymonmaslowski (changes requested):
One more small request 😉

apps/browser-extension-wallet/src/views/browser-view/components/Layout/Layout.tsx
greatertomi: Fixed on a80bda7. Took this approach to avoid the "eslint: consistent-return" 😅

greatertomi: But then, we would have to fix "ESLint: Do not use useless `undefined`.(unicorn/no-useless-undefined)", and `eslint --fix` automatically removes the undefined. 😄

greatertomi: I think disabling `eslint: consistent-return` would work perfectly in this case.

szymonmaslowski: I don't like those `unicorn` rules. I noticed, they complicating the development...

greatertomi: I do really like unicorn rules. They generally make the code cleaner. :)

emiride (approved):
Works as expected. It is good to go.

PR #932, opened 2024-03-01
test(extension): refactor e2e initial staking test

ljagiela:


packages/e2e-tests/src/fixture/walletRepositoryBuilder.ts
```diff
@@ -0,0 +1,56 @@
...
+      const wallets = await window.firstValueFrom(window.walletRepository.wallets$);
+      for (const wallet of wallets) {
+        await window.walletRepository.removeWallet(wallet.walletId);
+      }
+      return JSON.stringify(wallets);
+    })()
+  `);
+  Logger.log(`Removed wallets: ${removedWallets}`);
+};
+export const getWalletsFromRepository = async (): Promise<number> =>
```
wklos-iohk, line 29: Not sure if I understand it properly, but function name indicates that it returns wallets, but return type says that we would get only one element of Promise<number> which is misleading.

packages/e2e-tests/src/steps/stakingSteps.ts
```diff
@@ -131,21 +131,13 @@ When(/^I click on the "(.*)" column header$/, async (listHeader: string) => {
...
-          poolTicker: String(testContext.load('poolTicker')),
-          poolID: String(testContext.load('poolID'))
-        }
-      : {
-          transactionDescription: 'Delegation\n1 token',
-          status: 'Success',
-          poolID: String(testContext.load('poolID'))
-        };
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
+Then(/^The Tx details are displayed for Staking (with|without) metadata$/, async (_ignored: 'with' | 'without') => {
```
wklos-iohk, line 135: Why do we ignore `with|without` parameter?

wklos-iohk (changes requested):

packages/e2e-tests/src/features/e2e/StakingInitialFundsE2E.feature
```diff
@@ -26,64 +26,22 @@ Feature: Delegating funds to new pool E2E
...
-    Then Initial Delegation success screen is displayed in popup mode
-    When I click "Close" button on staking success drawer
-    And I wait until current stake pool switch to "ADA Ocean"
-    Then I see currently staking component for stake pool: "ADA Ocean" in popup mode
-    When I navigate to Transactions popup page
-    Then I can see transaction 1 with type "Delegation"
-    And I click and open recent transactions details until find transaction with correct poolID
-    Then The Tx details are displayed for Staking with metadata
+    And I close the drawer by clicking close button
+    And I send back all tADA to the "WalletSendingAdaToStakingE2E" wallet
```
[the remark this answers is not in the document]
ljagiela: I don't thin it's worth it. If the test fails on one of the 1~20 steps the hook will fail anyways.

packages/e2e-tests/src/steps/onboardingSteps.ts
```diff
@@ -530,21 +530,7 @@ Given(/^I create new wallet and save wallet information$/, async () => {
   await Modal.cancelButton.waitForDisplayed();
   await Modal.cancelButton.click();
   await settingsExtendedPageObject.switchNetworkAndCloseDrawer('Preprod', 'extended');
```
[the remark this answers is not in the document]
ljagiela: this is legacy code. I don't think it's worth to take it form env, as this is new wallet and it depends on test Logic to which network we want to switch. I think it would make more sense to parametrise this step, but it can be handled separately.

packages/e2e-tests/src/steps/stakingSteps.ts
ljagiela: this parameter should not be required anymore. I left it for the backwards compatibility and to keep sense of some tests. It might be improved in further tickets.

wklos-iohk (approved):

packages/e2e-tests/src/fixture/walletRepositoryInitializer.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const initialiseBasicLocalStorageData = async (
+  walletName: string,
+  chainName: 'Preprod' | 'Preview' | 'Mainnet'
+): Promise<void> => {
+  await localStorageInitializer.initializeAnalyticsAccepted('ACCEPTED');
+  await localStorageInitializer.initializeShowDAppBetaModal(false);
+  await localStorageManager.setItem('wallet', `{"name":"${walletName}"}`);
+  await localStorageManager.setItem('appSettings', `{"chainName":"${chainName}"}`);
+};
```
bslabiak, line 59: This method fits better to `localStorageInitializer` class

packages/e2e-tests/src/features/e2e/StakingInitialFundsE2E.feature
bslabiak, line 47: Do we really need this step for this scenario? It extending execution time, increasing number of transactions/syncing time in future. Feed source wallet 2 times with faucet should be enough

bslabiak (approved):

PR #933, opened 2024-03-04
Fixes for local run

oldGreg5:
More small fixes for local running full e2e regression.

### DIFF
--- a/packages/e2e-tests/src/assert/addressBook/AddressDetailsAssert.ts
+++ b/packages/e2e-tests/src/assert/addressBook/AddressDetailsAssert.ts
@@ -6,7 +6,7 @@ import { Address } from '../../data/Address';
 class AddressDetailsAssert {
   assertSeeAddressDetailsPage = async (shouldSee: boolean, mode: 'extended' | 'popup', expectedAddress: Address) => {
     if (shouldSee) {
-      await AddressDetails.drawerNavigationTitle.waitForDisplayed({ reverse: mode === 'popup' });
+      await AddressDetails.drawerNavigationTitle.waitForStable();
       await AddressDetails.drawerHeaderCloseButton.waitForDisplayed({ reverse: mode === 'popup' });
       await AddressDetails.drawerHeaderBackButton.waitForDisplayed({ reverse: mode === 'extended' });
       if (mode === 'extended') {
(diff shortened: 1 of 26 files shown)


emiride (approved):

packages/e2e-tests/wdio.conf.base.ts
```diff
@@ -57,6 +57,6 @@ export const config: WebdriverIO.Config = {
     tags: extensionUtils.isMainnet() ? '@Mainnet' : '@Testnet',
     tagsInTitle: true,
     timeout: 200_000,
-    retry: 1
+    retry: 2
```
bslabiak, line 60: I think we do not want to increase retry amount? @oldGreg5

oldGreg5: I noticed it helps when running locally but it actually may make our gitHub regression even longer. I'll drop this change for now. Thanks

packages/e2e-tests/src/assert/addressBook/AddressDetailsAssert.ts
```diff
@@ -6,7 +6,7 @@ import { Address } from '../../data/Address';
 class AddressDetailsAssert {
   assertSeeAddressDetailsPage = async (shouldSee: boolean, mode: 'extended' | 'popup', expectedAddress: Address) => {
     if (shouldSee) {
-      await AddressDetails.drawerNavigationTitle.waitForDisplayed({ reverse: mode === 'popup' });
+      await AddressDetails.drawerNavigationTitle.waitForStable();
```
wklos-iohk, line 9: Will it work if such element is not present?

packages/e2e-tests/src/steps/commonSteps.ts
```diff
@@ -106,7 +106,7 @@ When(
 
 // TODO: deprecated step, to be removed when remaining usages are replaced inside StakingPageDelegatedFundsExtended.feature
 Then(/(An|No) "([^"]*)" text is displayed/, async (expectedResult: string, expectedText: string) => {
-  await $(`//*[contains(text(), "${(await t(expectedText)) ?? expectedText}")]`).waitForDisplayed({
+  await $(`/ /*[contains(text(), "${(await t(expectedText)) ?? expectedText}")]`).waitForDisplayed({
```
wklos-iohk, line 109: typo?

packages/e2e-tests/src/assert/addressBook/AddressDetailsAssert.ts
oldGreg5: In this case yes because this title is always visible for popup and extended, I've run full regression on this branch many times locally and it's fine.

PR #936, opened 2024-03-05
fix: resolve onboarding create restore flow analytics

greatertomi:
# Checklist

- [x] JIRA - LW-9761, LW-9892, LW-9903
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


packages/core/src/ui/components/WalletSetupRevamp/WalletSetupMnemonicStepRevamp/WalletSetupMnemonicStepRevamp.tsx
```diff
@@ -70,13 +77,15 @@ export const WalletSetupMnemonicStepRevamp = ({
...
     setMnemonicWordsConfirm(newMnemonic);
+    onPasteFromClipboard();
   };
 
   const handleBack = () => {
     onReset(mnemonicStage);
   };
 
   const handleNext = () => {
+    onStepNext && onStepNext(mnemonicStage);
```
shawnbusuttil, line 88: Is it possible for analytics related code to be moved to onNext instead? Would potentially like to deprecate this prop eventully.

greatertomi: Looked into this. On this [line](https://github.com/input-output-hk/lace/blob/b31d60936b35da6314465a7257beafed31bb185a/packages/core/src/ui/components/WalletSetupRevamp/WalletSetupMnemonicStepRevamp/WalletSetupMnemonicStepRevamp.tsx#L89-L93), `onNext` does not get called during the `writedown` stage, which makes it impossible for the parent component to know when next is clicked, and when the`mnemonicStage` changes from `writedown` to `input`. 

`onStepNext` sends the current mnemonic stage every time next is clicked. If we remove it, we would only be able to track when `next` is clicked in the `writedown` stage.

shawnbusuttil: Can we not do this in the wizard?

```suggestion
    onNext={(stage) => {
    // logic related to stage
    moveForward();
    }}
```

The reason is because onStepNext ideally doesn't exist. Because it is a function related to the 8/16/24 functionality that we just erased - as the name implies. Similar to the `_1`, `_9`, `_16` analytic keywords that you just removed.

greatertomi: `onStepNext` exists because it is used to capture the transition between steps, previously it captured the transition between steps 1, 9, and 16. With the revamped onboarding update and considering that saving your mnemonic and entering it uses the same component, now it captures the transition between `writedown` and `input`.
 
 let's discuss offline

shawnbusuttil: Resolved on slack.

shawnbusuttil (approved):

emiride:
This now works as expected

emiride (approved):

PR #943, opened 2024-03-07
feat(core, extension): add balance style to the dapp bridge [LW-9887]

DominikGuzei:
# Checklist

- [x] https://input-output.atlassian.net/browse/LW-9887
- [x] Screenshots added.

---

## Proposed solution

There has been a small style change in the wallet activity to differentiate operations that are negative or positive so users don’t have any doubt about the balance change in their wallet. This task is about bringing the same styling to the dapp bridge.

## Testing

Try various DApp transactions and check that:
- All positive amounts are rendered in green (without a + sign) - except for collateral which is always displayed in default style
- Items that are not present in the Tx, or with value 0, should not be displayed (collateral, deposit)
- font sizes are unified for all tx summary amounts

## Screenshots

Bildschirmfoto 2024-03-20 um 12 42 26

### DIFF
--- a/packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
+++ b/packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
@@ -1,18 +1,23 @@
 @import '../../styles/theme.scss';
 @import '../../../../../common/src/ui/styles/abstracts/_typography.scss';
 
+.label {
+  color: var(--text-color-primary, #ffffff) !important;
+  font-size: 16px;
+  font-weight: 600;
+}
+
+.value {
+  color: var(--text-color-primary, #ffffff) !important;
+  font-size: 14px;
+  font-weight: 500;
+}
+
 .address {
   display: flex;
   justify-content: space-between;
   padding: size_unit(2.5) 0;
 }
-
-.addressInfo {
-  color: var(--text-color-primary, #ffffff) !important;
-  font-size: medium;
-  font-weight: 500;
-}
-
 
 .summaryContent {
   padding-bottom: size_unit(2.5);
--- a/packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
+++ b/packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
@@ -23,3 +23,7 @@
   justify-content: space-between;
   align-items: end;
 }
+
+.positiveAmount {
+  color: var(--data-green, #2CB67D) !important;
+}
--- a/packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
+++ b/packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
@@ -30,7 +30,7 @@ export interface DappTransactionProps {
   /** tokens send to being sent to or from the user */
   fromAddress: Map<Cardano.PaymentAddress, TokenTransferValue>;
   toAddress: Map<Cardano.PaymentAddress, TokenTransferValue>;
-  collateral?: string;
+  collateral?: bigint;
 }
 
 const isNFT = (asset: AssetInfoWithAmount) => asset.assetInfo.supply === BigInt(1);
@@ -150,9 +150,9 @@ export const DappTransaction = ({
           })}
         </div>
 
-        {collateral && (
+        {collateral !== undefined && collateral !== BigInt(0) && (
           <Collateral
-            collateral={collateral}
+            collateral={Wallet.util.lovelacesToAdaString(collateral.toString())}
             amountTransformer={amountTransformer({
               price: fiatCurrencyPrice,
               code: fiatCurrencyCode
--- a/packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
+++ b/packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
@@ -13,6 +13,7 @@ import { TransactionFee, Collateral } from '@ui/components/ActivityDetail';
 
 import { TransactionType, DappTransactionSummary, TransactionAssets } from '@lace/ui';
 import { DappAddressSections } from '../DappAddressSections/DappAddressSections';
+import cn from 'classnames';
 
 const amountTransformer = (fiat: { price: number; code: string }) => (ada: string) =>
   `${Wallet.util.convertAdaToFiat({ ada, fiat: fiat.price })} ${fiat.code}`;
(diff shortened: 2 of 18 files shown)


lucas-barros (approved):

apps/browser-extension-wallet/src/features/dapp/components/ConfirmTransaction.tsx
```diff
@@ -225,7 +225,7 @@ export const ConfirmTransaction = withAddressBookContext((): React.ReactElement
             dappInfo={dappInfo}
             fromAddress={fromAddressTokens}
             toAddress={toAddressTokens}
-            collateral={Wallet.util.lovelacesToAdaString(txCollateral.toString())}
+            collateral={txCollateral}
```
VanessaPC, line 228: is the `txCollateral` already in ADA format?

packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
```diff
@@ -144,9 +145,9 @@ export const DappTransaction = ({
           ))}
         </div>
 
-        {collateral && (
+        {collateral !== BigInt(0) && (
```
VanessaPC, line 148: it specifies that if an amount is 0 we shouldn't display it, but collateral will always come as something even if it's 0, right? so don't we need a check to hide it when null?

packages/ui/src/design-system/dapp-transaction-summary/dapp-transaction-summary.css.ts
```diff
@@ -5,10 +5,17 @@ export const transactionTypeContainer = style({
 });
 
 export const label = sx({
-  color: '$dapp_transaction_summary_label',
   fontWeight: '$semibold',
 });
 
+export const positiveBalance = sx({
+  color: '$transaction_summary_positive_balance_label_color',
```
VanessaPC, line 12: these two rules, they start with `transaction_summary` and `dapp_transaction...`, do we want to keep the same naming convention? and also, if dapp is not appropriate, you can mention and I can update the parent PR.

packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
```diff
@@ -22,4 +22,8 @@
   display: flex;
   justify-content: space-between;
   align-items: end;
-}
\ No newline at end of file
+}
+
+.positiveAmount {
+  color: var(--data-green, #2CB67D) !important;
```
VanessaPC, line 28: do we really need important? this section is new.. I was hoping maybe not to have to use it..

pczeglik-iohk: Agree. Let's try to remove it or @DominikGuzei to provide a good argumentation why it's needed :)

packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
pczeglik-iohk, line 148: I see collateral is optional. Does this condition check prevents us from displaying <Collateral /> when collateral is  undefined?

packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
DominikGuzei: @VanessaPC @pczeglik-iohk yeah there definitely was a good reason (i think mostly due to the old Lace code), but i can try again to get rid of it 👍

packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
DominikGuzei: good catch, will add the additional check 👍

packages/ui/src/design-system/dapp-transaction-summary/dapp-transaction-summary.css.ts
DominikGuzei: good catch, will add the `dapp_` part 👍

DominikGuzei: @VanessaPC i noticed that the variables used in this component are not following a clear pattern, let's unify them to `dapp_transaction_summary…`  like the name of the component?

DominikGuzei: oh, actually now that im digging deeper i can see that wouldn't be a smart idea - since the `$transaction_summary_label_color` is used by dapp-transaction-summary AND transaction-summary …

packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
DominikGuzei: @VanessaPC @pczeglik-iohk so you wanted to know, here is why it's not possible to change the color without `!important` in this case (trigger warning!):

Bildschirmfoto 2024-03-08 um 12 14 06

Bildschirmfoto 2024-03-08 um 12 16 58

So yes, i might be able to remove that `!important` rule from the SettingsLayout but I can't say what cascading effects it might have. Another question is, why didn't we use `@lace/ui` to build these new components in the [base branch](https://github.com/input-output-hk/lace/pull/883)? (i guess one answer is that it would have been much more work)

packages/core/src/ui/components/DappTransaction/DappTransaction.tsx
DominikGuzei: fixed in 82c5116c3fe64d219d7edd351298901aeabf6056

packages/ui/src/design-system/dapp-transaction-summary/dapp-transaction-summary.css.ts
DominikGuzei: corrected in 9fe2fcbfc9f318f7b90a05a1cfd56703d76603ba

packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
VanessaPC: hmmm I did try to build all base components in lace/ui though, but there were changes along the way that made it faster at the time to move the work forward this way. I will have a look at this case.

packages/ui/src/design-system/dapp-transaction-summary/dapp-transaction-assets.component.tsx
```diff
@@ -37,8 +29,15 @@ export const TransactionAssets = ({
   tokenName,
   testId,
   ...props
-}: Readonly<Props>): JSX.Element => {
+}: Readonly<{
```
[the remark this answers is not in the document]
DominikGuzei: I guess that was mostly style preference as the Props definition wasn't close to the class anymore and it felt redundant to give it a generic name like `Props` if it isn't used anywhere else … but im not married to that change and can revert if you like 👍

packages/core/src/ui/components/DappAddressSections/DappAddressSections.module.scss
DominikGuzei: @VanessaPC yeah i totally get it, i just wanted to point out why i did have to add `!important` in this case - i think we don't need to block this task because of such a minor detail if (as you know) there many more such troubles in the current code base.

VanessaPC: Yea, agreed. No reason to block it here.

VanessaPC (approved):

packages/ui/src/design-system/dapp-transaction-summary/dapp-transaction-assets.component.tsx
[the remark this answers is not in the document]
DominikGuzei: Im reverting this change 👍

bslabiak (approved):

PR #945, opened 2024-03-11
feat: move the register screen after the mnemonic screen for create and restore flow

greatertomi:
# Checklist

- [x] JIRA - LW-10000
- [x] Screenshots added.

---

## Proposed solution

This PR moves the register screen after the mnemonic screen on the  create and restore onboarding flow

## Testing

Follow the normal onboarding flow

## Screenshots

Screenshot 2024-03-11 at 18 22 07

Screenshot 2024-03-11 at 18 22 38

### DIFF
--- a/packages/core/src/ui/components/WalletSetup/WalletSetupNamePasswordStep/WalletSetupNamePasswordStep.tsx
+++ b/packages/core/src/ui/components/WalletSetup/WalletSetupNamePasswordStep/WalletSetupNamePasswordStep.tsx
@@ -15,12 +15,24 @@ import {
 import { WalletNameInput } from './WalletNameInput';
 import { WalletPasswordConfirmationInput } from './WalletPasswordConfirmationInput';
 import { WalletSetupStepLayoutRevamp } from '../../WalletSetupRevamp';
+import { TranslationsFor } from '@ui/utils/types';
 
 export interface WalletSetupNamePasswordStepProps {
   onBack: () => void;
   onNext: (params: WalletSetupNamePasswordSubmitParams) => void;
   initialWalletName?: string;
   onChange?: (state: { name: string; password: string }) => void;
+  translations: TranslationsFor<
+    | 'title'
+    | 'description'
+    | 'nameInputLabel'
+    | 'passwordInputLabel'
+    | 'confirmPasswordInputLabel'
+    | 'noMatchPassword'
+    | 'nameMaxLength'
+    | 'nameRequiredMessage'
+    | 'enterWallet'
+  >;
 }
 
 const INITIAL_WALLET_NAME = 'Wallet 1';
@@ -29,7 +41,8 @@ export const WalletSetupNamePasswordStep = ({
   onBack,
   onNext,
   initialWalletName = INITIAL_WALLET_NAME,
-  onChange
+  onChange,
+  translations
 }: WalletSetupNamePasswordStepProps): React.ReactElement => {
   const { t } = useTranslate();
   const [password, setPassword] = useState('');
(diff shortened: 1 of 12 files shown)


packages/core/src/ui/components/WalletSetup/wallet-steps.ts
```diff
@@ -1,14 +1,10 @@
...
 export const walletSetupWizard: WalletSetupWizard = {
-  [WalletSetupSteps.Register]: {
-    next: WalletSetupSteps.Mnemonic
-  },
   [WalletSetupSteps.Mnemonic]: {
-    prev: WalletSetupSteps.Register,
-    next: WalletSetupSteps.Create
+    next: WalletSetupSteps.Register
   },
-  [WalletSetupSteps.Create]: {
```
shawnbusuttil, line 11: Doesn't create need to stay for the third step to exist?

shawnbusuttil (commented):
Nice job. Left a question.

packages/core/src/ui/components/WalletSetup/wallet-steps.ts
szymonmaslowski: I get this point of view that a simple enum would do, however, IMO since we have this structure already implemented I would not invest in rewriting it to an enum as it also gives a value. I like that it is separated from the view. If I was implementing that from scratch I would go with enum, but since we have it, I would keep it 😉

greatertomi: Same. It would keep the current linked list structure. It's more extensible incase there's a need to add more steps in the future.

shawnbusuttil (approved):

packages/core/src/ui/lib/translations/en.json
```diff
@@ -406,6 +406,7 @@
       "confirmPasswordInputLabel": "Confirm password",
       "nameRequiredMessage": "Wallet name is required",
       "noMatchPassword": "Oops! The passwords don't match.",
+      "enterWallet": "Enter wallet",
```
szymonmaslowski, line 409: I guess those translations are supposed to be placed in the browser translations. Could you move them there? Otherwise, it will be inconsistent to have translations for a single step separate from others.

There is also a `walletSetupStep` key in this file that would be also worth moving.

szymonmaslowski (changes requested):
Goof Job. Left one comment.

packages/core/src/ui/lib/translations/en.json
greatertomi: There is no clear strategy for translations. I remember, there was a plan to move translations into their packages started by this https://github.com/input-output-hk/wallet-world/pull/1407, and that is why there is a translation file inside the core package.

I would move this step's translation into the browser's translation file since it differs from the rest, but there might be a need to move translations within their packages to avoid the unnecessary passing around of translation strings.

greatertomi: Moved the translations from `core` package on 691d532. 

I am doubtful about the importance of this change though as it affects multi-wallet codebase, and the discussion around if core should should be a UI library is something that should be done separately. I favour us leaving the translation file for this component the way it currently is. 🙂

packages/core/src/ui/components/WalletSetup/WalletSetupNamePasswordStep/WalletSetupNamePasswordStep.tsx
```diff
@@ -43,15 +56,11 @@ export const WalletSetupNamePasswordStep = ({
...
+    passHasBeenValidated && password !== passwordConfirmation ? translations.noMatchPassword : '';
 
   const walletNameErrorMessage = useMemo(() => {
-    const validationError = validateNameLength(walletName)
-      ? t('package.core.walletNameAndPasswordSetupStep.nameMaxLength')
-      : '';
-    return walletName ? validationError : t('package.core.walletNameAndPasswordSetupStep.nameRequiredMessage');
+    const validationError = validateNameLength(walletName) ? translations.nameMaxLength : '';
+    return walletName ? validationError : translations.nameRequiredMessage;
   }, [t, walletName]);
```
szymonmaslowski, line 64: Please fix the warning about dependency array

szymonmaslowski (changes requested):

packages/core/src/ui/components/WalletSetup/WalletSetupNamePasswordStep/WalletSetupNamePasswordStep.tsx
```diff
@@ -15,12 +15,24 @@ import {
...
+  translations: TranslationsFor<
+    | 'title'
+    | 'description'
+    | 'nameInputLabel'
+    | 'passwordInputLabel'
+    | 'confirmPasswordInputLabel'
+    | 'noMatchPassword'
+    | 'nameMaxLength'
+    | 'nameRequiredMessage'
+    | 'enterWallet'
```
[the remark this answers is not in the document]
greatertomi: Fixed on 4c3020c

szymonmaslowski (approved):

PR #961, opened 2024-03-18
fix: retry button on transfaction failure should check old firmware error for ledger

AngelCastilloB:
The retry button on the transaction failure doesnt not follow the same validations that the stack confirmation modal

# Checklist

- [X] JIRA - \<link>
- [ ] Screenshots added.

---

## Proposed solution

Make the MULTIDELEGATION_NOT_SUPPORTED validations on the TransactionFail modal

## Screenshots


packages/staking/src/features/Drawer/TransactionFail.tsx
```diff
@@ -57,23 +60,37 @@ export const TransactionFailFooter = ({ popupView }: TransactionFailProps): Reac
...
   const onSubmit = async () => {
     setIsLoading(true);
+
     try {
       await signAndSubmitTransaction();
       setIsLoading(false);
       portfolioMutators.executeCommand({ type: 'DrawerContinue' });
       removePassword();
-    } catch (error) {
+    } catch (error: any) {
```
DominikGuzei, line 87: This might be better and removes the warning about using `any`
```suggestion
} catch (error: unknown) {
```

DominikGuzei: or maybe even using the `Error` type works?
```suggestion
} catch (error: Error) {
```

DominikGuzei (approved):
LGTM 🚢  just left some minor suggestions

packages/staking/src/features/Drawer/TransactionFail.tsx
AngelCastilloB: It cant be unknown becuase we need to access a fields:

```
 if (error.message === 'MULTIDELEGATION_NOT_SUPPORTED') 
```

and it cant be Error, because I get this error:

```
TS1196: Catch clause variable type annotation must be any or unknown if specified
```

DominikGuzei: I think then you can use `unknown` and have to check for the `message` field to exist:

```ts
const isErrorType = typeof error === 'object' && error !== null && 'message' in error;

if (isErrorType && error.message === 'MULTIDELEGATION_NOT_SUPPORTED') {
  portfolioMutators.executeCommand({ type: 'HwSkipToDeviceFailure' });
}
```

DominikGuzei: @AngelCastilloB so in many places we are actually throwing specific instances of error classes which then can be checked against `instanceof` like here:
https://github.com/input-output-hk/lace/blob/0d1d68507430eaf3a75e6fabbea58f7a04abbbdb/apps/browser-extension-wallet/src/features/dapp/components/SignData.tsx#L32

DominikGuzei: In the SDK there are helper functions to construct these … but tbh i haven't looked deeply into why this is accepted from TS perspective while "normal" errors are not. @mkazlauskas probably knows more details about it
https://github.com/input-output-hk/cardano-js-sdk/blob/6b9fef16ffc854ee6ea8954714c94ccb2b788a63/packages/key-management/src/errors/AuthenticationError.ts#L8

DominikGuzei: @AngelCastilloB ah and i can see that there is a workaround to check against normal errors like in your example:
https://stackoverflow.com/questions/64452484/how-can-i-safely-access-caught-error-properties-in-typescript

(the first answer points to TS not allowing `Error` directly in the catch clause but you can use `unknown` and then use the instanceof operator like normal:

```ts
} catch (e: unknown) {
    if (e instanceof Error) {
        console.log(e.message)
    }
}
```

AngelCastilloB: this does work, thanks!. Testing before pushing the fixup

packages/staking/src/features/Drawer/TransactionFail.tsx
```diff
@@ -57,23 +60,37 @@ export const TransactionFailFooter = ({ popupView }: TransactionFailProps): Reac
...
       await signAndSubmitTransaction();
       setIsLoading(false);
       portfolioMutators.executeCommand({ type: 'DrawerContinue' });
       removePassword();
-    } catch (error) {
+    } catch (error: unknown) {
       console.error('failed to sign or submit tx due to:', error);
       setIsLoading(false);
+
+      if (error instanceof Error && error.message === 'MULTIDELEGATION_NOT_SUPPORTED') {
```
DominikGuzei, line 91: Nice :)

tomislavhoracek (approved):
@AngelCastilloB 🚀

PR #1003, opened 2024-03-28
[LW-10080] Single delegation now opens drawer on transition and fetches pools

VanessaPC:
# Checklist

- [x] JIRA - LW-10080
- [x] Screenshots added.

---

## Proposed solution
Remove the check that returns if it's a hardware wallet so we can still fetch the pools. 
Search the by id when searching for a pool id so it returns the correct pool ID and the flow continues in a full screen.

## Testing
- Run the app with a hardware wallet 
- Go to pop up view and try to stake 
- Switch to full screen and see the drawer opened with the delegation you started in the popup view 
- Go back to pop up view and check that the search and stake pools in delegation tab have the correct number of pools (274 in preprod) and the search works.

## Screenshots
Screenshot 2024-03-28 at 08 48 22


pczeglik-iohk (approved):
Code looks good. Thank you @VanessaPC

DominikGuzei (approved):

PR #1017, opened 2024-04-03
fix: send alias during wallet creation

greatertomi:
# Checklist

- [x] JIRA - LW-10181
- [ ] Screenshots added.

---

## Proposed solution

Send Alias event

## Testing

## Screenshots

### DIFF
--- a/apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/WalletSetupWizard.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/WalletSetupWizard.tsx
@@ -163,7 +163,6 @@ export const WalletSetupWizard = ({
   const goToMyWallet = useCallback(
     async (cardanoWallet: Wallet.CardanoWallet = walletInstance) => {
       if (enhancedAnalyticsStatus === EnhancedAnalyticsOptInStatus.OptedIn) {
-        void analytics.sendAliasEvent();
         const addresses = await firstValueFrom(cardanoWallet?.wallet?.addresses$.pipe(filter((a) => a.length > 0)));
         const hdWalletDiscovered = addresses.some((addr) => !isScriptAddress(addr) && addr.index > 0);
         if (hdWalletDiscovered) {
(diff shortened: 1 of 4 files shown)


danielmain (approved):

oldGreg5 (approved):

apps/browser-extension-wallet/src/views/browser-view/features/wallet-setup/components/WalletSetupWizard.tsx
```diff
@@ -213,8 +212,9 @@ export const WalletSetupWizard = ({
   );
 
   const handleSubmit = async (result: { password: string; walletName: string }) => {
-    sendAnalytics(postHogOnboardingActions[setupType]?.ENTER_WALLET);
+    void sendAnalytics(postHogOnboardingActions[setupType]?.ENTER_WALLET);
     await handleCompleteCreation(result.walletName, result.password);
+    void analytics.sendAliasEvent();
```
lucas-barros, line 217: I think we should only send the alias if the user has opted in for analytics.
As in [enhancedAnalyticsStatus === EnhancedAnalyticsOptInStatus.OptedIn](https://github.com/input-output-hk/lace/pull/1017/files#diff-cdc8ed49ef37bc5910fcc1587654a09bcc942f30e675bbafabd28028da9ce683L165)

greatertomi: Analytics rejection or acceptance is handled globally. When analytics is rejected `shouldOmitEvent` [here](https://github.com/input-output-hk/lace/blob/4879c2987a37d695464e977542c8ce48bf08bed2/apps/browser-extension-wallet/src/providers/AnalyticsProvider/analyticsTracker/AnalyticsTracker.ts#L94) would always be `true`, so it won't send an event to posthog

lucas-barros: You're right. Thanks!

PR #1022, opened 2024-04-04
Fix post integration unit test run

pczeglik-iohk:
Post integration workflow is misconfigured. I aligned config to the same we have in [ci.yml](https://github.com/input-output-hk/lace/blob/main/.github/workflows/ci.yml#L32).

Those two configs will be improved in close future. The common subsets of workflows are going to be extracted to actions and reused.


mkazlauskas (approved):

rhyslbw (approved):
Thanks @pczeglik-iohk

PR #1030, opened 2024-04-05
fix(staking): [LW-9583] storybook LocalThemeProvider, vanilla-extract usage and new linter rules

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-9583)

---

## Proposed solution
This PR revamps our approach to `vanilla-extract` in the `staking` package. The staking package now has its own `sx` extending the fundamental sprinkle properties from `@lace/ui`, also the new `LocalThemeProvider` is created with a custom staking theme.

Additionally, a new `no-restricted-imports` ESLint rule is introduced to enforce proper usage of `vanilla-extract` in `staking`.

Almost all of the existing problems with `sx` usage (in `staking`) are now resolved, we have better type-safety (and auto-completion) - it introduces lots of changes, so it's suggested to review this PR commit by commit.

**Important:** This PR exposes a new object from the `@lace/ui` called `sprinklesConfig` used to configure the `sx` per package. There's a new depreciation note above the re-export of `sx` but using it inside of `@lace/ui` is perfectly valid. `LocalThemeProvider` now allows passing `customDarkTheme` and `customLightTheme` to work with custom theme variables in package storybooks.

## Testing
Dear QAs, I suggest testing full regression for styling/theming in App/Storybook and especially in the following components/areas:
- BrowsePools
- BrowsePoolsPreferencesCard (top-right corner sort/filters card)
- PorfolioBar ("basket" for stake pools, displayed at the bottom of the BrowsePools)
- StakePoolsGrid
- StakePoolsList
- DelegationCard (Overview)


packages/staking/src/features/Drawer/preferences/NoPoolsSelected.css.ts
```diff
@@ -8,9 +8,9 @@ export const container = style({
 });
 
 export const icon = style({
-  fontSize: theme.spacing.$112,
+  fontSize: 112,
```
mchappell, line 11:
```suggestion
  fontSize: '$112',
``` ?

**Review by mchappell — APPROVED** [review]
Nice :)

File: packages/staking/src/features/Drawer/preferences/NoPoolsSelected.css.ts (hunk 1 above)
Comment (przemyslaw-wlodek) [reply packages/staking/src/features/Drawer/preferences/NoPoolsSelected.css.ts:11]: You can't use `$` values with `style` (that's why previously it was `theme.spacing.$112`. I decided to leave it like this because we need a task for refactoring (I may create one). A mixture of icon sizes and font sizes in one place feels wrong but if we need this I'd consider something similar to:
```
sx({ fontSize: '$icon_112' })
```

tommayeliog (approved):
Good work @przemyslaw-wlodek

PR #1033, opened 2024-04-05
[LW 10170] Store own addresses from all wallets and accounts

lucas-barros:
# Checklist

- [x] JIRA - https://input-output.atlassian.net/browse/LW-10170
- [x] Screenshots added.

---

## Proposed solution

Store all wallets addresses in each wallet repository metadata

## Testing

Account or wallet needs to be activated at least once before making a transaction.
Make transactions between different accounts or wallets and see that receiving addresses is correctly tagged as "own".

## Screenshots

### DIFF
--- a/apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts
+++ b/apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts
@@ -1,0 +1,107 @@
+/* eslint-disable no-magic-numbers */
+/* eslint-disable unicorn/no-useless-undefined */
+import { WalletManager, WalletRepository } from '@cardano-sdk/web-extension';
+import { cacheActivatedWalletAddressSubscription } from '../cache-wallets-address';
+import { Wallet } from '@lace/cardano';
+import { BehaviorSubject, of } from 'rxjs';
+
+describe('cacheActivatedWalletAddressSubscription', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should not trigger subscription for no active wallet', () => {
+    const mockWalletManager = {
+      activeWallet$: of(undefined),
+      activeWalletId$: of(undefined)
+    } as unknown as WalletManager<Wallet.WalletMetadata, Wallet.AccountMetadata>;
+
+    const mockWalletRepository = {
+      wallets$: of([]),
+      updateWalletMetadata: jest.fn()
+    } as unknown as WalletRepository<Wallet.WalletMetadata, Wallet.AccountMetadata>;
+
+    cacheActivatedWalletAddressSubscription(mockWalletManager, mockWalletRepository);
+
+    expect(mockWalletRepository.updateWalletMetadata).not.toHaveBeenCalled();
+  });
+
+  it('should subscribe and update metadata', () => {
+    const mockWalletManager = {
+      activeWallet$: of({ addresses$: of([{ address: 'address1' }]) }),
+      activeWalletId$: of({ walletId: 'walletId' })
+    } as unknown as WalletManager<Wallet.WalletMetadata, Wallet.AccountMetadata>;
+
+    const mockWalletRepository = {
+      wallets$: of([
+        {
+          walletId: 'walletId',
+          metadata: {}
+        }
+      ]),
+      updateWalletMetadata: jest.fn()
+    } as unknown as WalletRepository<Wallet.WalletMetadata, Wallet.AccountMetadata>;
+
+    cacheActivatedWalletAddressSubscription(mockWalletManager, mockWalletRepository);
+
+    expect(mockWalletRepository.updateWalletMetadata).toHaveBeenCalledWith({
+      walletId: 'walletId',
+      metadata: {
+        walletAddresses: ['address1']
+      }
+    });
+  });
+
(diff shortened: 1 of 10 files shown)


apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable no-magic-numbers */
```
VanessaPC, line 1: I think this is not needed.

apps/browser-extension-wallet/src/lib/scripts/background/cache-wallets-address.ts
```diff
@@ -0,0 +1,34 @@
+import { WalletManager, WalletRepository } from '@cardano-sdk/web-extension';
+import { Wallet } from '@lace/cardano';
+import { filter, switchMap, withLatestFrom, zip } from 'rxjs';
+
+export const cacheWalletAddressSubscription = (
+  walletManager: WalletManager<Wallet.WalletMetadata, Wallet.AccountMetadata>,
+  walletRepository: WalletRepository<Wallet.WalletMetadata, Wallet.AccountMetadata>
+): void => {
+  zip([
```
VanessaPC, line 9: I think you can `combinelatest` to get the `walletId`, `activeWallet` and `wallets` of the currently account/wallet? 
In that case you combine it into a single observable. 
If you use a map operator to transform the array into an object you can filter out values faster.

VanessaPC (approved):
Nice work! Left a couple of comments but it works great.

apps/browser-extension-wallet/src/lib/scripts/background/cache-wallets-address.ts
```diff
@@ -0,0 +1,34 @@
...
+        []
+      );
+
+      const addresses = new Set([...walletAddresses.map(({ address }) => address), ...allCachedWalletAddresses]);
+
+      walletRepository.updateWalletMetadata({
+        walletId: activeWallet.walletId,
+        metadata: {
+          ...wallet.metadata,
+          addresses: [...addresses]
```
DominikGuzei, line 30: Maybe `addresses` could be a bit too generic for the metadata section of the active wallet? (i think this could easily be misinterpreted as only the active wallets addresses). What about `allOwnAddresses` or something more descriptive?

DominikGuzei (approved):

apps/browser-extension-wallet/src/lib/scripts/background/cache-wallets-address.ts
lucas-barros: I initially tried `combinelatest`, but I cannot use it with `wallets` since it is updated in the subscription and it causes a trigger loop. Also, because these observables will emit values in sequence, I had to use zip to get "synced" values just once when activate function from `walletManager` is called: https://github.com/input-output-hk/cardano-js-sdk/blob/master/packages/web-extension/src/walletManager/walletManager.ts#L156.

apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts
lucas-barros: I get a warning here: https://github.com/input-output-hk/lace/blob/f9805506a75cf9dbbde0c4d3c7aabf7df86b5b4b/apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts#L100
"No magic number: 2."

apps/browser-extension-wallet/src/lib/scripts/background/cache-wallets-address.ts
lucas-barros: Makes sense, updated here: https://github.com/input-output-hk/lace/pull/1033/commits/3d29a04b87f0404dac09a73533fb5bebec5bc16d

apps/browser-extension-wallet/src/lib/scripts/background/cache-wallets-address.ts
VanessaPC: This makes sense! In that case, great work!

VanessaPC:
This looks good to me @lucas-barros , great work!

apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    expect(mockWalletRepository.updateWalletMetadata).toHaveBeenNthCalledWith(2, {
+      walletId: 'walletId2',
+      metadata: {
+        walletAddresses: ['address2', 'address3']
+      }
+    });
+  });
+
+  it('should not save duplicated addresses', () => {
```
DominikGuzei, line 108: isn't this already tested in the previous case?

DominikGuzei:
re-approved 👍

apps/browser-extension-wallet/src/lib/scripts/background/__tests__/cache-wallets-address.test.ts
lucas-barros: Yes, you're right!

PR #1038, opened 2024-04-08
test(extension): fixes

oldGreg5:
regression [here](https://17a5-185-161-93-8.ngrok-free.app/index.html)

### DIFF
--- a/packages/e2e-tests/src/elements/onboarding/mainPage.ts
+++ b/packages/e2e-tests/src/elements/onboarding/mainPage.ts
@@ -104,9 +104,6 @@ class OnboardingMainPage extends CommonOnboardingElements {
       case 'Terms of service':
         await this.agreementTermsOfServiceLink.click();
         break;
-      case 'Cookie policy':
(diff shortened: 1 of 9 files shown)


packages/e2e-tests/src/elements/onboarding/mainPage.ts
```diff
@@ -104,6 +104,9 @@ class OnboardingMainPage extends CommonOnboardingElements {
       case 'Terms of service':
         await this.agreementTermsOfServiceLink.click();
         break;
+      case 'Cookie policy':
```
wklos-iohk, line 107: Please make sure if it fits here, as I understand `clickOnLegalLink(...)` method applies to those 2 links only:
image

Probably you'll have to use different step in test scenario that uses `clickOnLegalLinkOnFooter(...)` from CommonOnboardingElements class -> `I click on "<legal_link>" legal link`

wklos-iohk (changes requested):

wklos-iohk (approved):

PR #1042, opened 2024-04-08
Update Sonar Cloud workflow configuration

pczeglik-iohk:
In my past PR I used a wrong pattern for release branch name and after reviewing this workflow it looks it's not needed to track feat, fix, etc branches as the same information we get from PRs in SonarCloud.

### DIFF
--- a/.github/workflows/sonar-cloud.yml
+++ b/.github/workflows/sonar-cloud.yml
@@ -3,18 +3,13 @@ on:
   push:
     branches:
       - main
-      - release/**/*
-      - feat/*
-      - feat-*
-      - fix/*
-      - fix-*
-      - lw-*
+      - release/**
   pull_request:
     types: [opened, synchronize, reopened]
 jobs:
   sonarcloud:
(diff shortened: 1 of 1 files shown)


.github/workflows/sonar-cloud.yml
```diff
@@ -3,12 +3,7 @@ on:
   push:
     branches:
       - main
-      - release/**/*
-      - feat/*
-      - feat-*
-      - fix/*
-      - fix-*
-      - lw-*
+      - release/**
```
pczeglik-iohk, line 6: wrong pattern and release branches

.github/workflows/sonar-cloud.yml
```diff
@@ -3,12 +3,7 @@ on:
   push:
     branches:
       - main
-      - release/**/*
-      - feat/*
```
pczeglik-iohk, line 7: this is the same data we get for PRs, so it turned out it's not needed to track feats, fixes etc.

mchappell (approved):

PR #1050, opened 2024-04-10
LW-10255 - update text for term and privacy policy

danielmain:
# Description
This PR changes the text displayed in the onboarding flow

# Screenshot

image

### DIFF
--- a/packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.tsx
+++ b/packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.tsx
@@ -66,7 +66,7 @@ export const WalletSetupOptionsStepRevamp = ({
       </div>
     </div>
     <div className={styles.legal} data-testid="agreement-text">
-      By proceeding you agree to Lace’s{' '}
+      By clicking the Create, Connect or Restore button above, you agree with Lace’s{' '}
       <a href={TERMS_OF_USE_URL} target="_blank" className={styles.link} data-testid="agreement-terms-of-service-link">
         Terms of Service
       </a>{' '}
--- a/packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.module.scss
+++ b/packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.module.scss
@@ -53,4 +53,6 @@
 .legal {
   margin-top: size_unit(2);
   text-align: center;
+  margin-left: -5px;
+  margin-right: -5px;
 }


pczeglik-iohk:
I would break this line into two. In my opinion that would look much better
```
By clicking the Create, Connect or Restore button above,
you agree with Lace’s Terms of Service and Privacy Policy.
```

packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.tsx
```diff
@@ -66,7 +66,7 @@ export const WalletSetupOptionsStepRevamp = ({
       </div>
     </div>
     <div className={styles.legal} data-testid="agreement-text">
-      By proceeding you agree to Lace’s{' '}
+      By clicking the Create, Connect or Restore button above, you agree with Lace’s{' '}
```
pczeglik-iohk, line 69: hm, it looks this text is not using i18n :)

danielmain: @pczeglik-iohk lets improve it after 1.10 🙏🏻 this is a quick action not to block the release. Also it was not in i18 before.

szymonmaslowski (approved):
Let's apply i18n changes in a subsequent PR

packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.tsx
pczeglik-iohk: yeah, for sure it does not make sense to do it now. But it would be good to improve it soon, as we want to provide Japanese language support for Lace

pczeglik-iohk (approved):

danielmain:
> I would break this line into two. In my opinion that would look much better
> 
> ```
> By clicking the Create, Connect or Restore button above,
> you agree with Lace’s Terms of Service and Privacy Policy.
> ```

@pczeglik-iohk we agreed with @vhulchenko-iohk and @allanleone to add a horizontal -5px margin and keep it in one line.

packages/core/src/ui/components/WalletSetupRevamp/WalletSetupOptionsStepRevamp.tsx
```diff
@@ -66,7 +66,7 @@ export const WalletSetupOptionsStepRevamp = ({
       </div>
     </div>
     <div className={styles.legal} data-testid="agreement-text">
```
szymonmaslowski, line 68:
```suggestion
    {/* TODO: remove inline style and allow for a single line with a better way */}
    <div className={styles.legal} style={{ marginLeft: -20, marginRight: -20 }} data-testid="agreement-text">
```

PR #1076, opened 2024-04-18
LW-10242 fix for enter element to be picked in the dom

VanessaPC:
# Checklist

- [x] JIRA - LW-10242
- [ ] Screenshots added.

---

## Proposed solution

Destroy on close the component that is stopping this button from being picked up by the DOM. 

## Testing

With a recently loaded wallet, move to production (do not reload) click on an asset token, and when it opens, press click, this should take you to the send flow. 

## Screenshots


DominikGuzei (approved):

apps/browser-extension-wallet/src/hooks/useEnterKeyPress.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-lonely-if */
```
lucas-barros, line 1: Is this needed?

lucas-barros (approved):

apps/browser-extension-wallet/src/hooks/useEnterKeyPress.ts
VanessaPC: you are right, fixed now!

PR #1086, opened 2024-04-22
test(extension): fix smoke tests

bslabiak:
# Checklist

- [ ] JIRA - \<link>
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots

### DIFF
--- a/packages/e2e-tests/src/assert/transactionDetailsAssert.ts
+++ b/packages/e2e-tests/src/assert/transactionDetailsAssert.ts
@@ -109,13 +109,8 @@ class TransactionsDetailsAssert {
     for (let i = 0; i <= rowsNumber && i < 10; i++) {
       await TransactionsPage.clickOnTransactionRow(i);
       await TransactionDetailsPage.transactionDetailsDescription.waitForClickable({ timeout: 15_000 });
-      await TransactionDetailsPage.transactionDetailsHash.waitForDisplayed();
-      await TransactionDetailsPage.transactionDetailsStatus.waitForDisplayed();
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -124,6 +119,13 @@ class TransactionsDetailsAssert {
         await TransactionDetailsPage.transactionDetailsStakepoolTicker.waitForDisplayed();
         await TransactionDetailsPage.transactionDetailsStakePoolId.waitForDisplayed();
       }
+      if (!txType.includes('Rewards')) {
```
wklos-iohk, line 122: Do we have new Figma screens for `Rewards` transactions? In the old one I see fields like status and timestamp available: https://www.figma.com/file/RPN3V4Et7hMbXkV5ZIF75i/Lace-Browser---Full-view-UI-Design?type=design&node-id=12354-232569&mode=design&t=xq6FogjqZwUf2Tpw-0

ljagiela: IMO, It is out of scope of this PR and the test itself. We should probably implement specific test for checking rewards tx type .

ljagiela (approved):

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
bslabiak: Should be implemented in specific rewards test case and preferably in specific wallet.

PR #1097, opened 2024-04-23
Rework CI

pczeglik-iohk:
# Checklist

- [x] JIRA: [LW-10413](https://input-output.atlassian.net/browse/LW-10413)
- [ ] Screenshots added.

[LW-10413]: https://input-output.atlassian.net/browse/LW-10413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yml
```diff
@@ -1,41 +1,205 @@
...
+      with:
+        name: packages-staking
+        path: packages/staking/dist
+
+    - name: Collect Workflow Telemetry Smoke Tests
+      uses: catchpoint/workflow-telemetry-action@v2
+      with:
+        comment_on_pr: false
+
+    - name: Build Lace browser extension
```
emiride, line 183: Would it make sense to move building in preparation phase?

.github/workflows/ci.yml
```diff
@@ -1,41 +1,205 @@
...
 
 on:
   pull_request:
   push:
     branches:
+      - main
       - 'release/**'
 
+permissions:
+  pull-requests: write
```
pczeglik-iohk, line 11: this permission is needed for `catchpoint/workflow-telemetry-action@v2` action

.github/CODEOWNERS
```diff
@@ -1,11 +1,13 @@
 # Root Level
 *                       @input-output-hk/lace-tech-leads
 yarn.lock               @input-output-hk/lace-core
+
 # Packages Teams
 /packages/cardano/      @input-output-hk/lace-core
 /packages/common/       @input-output-hk/lace-core
 /packages/core/         @input-output-hk/lace-core
 /packages/e2e-tests/    @input-output-hk/lace-test-engineers
+/packages/icons/        @input-output-hk/lace-core
```
pczeglik-iohk, line 10: this has been missed when icons package has been created.

.github/workflows/ci.yml
emiride: OK.

pparker:
@pczeglik-iohk - what Jira ticket is this PR related to?

It's mentioned in [this comment](https://input-output.atlassian.net/browse/LW-10075?focusedCommentId=287805) in https://input-output.atlassian.net/browse/LW-10075, so - maybe that one?

emiride (approved):
Looks good to me

.github/workflows/ci.yml
```diff
@@ -1,52 +1,207 @@
...
 
 permissions:
   pull-requests: write
   actions: read
 
 jobs:
-  buildAndTest:
-    name: Build & Test
-    runs-on: self-hosted
-    concurrency:
```
mchappell, line 17: @pczeglik-iohk is there a reason we removed concurrency?

pczeglik-iohk: This concurrency is for self-hosted runner which I do not use. Self hosted runner creates long queues. And in current shape of CI one PR with changes to staking consumes all 4 available runners. 1 for build staking, 1 for CI workflow, 1 for chromatic, and 1 for smoke tests. Some PRs had to wait more then an hour to execute

mchappell: What I mean is, should we not include this in the new reworked CI, particularly `cancel-in-progress: true`

pczeglik-iohk: you are right @mchappell. I reverted this change

.github/settings.yml
```diff
@@ -91,9 +91,9 @@ branches:
         strict: false
         # Required. The list of status checks to require in order to merge into this branch
         checks:
-          - context: Build & Test
           - context: block-fixup
-          - context: Build Staking Center
+          - context: Unit tests
+          - context: Smoke tests
```
mchappell, line 96: should `prepare` be an explicit step? Otherwise we're only testing it by proxy of these other tests, if we make modification to this in future where we change the workflow we might accidentally remove checking formatting/linting/etc

Also the chromatic steps should be part of these checks, this should have been added via code instead of the UI which is why it doesn't appear here.

pczeglik-iohk: Lets sync about this :)

mchappell (approved):
Great rework! 🍻

PR #1099, opened 2024-04-23
[LW-9793] - chore(extension): bump cardano-js-sdk, enable trezor

danielmain:
This PR enables Trezor support after adding suggestions of the cardano-js-sdk audit by bumping it

- [ ] JIRA - LW-9860 | LW-9793
- [ ] Trezor regression test


pczeglik-iohk (approved):

emiride (approved):
OK, I tested this and it seems to be working fine with live-preview. Note here is that it doesn’t work on dev-preview, but this is something unrelated to Lace.
Tested:
Sending and receiving transactions
Staking
Transactions through dApps

PR #1109, opened 2024-04-26
feat(ui,core,staking): [LW-9594] refactor Select component + design updates

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-9594)
- [x] Screenshots added.

---

## Proposed solution
This PR is introducing a new design and API of the Select component. We aim to make it as flexible as possible, which is the first step.

## Testing
Please test the existing usage of the component:
- `WalletSetupSelectAccountsStepRevamp` (Revamped Wallet setup flow)
- `QuorumOption` (Shared wallet feature)
- `BrowsePoolsPreferencesCard` (Staking - Browse Pools; Enabled with feature flag `USE_MULTI_DELEGATION_STAKING_FILTERS`; it may be refactored during "advanced filters" ticket)

## Screenshots
image

image

image

image


packages/staking/src/features/BrowsePools/BrowsePoolsPreferencesCard/BrowsePoolsPreferencesCard.tsx
```diff
@@ -123,15 +123,17 @@ export const BrowsePoolsPreferencesCard = ({
...
         showArrow
-        withOutline
-        className={styles.selectGroup}
         placeholder={t('browsePools.preferencesCard.filter.input.select')}
-        options={filterOption.opts as SelectOption[]}
-        selectedValue={selectedValue}
-      />
+        value={selectedValue}
+      >
+        {(filterOption.opts as SelectOption[]).map((opt) => (
```
vetalcore, line 133: `filterOption.opts` is typed as `string[] | SelectOption[];`, why would we cast it always as SelectOption[]? should we fix types instead?

przemyslaw-wlodek: Indeed, but this should be a part of the Advanced Filters task soon :) I am just replacing the component with a new API. I wouldn't touch it because types in this are will drastically change (using SDK types, etc).

packages/ui/src/design-system/decorators/page-provider.component.tsx
```diff
@@ -10,13 +10,21 @@ interface Props {
 export const PageProvider = ({
   children,
 }: Readonly<Props>): React.ReactElement => {
-  const [state, setState] = useState<HTMLElement>();
+  const [state, setState] = useState<{
```
mchappell, line 13: would it be better to reference the interface from [here?](https://github.com/input-output-hk/lace/pull/1109/files#diff-d27c6e5fd24e1347123b3b00ed1f08ff8f7a675c2e1c8ea87c7654da9c76e2d1R3)

packages/ui/src/design-system/select/select-root.component.tsx
```diff
@@ -0,0 +1,121 @@
...
+          position={align === 'selected' ? 'item-aligned' : 'popper'}
+        >
+          <Select.Viewport>
+            {React.Children.map(children, child => {
+              if (isValidSelectRootChild(child)) {
+                return React.cloneElement(child, {
+                  variant,
+                });
+              }
+              console.error(
```
mchappell, line 110: if we're saying that it 'must' be either an Item, or ItemRoot, we should throw here instead of just logging.

packages/ui/src/design-system/select/select-item/select-item-root.component.css.ts
```diff
@@ -0,0 +1,59 @@
+import { styleVariants } from '@vanilla-extract/css';
+
+import { style, vars } from '../../../design-tokens';
```
mchappell, line 3: we're mixing import statements between this style, and https://github.com/input-output-hk/lace/pull/1109/files#diff-cb50114b0683d1eb185b349a57198b7b54c61063f8942d31e9f5a5db67e9822fR1
please align

mchappell (dismissed):

packages/ui/src/design-system/select/select-item/select-item-root.component.css.ts
przemyslaw-wlodek: Good catch. We should remove re-exports that don't make sense, like `style` and use it directly from `vanilla-extract/css`. I would suggest a new though. Alternatively I can fix occurrences in this PR but it's not bringing any value if we still need to redo this for the whole codebase (not only `packages/ui`)

DominikGuzei (approved):
Great work 👏

packages/ui/src/design-system/decorators/page-section.component.tsx
```diff
@@ -4,15 +4,29 @@ import React from 'react';
 import { Box } from '../box';
 import { Text } from '../text';
 
+import { subtitleBox } from './page.css';
```
vetalcore, line 7: should we follow `import * as namespace` pattern?

packages/ui/src/design-system/select/select-item/select-item-root.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+      className={cx.root[variant]}
+      data-testid={testId}
+    >
+      {children}
+    </Select.Item>
+  ),
+);
+
+// eslint-disable-next-line functional/immutable-data
+ItemRoot.displayName = 'ItemRoot';
```
vetalcore, line 41:
```suggestion
export const ItemRootComponent = (
  {
    children,
    disabled,
    value,
    variant = 'plain',
    testId,
  }: Readonly<SelectItemPrivateProps & SelectItemRootProps>,
  forwardReference: React.Ref<HTMLDivElement>,
): React.ReactElement => (
  <Select.Item
    ref={forwardReference}
    disabled={disabled}
    value={value}
    className={cx.root[variant]}
    data-testid={testId}
  >
    {children}
  </Select.Item>
);

export const ItemRoot = forwardRef(ItemRootComponent);
```

just a suggestion

vetalcore:
Nicely done, @przemyslaw-wlodek.
just a small suggestion if we could consider in the future:
- splitting this or into 3 pieces: 1. component refactor, 2. updated usage. 3. storybook theme provider fixes/updates (not sure i understood if portal piece has been implemented here or now)
- or doing the same but as 3 separate commits

it could have really helped with the review and delivery.

packages/ui/src/design-system/select/select-item/select-item-root.component.tsx
przemyslaw-wlodek: Could be an option 🤔 but I think it complicates documentation. JSDoc has to be placed above the exported component and you may lack e.g. IDE support for `@param` annotations, etc. This is something to be discovered but I will leave it as-is for now because it works fine.

packages/ui/src/design-system/decorators/page-section.component.tsx
przemyslaw-wlodek: It was like this before 🤔 copy/paste probably -> now fixed

PR #1154, opened 2024-05-11
Chore/lw 10534 translation package

vetalcore:
# Checklist

- [x] JIRA -[LW-10534](https://input-output.atlassian.net/browse/LW-10534)
- [ ] Proper tests implemented
- [ ] Screenshots added.

---

## Proposed solution

Move all the translations files and i18n instance into the separate package.

## Testing

All the translations should work the same as before.

## Screenshots

~~Attach screenshots here if implementation involves some UI changes~~

[LW-10534]: https://input-output.atlassian.net/browse/LW-10534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1156, opened 2024-05-13
feat(extension): add midnight pre-launch settings banner LW-10398

DominikGuzei:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-10398)
- [x] Screenshots added.

---

## Proposed solution

Please note: the CTA buttons do not have any action defined yet (as per task requirement) and will be wired-up in later tasks.

- Introduce a `USE_MIDNIGHT_PRELAUNCH_EVENT` feature flag to hide the midnight related UI elements behind
- Implement the midnight pre-launch banner in `@lace/core` storybook
- Add the pre-launch ui setting item in the settings list (in all views)
- Integrate the pre-launch banner in the settings sidebar (in extended view)

## Testing

- Check that no pre-launch settings items are shown when feature flag is off
- With feature flag on check that the pre-launch settings item is always visible and the sidebar banner is visible in extended view in all responsive screen sizes except the smallest one.

## Screenshots

video of the responsive banner sidebar banner (please note that at time of recording the settings list item was not yet activated for extended view but is in latest version):

https://github.com/input-output-hk/lace/assets/172414/075372c3-50b8-4330-868a-78aa34b8e4c2


packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
```diff
@@ -0,0 +1,27 @@
...
+  onMoreDetailsClick
+}: {
+  bannerImageUrl: string;
+  onMoreDetailsClick?: () => void;
+}): React.ReactElement => {
+  const { t } = useTranslate();
+  return (
+    <Card.Outlined
+      className={styles.card}
+      style={{ '--midnight-pre-launch-banner-image': `url(${bannerImageUrl})` } as React.CSSProperties}
```
DominikGuzei, line 17: That was the only way I found to provide a background image via CSS variable to the component (since it didn't work to include the background-image in the compiled core component)

packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Button, Card } from '@lace/ui';
+import { useTranslate } from '@src/ui/hooks';
+import styles from './MidnightPreLaunchSettingsBanner.module.scss';
+
+export const MidnightPreLaunchSettingsBanner = ({
+  bannerImageUrl,
```
pczeglik-iohk, line 7: I have mixed feelings about `bannerImageUrl` prop. In my opinion this prop would have more sense if this banner was generic one.

When I look at place where it is used I see this weird looking import  `import MidnightPreLaunchBannerImage from '../../../../../../../../packages/core/src/ui/assets/images/midnight-launch-event-sidebar-banner.png';` 

I would recommend removing this prop, or making this component generic like `<SettingsBanner />` and place image in apps/browser-extension-wallet directory. The same for storybook.

packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
```diff
@@ -0,0 +1,27 @@
...
+  onMoreDetailsClick
```
pczeglik-iohk, line 8: Can we instead of `onMoreDetailsClick` use just `onClick`?

In case of button text change we won't need to change prop name to have consistency between what we see in UI and what we have in code.

packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
pczeglik-iohk: The more I think about it, the more I like the idea of making this component generic one, especially that it's implementation is in `core` package. If we would like to keep this name `MidnightPreLaunchSettingsBanner` to me it would make more sense to have it in `apps/browser-extension-wallet` as this is not something generic and there is probably 0% chance we would like to reuse it somewhere else.

DominikGuzei: @pczeglik-iohk I had to introduce the prop because the image must be used as CSS background (so that the rounded corners of the banner are correctly respected) AND unfortunately our core package doesn't correctly bundle css images (the path is simply retained and then wrong when component is imported in the extension) - I can move the image to the extension package but then what should be shown in the core storybook (it also acts as an example)?

packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
DominikGuzei: what about `onCtaButtonClick`?

packages/core/src/ui/lib/translations/en.json
```diff
@@ -250,6 +250,9 @@
   "core.infoWallet.addressCopied": "Address copied",
   "core.infoWallet.copy": "Copy",
   "core.infoWallet.handleCopied": "Handle copied",
+  "core.MidnightPreLaunchSettingsBanner.heading": "Midnight Pre-Launch Event",
```
pczeglik-iohk, line 253: If we decide to go with generic component translations should be in apps/browser-extension-wallet?

I would recommend creating dedicated namespace for all midnight related translations like `browserView.midnight` or just `midnight`.

pczeglik-iohk: just like you did it here https://github.com/input-output-hk/lace/pull/1156/files#diff-1ca0f877e9a3ed7324d37abb38e80309f71083101665c52a39e7c338581db781R176

pczeglik-iohk (changes requested):
Looks good @DominikGuzei, a few changes proposed. 

Let me know what do you think :)

packages/core/src/ui/lib/translations/en.json
DominikGuzei: @pczeglik-iohk it's not a generic component, the only reason why we have storybook in core package is because we haven't been able to set it up correctly in the extension app package 😆 if you look around you will see plenty of non-generic components in core package. Only `@lace/ui` is truly generic.

In both cases i followed the existing naming scheme for translations in the app and core package. I'm not sure if introducing a dedicated namespace will make things clearer in the long run.

btw. the way we handle translations will be changed dramatically anyway (moved into a dedicated package): https://github.com/input-output-hk/lace/pull/1154

packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
VanessaPC: I agree with @DominikGuzei I think the issue is on the way we have of working and exporting images from `core` which should be addressed. But I think the way it was set up it's consistent with how other components currently do it.

DominikGuzei: @pczeglik-iohk sure, i can turn this into a generic component BUT it doesn't really make sense since the design is quite special in that it looks the same in light/dark theme (font colors). I'm not sure if it's the best case for a generic component tbh.

packages/core/src/ui/components/MidnightPreLaunchSettingsBanner/MidnightPreLaunchSettingsBanner.tsx
DominikGuzei: @pczeglik-iohk renamed to `onCtaButtonClick` and updated

pczeglik-iohk (approved):
Not 100% happy, but I totally understand your arguments @DominikGuzei 

Code approved! Thank you @DominikGuzei

PR #1162, opened 2024-05-17
Fix/lw 10502 fix for multiple stake pools requests

vetalcore:
# Checklist

- [x] JIRA - [LW-10502](https://input-output.atlassian.net/browse/LW-10502)
- [ ] Screenshots added.

---

## Proposed solution

- [x] memoize all props of virtuoso related components to avoid additional re-rendering;
- [x] do not unmount virtuoso grid and table component (is not done for grid comp yet);
- [x] remove as match as possible conditional rendering from layout, eg: `EmptyTable` component or `Table Header` is a good example, removing them from the DOM, would cause `Virtuoso components` to change their `height` (**trigger `rangeChange` cb -> `load more data` as a result**)
- [x] in case the next chunk of requested data is smaller then previous one (eg: search query is more specific), Virtuoso component would trigger `rangeChange` cb (obviously, because the range has just changed) in addition to the one we've just fired manual (due to search query change).  **Seems like an expected behaviour for me**?

## Testing

- there should be no that on request for single user input*

## Screenshots

[LW-10502]: https://input-output.atlassian.net/browse/LW-10502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/staking/src/features/BrowsePools/StakePoolsGrid/StakePoolsGridItem.tsx
+++ b/packages/staking/src/features/BrowsePools/StakePoolsGrid/StakePoolsGridItem.tsx
@@ -3,6 +3,7 @@ import { StakePoolCard } from 'features/BrowsePools/StakePoolCard';
 import React from 'react';
 import { useOutsideHandles } from '../../outside-handles-provider';
 import { MAX_POOLS_COUNT, StakePoolDetails, isPoolSelectedSelector, useDelegationPortfolioStore } from '../../store';
+import { DEFAULT_SORT_OPTIONS } from '../constants';
 import { getFormattedStakePoolProp } from '../formatters';
 import { useQueryStakePools } from '../hooks';
 import { SortField } from '../types';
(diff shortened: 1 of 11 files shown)


packages/staking/src/features/BrowsePools/StakePoolsGrid/StakePoolsGridItem.tsx
```diff
@@ -4,20 +4,18 @@ import React from 'react';
...
-  hexId,
-  id,
-  sortField = 'ticker',
-  ...data
-}: StakePoolsGridItemProps): React.ReactElement => {
+export const StakePoolsGridItem = ({ stakePool, hexId, id, ...data }: StakePoolsGridItemProps): React.ReactElement => {
   const { analytics } = useOutsideHandles();
+  const { sort } = useQueryStakePools();
+
+  const sortField = sort?.field || 'ticker';
```
tommayeliog, line 18: @vetalcore just a suggestion, maybe you can use DEFAULT_SORT_OPTIONS.field

tommayeliog (approved):
LGTM, just a tiny comment left

packages/staking/src/features/BrowsePools/StakePoolsGrid/StakePoolsGridItem.tsx
vetalcore: fixed here https://github.com/input-output-hk/lace/pull/1162/commits/7cd176bc31e5cdccaf833c52273b7e73421ea134

wklos-iohk (approved):

PR #1175, opened 2024-05-22
test(extension): e2e for sent/received/self/rewards details of tx

ljagiela:
https://input-output.atlassian.net/browse/LW-10338
https://input-output.atlassian.net/browse/LW-10339

E2E tests for:
https://input-output.atlassian.net/browse/LW-10616
https://input-output.atlassian.net/browse/LW-10617


packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -38,9 +41,10 @@ class TransactionsDetailsAssert {
 
   async assertSeeActivityDetailsDrawer(shouldBeDisplayed: boolean) {
     await TransactionDetailsPage.transactionDetails.waitForDisplayed({ reverse: !shouldBeDisplayed });
+    await TransactionDetailsPage.transactionDetails.waitForStable();
```
wklos-iohk, line 44: This might not work for `assertSeeActivityDetailsDrawer(false)`

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -255,6 +260,211 @@ class TransactionsDetailsAssert {
     const currentMetadata = await TransactionDetailsPage.transactionDetailsMetadata.getText();
     expect(currentMetadata).to.equal(testContext.load('metadata'));
   }
+
+  // eslint-disable-next-line max-statements
+  async assertSeeSelfTransactionDetails(txType: 'Sent' | 'Received' | 'Self') {
+    expect(await TransactionDetailsPage.transactionDetailsType.getText()).to.equal(txType);
```
bslabiak, line 266: assertion name says this is only just for Self tx Type. but we are accepting also 'Sent' | 'Received' | 'Self'. Is this method work for all tx details?

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -255,6 +260,211 @@ class TransactionsDetailsAssert {
...
+        ]);
+      }
+    }
+
+    await TransactionDetailsPage.transactionDetailsStatusTitle.waitForDisplayed();
+    expect(await TransactionDetailsPage.transactionDetailsStatusTitle.getText()).to.equal(
+      await t('core.activityDetails.status')
+    );
+    await TransactionDetailsPage.transactionDetailsHash.waitForDisplayed();
+    expect(await TransactionDetailsPage.transactionDetailsHash.getText()).not.to.be.empty;
```
bslabiak, line 325: `await TransactionDetailsPage.transactionDetailsHash` is beeing checked twice

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -255,6 +260,211 @@ class TransactionsDetailsAssert {
...
+      case 'Self Transaction':
+        await this.assertSeeSelfTransactionDetails('Self');
+        break;
+      case 'Rewards':
+        await this.assertSeeRewardsTransactionDetails();
+        break;
+      default:
+        throw new Error(`Unknown tx type ${txType}`);
+    }
+  }
```
bslabiak, line 467: Do we really need `waitForDisplayed()` check for each UI element?. I suggest just to check few main ones to be sure that tx details are loaded wdyt?

ljagiela: I think that we need this, because otherwise it might check values only, even when element (drawer) is not shown or covered with another element, etc.

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -255,6 +260,211 @@ class TransactionsDetailsAssert {
     const currentMetadata = await TransactionDetailsPage.transactionDetailsMetadata.getText();
     expect(currentMetadata).to.equal(testContext.load('metadata'));
   }
+
+  // eslint-disable-next-line max-statements
+  async assertSeeSelfTransactionDetails(txType: 'Sent' | 'Received' | 'Self') {
```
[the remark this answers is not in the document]
ljagiela: I tried, but there is not that much to extract tbh. Those methods very precisely check every type of transaction (that was the main purpose of this test).

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
```diff
@@ -216,6 +220,7 @@ class TransactionsDetailsAssert {
       await TransactionsPage.clickOnTransactionRow(i);
       await TransactionDetailsPage.transactionDetailsDescription.waitForClickable({ timeout: 15_000 });
       if (
+        // eslint-disable-next-line sonarjs/no-duplicate-string
```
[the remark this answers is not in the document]
ljagiela: IMHO it does not bring any value in terms of code readability.

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
ljagiela: Fixed.

packages/e2e-tests/src/assert/transactionDetailsAssert.ts
ljagiela: Good catch! Fixed.

wklos-iohk (approved):

bslabiak (approved):

PR #1187, opened 2024-05-24
ci: sync fork workflow

pczeglik-iohk:


mchappell (approved):
Great! now let's test it :)

PR #1166, opened 2024-05-20
feat: add cosigner screen to shared wallet flow

greatertomi:
# Checklist

- [x] JIRA - LW-10536
- [x] Proper tests implemented
- [x] Screenshots added.

---

## Proposed solution

- CoSigners step with inputs for name and keys
- Confirmation dialog before quorum step
- Fixed timeline - now previous steps are selected

## Testing

Describe here, how the new implementation can be tested.
Provide link or briefly describe User Acceptance Criteria/Tests that need to be met

## Screenshots

image
image
image
image
image
image

PR #1191, opened 2024-05-27
feat: quorum step of shared wallet creation

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-10537)
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationFlow.tsx
```diff
@@ -17,6 +17,22 @@ export const SharedWalletCreationFlow: VFC = () => (
             walletName={state.walletName}
           />
         )}
+        {state.step === SharedWalletCreationStep.CoSigners && (
+          <span>
+            {'CoSigners '}
+            <button onClick={() => dispatch({ type: 'back' })}>back</button>
+            <button onClick={() => dispatch({ type: 'next' })}>next</button>
+          </span>
+        )}
```
greatertomi, line 26: Maybe you should point to #1166 so that this and several other placeholder implementations of the cosigners can be removed.

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
```diff
@@ -79,11 +90,10 @@ type SharedWalletCreationStoreProps = {
...
   const history = useHistory();
   const { walletRepository } = useWalletManager();
-  const {
-    walletInfo: { name: activeWalletName }
-  } = useWalletStore();
+  const { walletInfo } = useWalletStore();
 
-  const initialState = makeInitialState(activeWalletName);
+  const initialState = makeInitialState(walletInfo?.name || '');
+  // eslint-disable-next-line sonarjs/cognitive-complexity, complexity
```
greatertomi, line 96: Instead of turning this off, we should either refactor this not to use a bunch of ifs or create a general handler for `next` and `back`, maybe `handleBack` and `handleNext` or extract the reducer implementations to another file.

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
```diff
@@ -29,34 +27,46 @@ type MakeState<S extends BaseState> = S;
...
 
-type State = StateSetup | StateCoSigners;
+export type StateQuorum = MakeState<{
+  activeWalletName: string;
+  coSignersKeys: [string, string];
+  quorumOptionValue: QuorumOptionValue;
+  step: SharedWalletCreationStep.Quorum;
+  walletName: string;
+  walletNameInitiallySet: boolean;
+}>;
```
greatertomi, line 52: Having most of these objects repeated four different times does not make sense to me. Can we reduce them to a max of two repetitions?

greatertomi (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationFlow.tsx
szymonmaslowski: I would prefer to avoid stacking PRs TBH. I don't see this as a problem. I just added basic implementation to make it possible to proceed to the subsequent step. If your PR gets merged first I will merge your changes in, if my goes first you will override my changes. Does it work for you?

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
szymonmaslowski: This might look like a lot of repetition but this is done intentionally. The whole point of those types is to define how the state changes between each step so all the changing data have to be explicitly typed. However, the `activeWalletName` property maintains the same type. In recent changes, I managed to reduce the repetition of this property.

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
szymonmaslowski: > refactor this not to use a bunch of ifs

Please take a look at [the current version](https://github.com/input-output-hk/lace/pull/1191/commits/31411f28b0a754f43a1d024b4f1c1bf7c4e5a4d4). I managed to express it with an object, so there are no ifs checking the state. I managed to preserve the type safety. I hope you like it.

> create a general handler for next and back

I am not convinced to this idea. Not every state will have both commands available, so we would need to check for the current state, which inverts the logic. I think we could define such global handlers only if they are possible in all the states.

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationFlow.tsx
greatertomi: Yeah. That works.

greatertomi (approved):
Looks good to me. Amazing stuff. 🚀

emiride (approved):
As there are few minor issues currently:
- Inability to test with variable number of co-signers
- Steps not being highlighted on the side

However, we will merge this and fix it with https://input-output.atlassian.net/browse/LW-10536

PR #1203, opened 2024-06-04
test(extension): add test coverage - LW-10647

wklos-iohk:
# Checklist

- [X] JIRA - LW-10647
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


packages/e2e-tests/src/assert/WalletAccountsAssert.ts
```diff
@@ -0,0 +1,40 @@
+import WalletAccounts from '../elements/accounts/WalletAccounts';
+import { expect } from 'chai';
+import { t } from '../utils/translationService';
+import WalletAccountItem from '../elements/accounts/WalletAccountItem';
+
+class WalletAccountsAssert {
```
ljagiela, line 6: maybe  you could rename to WalletAccountsMenuAssert ?

packages/e2e-tests/src/assert/WalletAccountsAssert.ts
```diff
@@ -0,0 +1,40 @@
...
+    for (let accountIndex = 0; accountIndex < 24; accountIndex++) {
+      const accountItem = new WalletAccountItem(accountIndex + 1);
+      await accountItem.container.scrollIntoView();
+      await accountItem.container.waitForDisplayed();
+      await accountItem.icon.waitForDisplayed();
+      expect(await accountItem.icon.getText()).to.equal(`${accountIndex}`);
+      await accountItem.label.waitForDisplayed();
+      expect(await accountItem.label.getText()).to.equal(`Account #${accountIndex}`);
+      await accountItem.path.waitForDisplayed();
+      expect(await accountItem.path.getText()).to.equal(`m/1852'/1815'/${accountIndex}'`);
```
ljagiela, line 35: is it possible to check also enable button ?

packages/e2e-tests/src/elements/accounts/WalletAccountItem.ts
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable no-undef */
+import { ChainablePromiseElement } from 'webdriverio';
+
+class WalletAccountItem {
```
ljagiela, line 4: WalletAccountMenuItem ?

packages/e2e-tests/src/elements/accounts/WalletAccounts.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable no-undef */
+import { ChainablePromiseElement } from 'webdriverio';
+import { ChainablePromiseArray } from 'webdriverio/build/types';
+
+class WalletAccounts {
```
ljagiela, line 5: WalletAccountsMenu ? I'm just worried that without menu after some time it might be a bit confusing when someone jumps into this

ljagiela (changes requested):

packages/e2e-tests/src/assert/WalletAccountsAssert.ts
wklos-iohk: I'd rather handle this in separate step/assertion as not all account items have such button. In general there would be 3 cases: 
1. no button at all for account in use,
2. `Enable` button for disabled accounts,
3. `Disable` button for accounts that are enabled but not in use.

PR #1213, opened 2024-06-07
feat: share wallet details screen

shawnbusuttil:
# Checklist

- [ ] JIRA - \<LW-10550>
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
```diff
@@ -73,7 +74,15 @@ export type StateQuorum = MakeState<{
...
 }>;
 
-type State = StateSetup | StateCoSigners | StateCoSignersImportantInfo | StateQuorum;
+export type StateShareDetails = MakeState<
+  StateCoSignersCommon & {
+    coSignersKeys: string[];
+    step: SharedWalletCreationStep.ShareDetails;
+    walletName: string;
+  }
+>;
```
szymonmaslowski, line 83: The `StateCoSignersCommon` type is related to the CoSigners step. I would rather avoid reusing it here. It is there because we have two CoSigngers steps (the second one displays modal) and both are identical except for the step property. Also, there is no `coSignersKeys` property.

However, if you strongly want to extend some other type I would suggest extending the quorum state without the step property. This might make sense because the state doesn't change between the quorum and the share details step.
```suggestion
export type StateShareDetails = MakeState<
  Omit<StateQuorum, 'step'> & {
    step: SharedWalletCreationStep.ShareDetails;
  }
>;
```

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
```diff
@@ -231,9 +246,10 @@ const makeStateMachine = ({ navigateHome }: { navigateHome: () => void }): State
       return {
         ...prevState,
         coSigners: undefined,
+        coSignersKeys: undefined,
         coSignersErrors: undefined,
         quorumRules: undefined,
-        step: SharedWalletCreationStep.Setup
+        step: SharedWalletCreationStep.ShareDetails
```
szymonmaslowski, line 252: We will need all the data for constructing the json.
```
        ...prevState,
        step: SharedWalletCreationStep.ShareDetails
```

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
```diff
@@ -124,7 +134,13 @@ const makeInitialState = (activeWalletName: string): State => ({
 
 const getInitialCoSignerValue = (): CoSigner => ({ id: uuid(), keys: '', name: '' });
 
-const makeStateMachine = ({ navigateHome }: { navigateHome: () => void }): StateMachine => ({
+const makeStateMachine = ({
+  navigateHome,
+  navigateToStart
+}: {
+  navigateHome: () => void;
```
szymonmaslowski, line 141: IMO Home might be understood differently. If you agree consider renaming to something more explicit.
```suggestion
  navigateToAssetsPage: () => void;
```

szymonmaslowski (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
shawnbusuttil: If we decide some day that home !== assets, then we would have to change this. :)

apps/browser-extension-wallet/src/views/browser-view/features/shared-wallet/create-flow/SharedWalletCreationStore.tsx
shawnbusuttil: If you remove it, you get an error under the Quorum key asking for coSignersKeys.

szymonmaslowski (approved):

PR #1250, opened 2024-06-28
test(extension): add tests for LW-9334 and LW-9335

wklos-iohk:
# Checklist

- [X] JIRA - LW-10817
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots

### DIFF
--- a/packages/e2e-tests/src/elements/addNewWallet/MainModal.ts
+++ b/packages/e2e-tests/src/elements/addNewWallet/MainModal.ts
@@ -1,0 +1,17 @@
+/* eslint-disable no-undef */
+import { ChainablePromiseElement } from 'webdriverio';
+import { OnboardingMainPage } from '../onboarding/mainPage';
+
+class MainModal extends OnboardingMainPage {
+  private MODAL = '.ant-modal-content';
+  private CLOSE_BUTTON = '[data-testid="navigation-button-cross"]';
+
+  get modal(): ChainablePromiseElement<WebdriverIO.Element> {
+    return $(this.MODAL);
+  }
+  get closeButton(): ChainablePromiseElement<WebdriverIO.Element> {
+    return $(this.CLOSE_BUTTON);
+  }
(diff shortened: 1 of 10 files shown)


packages/e2e-tests/src/elements/addNewWallet/MainModal.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable no-undef */
+import { ChainablePromiseElement } from 'webdriverio';
+import { OnboardingMainPage } from '../onboarding/mainPage';
+
+class MainModal extends OnboardingMainPage {
```
ljagiela, line 5: Maybe you could rename it to AddNewWalletMainModal or smth like this. When this PO is called later it does not say much what kind of MainModal it is.

packages/e2e-tests/src/elements/addNewWallet/MainModal.ts
```diff
@@ -0,0 +1,17 @@
...
+  private MODAL = '.ant-modal-content';
+  private CLOSE_BUTTON = '[data-testid="navigation-button-cross"]';
+
+  get modal(): ChainablePromiseElement<WebdriverIO.Element> {
```
ljagiela, line 9: maybe rename to container ? then when called it looks a bit better than MainModal.modal :)

ljagiela (changes requested):

packages/e2e-tests/src/elements/addNewWallet/MainModal.ts
wklos-iohk: updated

packages/e2e-tests/src/elements/addNewWallet/MainModal.ts
wklos-iohk: updated

oldGreg5 (approved):

PR #1326, opened 2024-07-26
test(extension): add option to run e2es on staging for split

ljagiela:


pczeglik-iohk:
# Allure Report



**processReports**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/smoke/linux/chrome/6399/10112957563/index.html) for [9d40082d](https://github.com/input-output-hk/lace/pull/1326/commits/9d40082d32df2b94ce8e1c07f284abc1a0745702)

.github/shared/build/action.yml
```diff
@@ -31,6 +31,9 @@ inputs:
     description: 'Banxa`s integration URL'
     required: false
     default: 'https://lacewallet.banxa-sandbox.com/'
+  CARDANO_SERVICES_URL_PREPROD:
+    description: 'Url for preprod env'
+    required: false
```
pczeglik-iohk, line 36: what about default?

.github/workflows/ci.yml
```diff
@@ -162,10 +162,10 @@ jobs:
           BATCH: ${{ matrix.batch }}
           SMOKE_ONLY: true
 
-  smokeTests:
-    name: Smoke tests
+  processReports:
+    name: Process test reports
```
pczeglik-iohk, line 166: small details but maybe it is worth adding `Process smoke test reports` as we have also unit tests, and we do nothing here with unit

.github/workflows/e2e-tests-linux-split.yml
```diff
@@ -35,6 +35,14 @@ on:
         description: 'Run only smoke tests'
         required: true
         default: false
+      preprod_env:
+        type: choice
+        required: true
+        default: dev
+        description: 'preprod env to use'
```
pczeglik-iohk, line 42: what will happen if I select `network to use` = `mainnet` and then will change here `staging`?

.github/workflows/e2e-tests-linux-split.yml
```diff
@@ -45,6 +53,7 @@ env:
   DISPLAY: ':99.0'
   NODE_OPTIONS: --max-old-space-size=16384
   BRANCH: ${{ github.ref_name }}
+  PREPROD_ENV: ${{ github.event.inputs.preprod_env || 'dev' }}
```
pczeglik-iohk, line 56: `preprod_env` has `dev` as default. Do we need to use `||  'dev`?

pczeglik-iohk: maybe we can remove `chrome` and `preprod` from lines 50 and 51 as well, as `chrome` is default for `inputs.browser` and maybe we can hoist `preprod` to be default for `inputs.network`?

.github/shared/build/action.yml
ljagiela: I don't want to set default, because there is a risk that it will override eg. release builds. If it's not set, then the URL from env.defaults should be used.

.github/workflows/e2e-tests-linux-split.yml
ljagiela: nothing :) urls for mainnet are not modified anywhere, we don't even have mainnet on staging. 
New param says clearly that it's `preprod env to use`

.github/workflows/e2e-tests-linux-split.yml
ljagiela: fixed & removed

pczeglik-iohk (approved):
Looks good

PR #1507, opened 2024-10-31
feat: providers are now consumed from the service worker instead of the UI script

AngelCastilloB:
Provider objects (or at least an instance of RateLimiter) must be shared between SW and UI.
# Checklist

- [x] JIRA - \<[link](https://input-output.atlassian.net/browse/LW-11761)>
- [ ] Screenshots added.

---

## Proposed solution

UI script now consumes Provider objects instead of creating their own providers


pczeglik-iohk:
# Allure Report



**processReports**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/smoke/linux/chrome/7272/11606114687/index.html) for [aa76a7e7](https://github.com/input-output-hk/lace/pull/1507/commits/aa76a7e7d70838a272e04d08f3bf30f4d2cce5e5)
| Total | 23     | 9      | 4       | 0     | 36    | ❌     |

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
```diff
@@ -232,6 +232,28 @@ walletManager
...
+          switchMap((activeWallet) => {
+            if (!activeWallet) {
+              return;
+            }
+
+            const chainId = activeWallet.props.chainId;
+            const chainName: Wallet.ChainName = chainIdToChainName(chainId);
+
+            // eslint-disable-next-line consistent-return
+            return from(getProviders(chainName));
```
mkazlauskas, line 249: I think this creates new provider objects rather than reusing the ones already used by the wallet. An easy fix (a little bit hacky though :sweat_smile:) would be to memoize getProviders function

mkazlauskas (changes requested):

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
AngelCastilloB: It does, but internally it uses the same RateLmiter object which was the point no?

AngelCastilloB: done here https://github.com/input-output-hk/lace/pull/1507/commits/8d9d021d58664e3bf9fa8b79be3c6c49c015e159

mkazlauskas: Good point, so it was already working correctly :) Thanks @AngelCastilloB

mkazlauskas (approved):
Nice work! :rocket:

PR #1522, opened 2024-11-07
test: remove legacy way of injecting wallets

ljagiela:
* Refactoring Wallet Injection in Lace: Previously, wallet injection in Lace relied on outdated methods of data storage, using browser storage. This PR replaces these methods with a new approach that utilises the wallet repository for wallet storage.
* Disabling Service Worker Logs for Nightly E2E Tests: Service worker logs were causing instability in the nightly E2E tests. This PR disables these logs by default for nightly runs, they can now be enabled on demand as needed.


pczeglik-iohk:
# Allure Report



**processReports**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/smoke/linux/chrome/7341/11728015355/index.html) for [8f8fc459](https://github.com/input-output-hk/lace/pull/1522/commits/8f8fc45967c40ead36120eeb20232070b2419acd)
| Total | 29     | 3      | 4       | 0     | 36    | ❌     |

packages/e2e-tests/src/fixture/localStorageInitializer.ts
```diff
@@ -97,14 +56,33 @@ class LocalStorageInitializer {
...
-  ): Promise<void> => {
+  async initializeAppSettings(): Promise<void> {
+    const network = extensionUtils.getNetwork().name;
+    const appSettings = `{"chainName":"${network}"}`;
+    await localStorageManager.setItem('appSettings', appSettings);
+  }
+
+  initialiseBasicLocalStorageData = async (walletName = 'TestAutomationWallet'): Promise<void> => {
+    // Pause fix for flaky tests where local storage keys are disappearing when executed right after opening the extension
+    await browser.pause(500);
```
mkazlauskas, line 67: I know this wasn't a change in this PR so not a change request, but want to discuss the use of `browser.pause`: imho it should only be used as last resort, as it indicates a race condition, which can be flaky even with the pause. Also it can make tests slower than necessary. Would this be possible to implement here, for example?
```suggestion
    // Pause fix for flaky tests where local storage keys are disappearing when executed right after opening the extension
    await localStorageReady();
```

mkazlauskas (approved):
Looks good! It's an odd commit scope though, why not just `test:`?
image

wklos-iohk (approved):

packages/e2e-tests/src/fixture/localStorageInitializer.ts
ljagiela: I totally agree about browser.pause and we generally stick to this rule when writing tests. There are some rare exceptions and I believe this is one of them :) I think that this particular one could be removed. I will do that in a separate PR.

PR #1584, opened 2024-12-09
fix: collateral password issue [LW-11953]

greatertomi:
# Checklist

- [x] JIRA - LW-11953
- [ ] Screenshots added.

---

## Proposed solution

Adjusted the flow to submit transaction before clearing password

### DIFF
--- a/apps/browser-extension-wallet/src/views/browser-view/features/settings/components/Collateral/send/CollateralFooterSend.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/settings/components/Collateral/send/CollateralFooterSend.tsx
@@ -44,8 +44,6 @@ export const CollateralFooterSend = ({
   const submitTx = async () => withSignTxConfirmation(submitCollateralTx, secretsUtil.password.value);
 
   const handleClick = async () => {
-    onClaim();
-
     if (!hasEnoughAda) {
       return onClose();
     }
@@ -53,6 +51,7 @@ export const CollateralFooterSend = ({
       if (popupView && !isInMemory)
         return await backgroundServices?.handleOpenBrowser({ section: BrowserViewSections.COLLATERAL_SETTINGS });
       await submitTx();
+      onClaim();
       toast.notify({ text: t('browserView.settings.wallet.collateral.toast.add') });
       if (isInMemory) onClose();
     } catch {
--- a/packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
+++ b/packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
@@ -44,6 +44,11 @@ export const EnterPassword: VFC<EnterPasswordProps> = ({
   const icon = mapOfWalletTypeIconProperties[kind];
 
   const next = () => {
+    if (!password.value) {
+      console.error('Password is undefined');
+      return;
+    }
+
     onGenerateKeys(password.value);
     clearSecrets();
   };
--- a/packages/core/src/ui/components/WalletSetup/WalletSetupPasswordStep.tsx
+++ b/packages/core/src/ui/components/WalletSetup/WalletSetupPasswordStep.tsx
@@ -60,10 +60,9 @@ export const WalletSetupPasswordStep = ({
   const passwordConfirmationErrorMessage =
     passHasBeenValidated && password !== passwordConfirmation ? translations.noMatchPassword : '';
   const isNextEnabled = Boolean(
-    passHasBeenValidated &&
-      !passwordConfirmationErrorMessage &&
-      score >= minimumPassLevelRequired &&
-      password.value.length > 0
+    passHasBeenValidated && !passwordConfirmationErrorMessage && score >= minimumPassLevelRequired && password?.value
+      ? password.value.length > 0
+      : false
   );
 
   const complexityBarList: BarStates = useMemo(() => getComplexityBarStateList(score), [score]);
@@ -79,7 +78,7 @@ export const WalletSetupPasswordStep = ({
       description={translations.description}
       onBack={onBack}
       onNext={() => {
-        onNext({ password: password.value });
+        password.value ? onNext({ password: password.value }) : console.error('Password is undefined');
       }}
       isNextEnabled={isNextEnabled}
       currentTimelineStep={WalletTimelineSteps.WALLET_SETUP}


pczeglik-iohk:
# Allure Report



**processReports**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/smoke/linux/chrome/7616/12244034293/index.html) for [28a04491](https://github.com/input-output-hk/lace/pull/1584/commits/28a0449159a289c491708c2f2e446756f9b07fcf)
| Total | 32     | 0      | 4       | 0     | 36    | ✅     |

apps/browser-extension-wallet/src/views/browser-view/features/settings/components/Collateral/send/CollateralFooterSend.tsx
```diff
@@ -44,15 +44,14 @@ export const CollateralFooterSend = ({
...
-    onClaim();
-
     if (!hasEnoughAda) {
       return onClose();
     }
     try {
       if (popupView && !isInMemory)
         return await backgroundServices?.handleOpenBrowser({ section: BrowserViewSections.COLLATERAL_SETTINGS });
       await submitTx();
+      onClaim();
```
pczeglik-iohk, line 54: I'm leaving the reference to `onClaim` [definition](https://github.com/input-output-hk/lace/blob/fix/lw-11953-collateral-signing/apps/browser-extension-wallet/src/views/browser-view/features/settings/components/Collateral/CollateralDrawer.tsx#L136) for the next reviewer, so it's slightly easier to grasp how this change fixes the issue.

packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
```diff
@@ -44,7 +44,7 @@ export const EnterPassword: VFC<EnterPasswordProps> = ({
   const icon = mapOfWalletTypeIconProperties[kind];
 
   const next = () => {
-    onGenerateKeys(password.value);
+    onGenerateKeys(password?.value || '');
```
pczeglik-iohk, line 47: This may help with broken build, but are we sure that empty string is acceptable value for key generation?

This is a [place where the password is used](https://github.com/input-output-hk/lace/blob/fix/lw-11953-collateral-signing/packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/Store/machine.ts#L54) and after traversing x levels of components / abstractions etc. I discovered this [place](https://github.com/input-output-hk/lace/blob/fix/lw-11953-collateral-signing/packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/generate-shared-wallet-key.ts#L19) where password is converted to passphrase buffer and used to `getSharedWalletExtendedPublicKey`

To me it looks like empty string is not a good fit.
@szymonmaslowski would you be able to review this change, please? I see you have worked a lot on that area.

packages/core/src/ui/components/WalletSetup/WalletSetupPasswordStep.tsx
```diff
@@ -60,10 +60,7 @@ export const WalletSetupPasswordStep = ({
   const passwordConfirmationErrorMessage =
     passHasBeenValidated && password !== passwordConfirmation ? translations.noMatchPassword : '';
   const isNextEnabled = Boolean(
-    passHasBeenValidated &&
-      !passwordConfirmationErrorMessage &&
-      score >= minimumPassLevelRequired &&
-      password.value.length > 0
```
pczeglik-iohk, line 66: To me this validation check was to make sure the password is not blank string

pczeglik-iohk: `!!password.value` does similar check, but it allows value to be undefined, which we do not allow with types

```
const password = {};
console.log(!!password.value);
```

apps/browser-extension-wallet/src/views/browser-view/features/settings/components/Collateral/send/CollateralFooterSend.tsx
greatertomi: What we do currently is that secrets are cleared before the collateral is signed or submitted, so we try to sign collateral with an empty password. Moving this after `submitTx` allows for the transaction to be submitted before the password is cleared.

packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
greatertomi: Maybe, but considering that it only uses an empty string when the password is empty, I don't see any issue with this. But alternatively, we could handle it like this

```
if (password.value) {
      onGenerateKeys(password.value);
      clearSecrets();
    } else {
      console.error('Password is undefined');
    }
```

packages/core/src/ui/components/WalletSetup/WalletSetupPasswordStep.tsx
greatertomi: How about making the check like this? `password?.value ? password.value.length > 0 : false`

packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
pczeglik-iohk: Yes, I think we should check password has value and throw an error in case is invalid. But I'm not too familiar with this part of the codebase. Let's see what @szymonmaslowski think about this :)

Am I right that the logic should be something like this. 

1. Use password for key generation (password is required and cannot be blank)
2. If password is empty no need to generate key.

And, If password is empty, and we still try to generate key, we have some flaw in the logic? 

By passing empty string as a password do we introduce a security flaw?

greatertomi: Yes, I agree @pczeglik-iohk. We should ensure that the password is defined before generating key. 

My first approach was based on the assumption that an empty string or invalid password would throw an error during key generation.

greatertomi: Fixed on 2dce5bf

packages/core/src/ui/components/WalletSetup/WalletSetupPasswordStep.tsx
greatertomi: Fixed on 2dce5bf

packages/core/src/ui/components/WalletSetup/WalletSetupPasswordStep.tsx
```diff
@@ -79,7 +76,7 @@ export const WalletSetupPasswordStep = ({
       description={translations.description}
       onBack={onBack}
       onNext={() => {
-        onNext({ password: password.value });
+        onNext({ password: password.value || '' });
```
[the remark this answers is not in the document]
greatertomi: Fixed on 2dce5bf

packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
szymonmaslowski: looks good!

packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
```diff
@@ -44,6 +44,11 @@ export const EnterPassword: VFC<EnterPasswordProps> = ({
   const icon = mapOfWalletTypeIconProperties[kind];
 
   const next = () => {
+    if (!password.value) {
+      console.error('Password is undefined');
```
pczeglik-iohk, line 48: console.error is commonly used in the lace codebase, so I think it's fine. 

But in the same time I wonder if this is the way how we should handle it. Especially as we have Sentry in place.

Maybe we do not need to pollute users console with application error, but report them to Sentry, so in this case `throw new Error('EnterPassword: password is undefined')` maybe a bit better.

It can be as it is, no need to change it right now, but it's worth exploring this area in upcoming weeks, so we can refactor.

pczeglik-iohk (approved):
Thank you @greatertomi Code looks good. 

There is only one decision to be made for the future we can ignore right now.

packages/core/src/shared-wallets/add-shared-wallet/generate-key-flow/EnterPassword/EnterPassword.tsx
greatertomi: Should I create a ticket to explore that?

szymonmaslowski: We need to be careful with throwing errors in react code, as it might unmount the whole app, as we don't have any error boundary as far as I know. The disappearing application would be a far more severe issue than the benefit of catching the problem with sentry.

pczeglik-iohk: there are methods like captureException and captureMessage in Sentry https://docs.sentry.io/platforms/javascript/guides/react/usage/
we may be interested to use. We could wrap it into some abstraction layer so we are provider agnostic etc

PR #1639, opened 2025-01-10
Feat/lw 12074 enable new input resolver in lace

AngelCastilloB:
# Checklist

- [x] JIRA - \<link>
- [ ] Screenshots added.

---

Enable new Input resolver on Lace

## Proposed solution

Add a new feature flag to toggle blockfrost input resolver on Lace


apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
```diff
@@ -130,14 +129,36 @@ const walletFactory: WalletFactory<Wallet.WalletMetadata, Wallet.AccountMetadata
...
 
+    const featureFlags = await getFeatureFlags(chainId.networkMagic);
+    const useBlockfrostInputResolver = isExperimentEnabled(featureFlags, ExperimentName.BLOCKFROST_INPUT_RESOLVER);
+
+    let inputResolver;
+
+    if (useBlockfrostInputResolver) {
+      const { BLOCKFROST_CONFIGS } = config();
+
+      const blockfrostClient = new BlockfrostClient(
```
mkazlauskas, line 140: We must have only a single instance of BlockfrostClient per application. Maybe creating input resolver should be moved to `getProviders`

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
```diff
@@ -152,6 +173,16 @@ const walletFactory: WalletFactory<Wallet.WalletMetadata, Wallet.AccountMetadata
           witnesser
         }
       );
+
+      if (inputResolver) {
+        inputResolver.setContext(sharedWallet);
```
mkazlauskas, line 178: (just an idea, not a change req) I think we had some kind of 'combineInputResolvers` function. It could be used instead of context approach

mkazlauskas (changes requested):

apps/browser-extension-wallet/src/lib/scripts/background/wallet.ts
AngelCastilloB: done here https://github.com/input-output-hk/lace/pull/1639/commits/c5eef3080cb403c5173d330618e57e090f453e27

mkazlauskas (approved):
Great work! 🚀

PR #1658, opened 2025-01-22
feat: add caching for dapp radar api requests

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-12143)
- [ ] Screenshots added.

---

## Proposed solution

- storing DApp Radar API responses in a local storage entry
- each entry has a timestamp of insertion
- if the timestamp is older than one day the entry is considered expired
- if no cache entry then DApp Radar API request is performed

## Testing

- Ensure requests are performed only once
- Modify timestamp of particular cache entry (the one with `dapp-explorer-data-cache` name) to make it older than 1 day and ensure the request is made.

### DIFF
--- a/apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.test.ts
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.test.ts
@@ -1,0 +1,108 @@
+import { CacheRepo, localStorageItemName, makeCacheRequest, oneDayOfCacheValidity } from './cache';
+
+const currentDate = new Date();
+jest.useFakeTimers().setSystemTime(currentDate);
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const prepareCacheRequestFn = (params: { getLocalStorageItem?: (key: string) => any } = {}) => {
+  const getLocalStorageItem = params.getLocalStorageItem ?? jest.fn();
+  const setLocalStorageItem = jest.fn();
+
+  return {
+    cacheRequest: makeCacheRequest({ getLocalStorageItem, setLocalStorageItem }),
+    setLocalStorageItem
+  };
+};
+
+const cacheKey = 'cacheKey';
+
+describe('DApp Explorer data cache', () => {
+  it('returns cached response', async () => {
+    const expectedValue = 'cached value';
+    const { cacheRequest } = prepareCacheRequestFn({
+      getLocalStorageItem: jest
+        .fn()
+        .mockReturnValue(JSON.stringify({ [cacheKey]: { data: expectedValue, timestamp: currentDate.getTime() } }))
+    });
+
+    expect(await cacheRequest(cacheKey, () => void 0)).toEqual(expectedValue);
+  });
+
+  it('calls fetchData function if there is no particular data in cache', async () => {
+    const { cacheRequest } = prepareCacheRequestFn();
+    const fetchDataMock = jest.fn();
+    await cacheRequest(cacheKey, fetchDataMock);
+
+    expect(fetchDataMock).toHaveBeenCalledTimes(1);
+  });
+
+  it('calls fetchData function if cached data expired', async () => {
+    const { cacheRequest } = prepareCacheRequestFn({
+      getLocalStorageItem: jest.fn().mockReturnValue(
+        JSON.stringify({
+          [cacheKey]: { data: 'data', timestamp: currentDate.getTime() - oneDayOfCacheValidity }
+        })
+      )
+    });
+    const fetchDataMock = jest.fn();
+    await cacheRequest(cacheKey, fetchDataMock);
+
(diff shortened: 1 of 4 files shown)


apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.ts
```diff
@@ -0,0 +1,52 @@
...
+  data: Data;
+  timestamp: number;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type CacheRepo = Record<string, CacheEntry<any>>;
+
+const localStorageItemName = 'dapp-explorer-data-cache';
+// eslint-disable-next-line no-magic-numbers
+const oneDayOfCacheValidity = 1000 * 60 * 60 * 1000;
```
rhyslbw, line 11:
```suggestion
const oneDayOfCacheValidity = 1000 * 60 * 60 * 24;
```

rhyslbw (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.ts
szymonmaslowski: [Done.](https://github.com/input-output-hk/lace/pull/1658/files#diff-f8393097bea29dbb335435e47e64a8e535cb25dad77d84a62fbb502c78f52486R11)

rhyslbw (approved):
Great work @szymonmaslowski 🎉

pczeglik-iohk:
# Allure Report



**processReports**: ✅ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/smoke/linux/chrome/8011/12909611639/index.html) for [7c2980b0](https://github.com/input-output-hk/lace/pull/1658/commits/7c2980b0b903a963ece48a83d5f9514d6bf2fd9e)
| Total | 34     | 0      | 3       | 0     | 37    | ✅     |

apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    } catch (error) {
+      console.error('Failed to parse dapp explorer data cache', error);
+    }
+
+    const cacheEntry: CacheEntry<Response> | undefined = cacheRepo[cacheKey];
+    if (isCacheStillValid(cacheEntry)) return cacheEntry.data;
+
+    const response = await fetchData();
+    const newCacheEntry: CacheEntry<Response> = {
```
greatertomi, line 38: This cache stores records regardless of whether the request succeeds or fails. To test this, clear the cache, remove the API key, and make a network request that results in an error. Then, restore the API key and make the request again. The cache will save the empty array from the first failed request and will not replace it with the data from the second successful request.

apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it('calls fetchData function if there is no particular data in cache', async () => {
+    const { cacheRequest } = prepareCacheRequestFn();
+    const fetchDataMock = jest.fn();
+    await cacheRequest(cacheKey, fetchDataMock);
+
+    expect(fetchDataMock).toHaveBeenCalledTimes(1);
+  });
+
+  it('calls fetchData function if cached data expired', async () => {
```
greatertomi, line 39: Maybe we should add this test case. "calls fetchData function if cached data is an empty array"

greatertomi (changes requested):

apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.ts
szymonmaslowski: Good catch! I think the fetch data should be expected to throw and in that case the response should not be cached. I will change that in a second.

szymonmaslowski: I changed the usages of `cacheRequest` fn. The function itself should cache successfull request. In case of an exception it should be transparent for the `cacheRequest` so the exception boubles up and should be catched by the logic which calls the `cacheRequest`.

apps/browser-extension-wallet/src/views/browser-view/features/dapp/explorer/services/cache.test.ts
szymonmaslowski: I didn't add any test case regarding that considering [this comment](https://github.com/input-output-hk/lace/pull/1658#discussion_r1925752091).

greatertomi (approved):
Looks good. 🚀

PR #1656, opened 2025-01-20
feat(extension): load historical tx on demand when scrolling [LW-12099]

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-12099)

---

## Proposed solution
- update SDK packages
- utilise infinite scroll lazy loading

## Testing

- activities list in the activity tab (popup and browser view), please consider next scenarios
  - no initial txs are available
  - pending/signed tx should always appear at the top of the list
  - pending tx should change it status to "confirmed"
  - particular amount of initial txs are available with no more items to load
  - particular amount of initial txs are available with more items to load
  - initially available tx should be displayed almost immediately with no additional requests to blockfrost
- activities list in the reward history section in the staking page.

PR #1697, opened 2025-02-06
feat: integrate persistent cache with blockchain providers

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-12228)
- [ ] Screenshots added.

---

## Proposed solution

Enabling persistent cache for:
 - chain history provider
 - input resolver
 - utxo provider

## Testing

Initial loading of lace is expected do make following blockfrost requests:
 - `txs/${hash}/cbor`
 - `txs/${txId}`
 - `txs/${txIn}/utxos`

When particular cache size reaches the max size defined here, then 10% of items with oldest access time should be removed.


packages/cardano/src/wallet/lib/providers.ts
```diff
@@ -94,6 +95,32 @@ interface ProvidersConfig {
...
+  [CacheName.inputResolver]: {
+    count: 170,
+    // eslint-disable-next-line no-magic-numbers
+    size: 0.06 * sizeOf1mb
+  },
+  [CacheName.utxoProvider]: {
+    count: 198,
+    // eslint-disable-next-line no-magic-numbers
+    size: 0.3 * sizeOf1mb
+  }
```
szymonmaslowski, line 121: Values have been calculated with first populating data to cache running lace and then running this script:

<details>
  <summary>script</summary>

      const getPropsForCollection = async (totalMaxSize) => {
      const chainHistoryProviderMetadataKey = 'chain-history-provider-cache-metadata';
      const inputReslverMetadataKey = 'input-resolver-cache-metadata';
      const utxoProviderMetadataKey = 'utxo-provider-cache-metadata';
  
      const getSize = async (metadataKey) => {
        const { [metadataKey]: metadata } = await chrome.storage.local.get(metadataKey);
        const bytes = await chrome.storage.local.getBytesInUse([...Object.keys(metadata), metadataKey]);
        const itemsCount = Object.keys(metadata).length;
        const sizeInMb = bytes / 1024 / 1024;
        const sizeOfSingleItem = sizeInMb / itemsCount;
  
        return {
          sizeInMb,
          sizeOfSingleItem
        };
      };
  
      const chainHistorySize = await getSize(chainHistoryProviderMetadataKey);
      const inputResolverSize = await getSize(inputReslverMetadataKey);
      const utxoSize = await getSize(utxoProviderMetadataKey);
  
      const totalSize = chainHistorySize.sizeInMb + inputResolverSize.sizeInMb + utxoSize.sizeInMb;
  
      const targetSizeOfChainHistory = totalMaxSize * (chainHistorySize.sizeInMb / totalSize);
      const targetSizeOfInputResolver = totalMaxSize * (inputResolverSize.sizeInMb / totalSize);
      const targetSizeOfUtxo = totalMaxSize * (utxoSize.sizeInMb / totalSize);
  
      const targetCountOfChainHistory = targetSizeOfChainHistory / chainHistorySize.sizeOfSingleItem;
      const targetCountOfInputResolver = targetSizeOfInputResolver / inputResolverSize.sizeOfSingleItem;
      const targetCountOfUtxo = targetSizeOfUtxo / utxoSize.sizeOfSingleItem;
  
      return {
        chainHistory: {
          count: targetCountOfChainHistory,
          size: targetSizeOfChainHistory
        },
        inputResolver: {
          count: targetCountOfInputResolver,
          size: targetSizeOfInputResolver
        },
        utxo: {
          count: targetCountOfUtxo,
          size: targetSizeOfUtxo
        }
      };
    };

</details>

pczeglik-iohk:
# Allure Report



**processReports**: ❌ [test report](https://lace-qa:8PP5wBaDV6UoXj@dq4ajm5i5q7bz.cloudfront.net/smoke/linux/chrome/8241/13183460185/index.html) for [b26a82ca](https://github.com/input-output-hk/lace/pull/1697/commits/b26a82caa4edc973a2d4e3ae51f2bb41dea37898)
| Total | 31     | 2      | 4       | 0     | 37    | ❌     |

packages/cardano/src/wallet/lib/providers.ts
vetalcore: @szymonmaslowski would it make sense to check limit (smallest and biggest) values per collection, right now u assume (`sizeInMb / itemsCount;`) that all of them are equal?

vetalcore: would be great if you could test it on https://github.com/input-output-hk/lace/pull/1656 :)

vetalcore (changes requested):

vetalcore:
Great Work! Left one comment/suggestion*

packages/cardano/src/wallet/lib/providers.ts
```diff
@@ -94,6 +95,32 @@ interface ProvidersConfig {
...
+    // eslint-disable-next-line no-magic-numbers
+    size: 6.64 * sizeOf1mb
+  },
+  [CacheName.inputResolver]: {
+    count: 170,
+    // eslint-disable-next-line no-magic-numbers
+    size: 0.06 * sizeOf1mb
+  },
+  [CacheName.utxoProvider]: {
+    count: 198,
```
vetalcore, line 118: would it make sense to check for the items with min/max size in each collection, right now you do the average?
would be great to check on https://github.com/input-output-hk/lace/pull/1656 :)

vetalcore (changes requested):

packages/cardano/src/wallet/lib/providers.ts
szymonmaslowski: Yes, right now I calculate an average size of item (collection size / items count) and then check how many such items would fit in the target collection size. What is your suggestion?

szymonmaslowski: For the context, the count represents a max count of items for a collection, and it is a fallback - primarily, cache logic will attempt to read the size in bytes, but there is a small chance this API won't be available.

przemyslaw-wlodek (approved):

packages/cardano/src/wallet/lib/providers.ts
DominikGuzei: I guess it would be better if the SDK did a rough calculation based on the stored data itself instead of taking the optional count param 🤔 something like this should do the trick well enough:
```js
new TextEncoder().encode(
  Object.entries(await browser.storage.sync.get())
    .map(([key, value]) => key + JSON.stringify(value))
    .join('')
).length
```

DominikGuzei (approved):
Code LGTM 🚢  the only potential improvement i can see is that approximation of `count` being used (due to SDK limitations), which might be better solved by the SDK doing some rough "bytes in use" calculation itself. But that's just a side note :)

packages/cardano/src/wallet/lib/providers.ts
vetalcore: just thinking if there could be a case that size of an item is actually way bigger than we've initially calculated, so we end up having to small amount of items in the cache, and would be constantly just replacing items there. Feel free to ignore this comment though :)

vetalcore (approved):

packages/cardano/src/wallet/lib/providers.ts
szymonmaslowski: The context here is that default storage space is 10MB, but it can be increased to unlimited but requires requesting additional permission, which will most likely increase review time of new submitted extension version. The idea was to go with 10MB this release and increase for the next release.

@DominikGuzei, that's a nice suggestion! I would definitely go for it, but there is a requirement that we don't want to load the whole cache to memory. Having unlimited storage we will probably aim for more size per cache collection.

@vetalcore Considering having the unlimited storage size we will also increase the limits per collection so the problem you described will be even less likely to happen.

PR #1728, opened 2025-02-19
feat: improve windows managemet when switching to popup [LW 12347]

szymonmaslowski:
# Checklist

- [x] JIRA - [\<link>](https://input-output.atlassian.net/browse/LW-12347)
- [ ] Screenshots added.

---

## Proposed solution

Improve logic behind  management of browser windows during switching to popup mode.

1. Get all windows
2. From all windows, get those that have some 3rd party tabs (`windowsWith3rdPartyTabs`) - they may also have any lace tabs. The preference is to open the popup in a window which does not close after we close all the lace tabs.
3. If `windowsWith3rdPartyTabs` contains the currently focussed one, then it is put in at index 0. (`candidateWindowsWithPreferenceForCurrentlyFocused`) - We prefer to open the popup in a window which is currently foccused. If the list does not contain a currently focussed window, it is fine; we will choose another one. The focussed one will close anyway after we close all lace tabs as it does not have any other tab.
4. We select a window that will be focussed next (`nextFocusedWindow`) - we take the first one from `candidateWindowsWithPreferenceForCurrentlyFocused` (which may be the currently focussed one)
5. If `nextFocusedWindow` is empty, then it means there are no windows with 3rd party tabs open. In this case, we want to choose any of the windows to be focused next and open an empty tab on that window so the window does not close when all its lace tabs get closed.
6. We close all the lace tabs in all windows, no matter how many windows there are. Windows having only lace tabs will close. Those having 3rd party tabs remain open.
7. We open the popup in the window that was nominated to be the next active (a currently active one or any other).

## Testing

1. Single window with only lace tab - should close lace, open empty tab and open lace popup
2. Single window with lace and some other tab - should close lace, switch to other tab and open popup
3. Two windows, active one with only lace and background one with some other tab - should close active window and open popup in background window
3. Two windows, active one with lace and some other tab, background one with lace and some other tab - should close all lace tabs in all windows, switch to other tabs in all windows and open popup in active window

### DIFF
--- a/apps/browser-extension-wallet/src/features/dapp/components/SignData.tsx
+++ b/apps/browser-extension-wallet/src/features/dapp/components/SignData.tsx
@@ -81,7 +81,7 @@ export const SignData = (): React.ReactElement => {
     <Layout title={undefined}>
       <div className={styles.passwordContainer}>
         <Spin spinning={isLoading}>
-          <h5 className={styles.message}>
+          <h5 className={styles.message} data-testid="sign-transaction-description">
             {t('browserView.transaction.send.enterWalletPasswordToConfirmTransaction')}
           </h5>
           <Password
--- a/apps/browser-extension-wallet/src/release-notes/1_20_0.tsx
+++ b/apps/browser-extension-wallet/src/release-notes/1_20_0.tsx
@@ -1,0 +1,25 @@
+/* eslint-disable camelcase */
+import React from 'react';
+
+const ReleaseNote_1_20_0 = (): React.ReactElement => (
+  <>
+    <p>This release:</p>
+    <ul style={{ listStyleType: 'disc', margin: 0 }}>
+      <li>Rename wallets & accounts for better organization</li>
+      <li>Governance tab as a starting point for an integrated experience</li>
+      <li>Restored full activity feed to view all past transactions & rewards</li>
+      <li>
+        Improved input focusing so your password can be immediately typed when signing a DApp transaction or message
+      </li>
+      <li>Optimized caching for reduced network demand and increased application performance</li>
+      <li>Refined landing page to help guide users switching to Nami mode</li>
+      <li>Opt-in Debug Logging – control diagnostics on your terms</li>
+    </ul>
+    <p>
+      We’re continuously improving Lace with every update — this release includes numerous bug fixes and performance
+      improvements based on your feedback!
+    </p>
+  </>
+);
+
+export default ReleaseNote_1_20_0;
--- a/apps/browser-extension-wallet/src/views/browser-view/features/assets/components/AssetsPortfolio/AssetPortfolioContent.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/assets/components/AssetsPortfolio/AssetPortfolioContent.tsx
@@ -68,7 +68,7 @@ export const AssetPortfolioContent = ({
 
   return (
     <>
-      {assetList?.length > MIN_ASSETS_COUNT_FOR_SEARCH && (
+      {totalAssets > MIN_ASSETS_COUNT_FOR_SEARCH && (
         <div className={styles.searchBoxContainer}>
           <SearchBox
             placeholder={t('browserView.assets.searchPlaceholder')}
--- a/apps/browser-extension-wallet/src/views/browser-view/features/settings/components/SettingsPreferences/SettingsPreferences.tsx
+++ b/apps/browser-extension-wallet/src/views/browser-view/features/settings/components/SettingsPreferences/SettingsPreferences.tsx
@@ -81,7 +81,7 @@ export const SettingsPreferences = ({ popupView = false }: SettingsPreferencesPr
         sendCurrencyChangeEvent={handleSendCurrencyChangeEvent}
       />
       <SettingsCard>
-        <Title level={5} className={styles.heading5} data-testid="wallet-settings-heading">
+        <Title level={5} className={styles.heading5} data-testid="wallet-preferences-heading">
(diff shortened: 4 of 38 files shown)


apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts
```diff
@@ -109,18 +110,38 @@ const handleOpenNamiBrowser = async (data: OpenNamiBrowserData) => {
   await tabs.create({ url: `popup.html#${data.path}` }).catch((error) => logger.error(error));
 };
 
+const isItLaceOrNamiTab = (tab: Tabs.Tab) => ['Lace', POPUP_WINDOW_NAMI_TITLE].includes(tab.title);
+const doesWindowHaveOtherTabs = (browserWindow: Windows.Window) =>
+  browserWindow.tabs.some((t) => !isItLaceOrNamiTab(t));
```
przemyslaw-wlodek, line 115: This won't capture Nami tab AFAIK, because popup doesn't have own `windowId`.

apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts
```diff
@@ -109,18 +110,38 @@ const handleOpenNamiBrowser = async (data: OpenNamiBrowserData) => {
...
+    const allWindows = await windows.getAll();
+    if (allWindows.length === 0) return;
+
+    const allWindowsWithTheirTabs = await ensureTabsAvailable(allWindows);
+    const windowsWithTabsOtherThanLaceOrNami = allWindowsWithTheirTabs.filter((w) => doesWindowHaveOtherTabs(w));
+    const windowsOrderedWithActiveFirst = windowsWithTabsOtherThanLaceOrNami.sort(
+      (w1, w2) => Number(w2.focused) - Number(w1.focused)
+    );
+
+    let windowToBeUsed = windowsOrderedWithActiveFirst[0];
```
przemyslaw-wlodek, line 135: Wouldn't it be wrong if we have an active window with only the Lace tab and then `windowsOrderedWithActiveFirst` doesn't contain the active one? The rest of the logic fails to apply because there will be `windowToBeUsed` but it's not active.

apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts
```diff
@@ -109,18 +110,38 @@ const handleOpenNamiBrowser = async (data: OpenNamiBrowserData) => {
...
+    const allWindowsHaveOnlyLaceOrNamiTabs = !windowToBeUsed;
+    if (allWindowsHaveOnlyLaceOrNamiTabs) {
+      windowToBeUsed = allWindowsWithTheirTabs[0];
+      await tabs.create({ active: true, windowId: windowToBeUsed.id });
+    }
+
+    await windows.update(windowToBeUsed.id, { focused: true });
```
przemyslaw-wlodek, line 142: What if we focus the Lace tab, and then we close all Lace tabs? I think it will move focus to the other tab, but is this what we wanted?

apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts
```diff
@@ -109,18 +110,38 @@ const handleOpenNamiBrowser = async (data: OpenNamiBrowserData) => {
   await tabs.create({ url: `popup.html#${data.path}` }).catch((error) => logger.error(error));
 };
 
+const isItLaceOrNamiTab = (tab: Tabs.Tab) => ['Lace', POPUP_WINDOW_NAMI_TITLE].includes(tab.title);
```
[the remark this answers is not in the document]
szymonmaslowski: [Done.](https://github.com/input-output-hk/lace/compare/b29e2755af97c07be5ee5f2fe0af6e56042f9143..277b499757fd159ff05f9e5f06f1e523fbd3b207#diff-b3dd434d5790ec525db5c2204306bd0da3b753e909071d4aceed6814359a7578R122)

apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts
```diff
@@ -109,18 +110,38 @@ const handleOpenNamiBrowser = async (data: OpenNamiBrowserData) => {
...
+const doesWindowHaveOtherTabs = (browserWindow: Windows.Window) =>
+  browserWindow.tabs.some((t) => !isItLaceOrNamiTab(t));
+const ensureTabsAvailable = (browserWindows: Windows.Window[]) => {
```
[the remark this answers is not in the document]
szymonmaslowski: [changed to `enrichWithTabsDataIfMissing`](https://github.com/input-output-hk/lace/compare/b29e2755af97c07be5ee5f2fe0af6e56042f9143..277b499757fd159ff05f9e5f06f1e523fbd3b207#diff-b3dd434d5790ec525db5c2204306bd0da3b753e909071d4aceed6814359a7578R113)

apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts
szymonmaslowski: We intend to focus a window, not a tab. We want to move the focus to any non-lace tab in any open window, but preferably the currently focussed one if it has any other tabs.

I explained the algorithm [here](https://github.com/input-output-hk/lace/pull/1728#issue-2863273026).

przemyslaw-wlodek (approved):

przemyslaw-wlodek:
Great work @szymonmaslowski! 🚀 🫶

PR #1736, opened 2025-02-24
fix(extension): race condition in activity list [LW-12226]

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-12226)

---

## Proposed solution
This PR removes possible race conditions from Activity List mappers.

## Testing
See [LW-12226](https://input-output.atlassian.net/browse/LW-12226)

[LW-12226]: https://input-output.atlassian.net/browse/LW-12226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser-extension-wallet/src/hooks/useWalletActivities.ts
+++ b/apps/browser-extension-wallet/src/hooks/useWalletActivities.ts
@@ -7,6 +7,7 @@ import { mapWalletActivities } from '@src/stores/slices';
 import { Wallet } from '@lace/cardano';
 import { AssetActivityListProps, useItemsPageSize } from '@lace/core';
 import { useTxHistoryLoader } from './useTxHistoryLoader';
+import { useAsyncSwitchMap } from '@hooks/useAsyncSwitchMap';
 
 type UseWalletActivitiesProps = {
   sendAnalytics: () => void;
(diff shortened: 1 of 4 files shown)


apps/browser-extension-wallet/src/hooks/useWalletActivities.ts
```diff
@@ -133,20 +135,32 @@ export const useWalletActivitiesPaginated = ({
     [fetchActivitiesDeps, fetchActivitiesProps, walletState]
   );
 
+  const handleUpdateWalletActivities = useCallback(
```
vetalcore, line 138: @przemyslaw-wlodek any chance you could add simple test covering this scenario? eg: race conditions between tx is removed from pending vs tx appears in history?

apps/browser-extension-wallet/src/hooks/useWalletActivities.ts
```diff
@@ -63,6 +63,8 @@ export type WalletActivitiesPaginated = Pick<WalletActivities, 'walletActivities
   loadedTxLength?: number;
 };
 
+let mapActivitiesCallCount = 0;
```
mkazlauskas, line 66: This is global, while the exported hook can be used in multiple components. I suggest to `useState` within the hook instead of this global var

przemyslaw-wlodek: > the exported hook can be used in multiple components
You're right! Thanks! I wanted to prevent races globally, but this is pointless for now (and may work incorrectly with my approach). Local state can suit our needs, but I would recommend `useRef` though.

apps/browser-extension-wallet/src/hooks/useWalletActivities.ts
przemyslaw-wlodek: Thanks, I will add one! This module has no tests at all 💀

mkazlauskas (approved):
Great work! 🚀

PR #1770, opened 2025-03-08
chore: [lw-12077] upgrade deps reported by modus audit

vetalcore:
upgrade deps reported by modus audit to their safe versions

# Checklist

- [x] JIRA - [LW-12077](https://input-output.atlassian.net/browse/LW-12077)
- [ ] Screenshots added.

---

## Proposed solution

Upgraded via fixing version in `resolutions` in package.json:
- `postcss` to `8.4.31` or later
- `webpack` to `5.94.0` or later
- `babel (@babel/core)` to `7.23.2` or later
- `braces` to `3.0.3` or later
- `browserify-sign` to `4.2.2` or later
- `elliptic` to `6.6.0` or later

## Testing

## Screenshots

[LW-12077]: https://input-output.atlassian.net/browse/LW-12077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/browser-extension-wallet/webpack.common.js
```diff
@@ -75,7 +75,8 @@ module.exports = () => {
         dns: false,
         tls: false,
         process: false,
-        child_process: false
+        child_process: false,
+        vm: require.resolve('vm-browserify')
```
mchappell, line 79: can you drop an inline comment explaining why this is required please

vetalcore: since webpack version 5 has been "fixed" via `resolutions`  in package.json, we would get:
```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "vm": require.resolve("vm-browserify") }'
        - install 'vm-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "vm": false }
```
when trying to build an app

vetalcore: i mean that it would throw if we remove it, so no need to add a comment, wdyt?

mchappell (approved):

PR #1809, opened 2025-04-01
[LW-12572] Optimize BitcoinWallet requests to backend

AngelCastilloB:
# Checklist

- [x] JIRA - \<[link](https://input-output.atlassian.net/browse/LW-12572)>
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


apps/browser-extension-wallet/src/views/bitcoin-mode/features/activity/components/ActivityLayout.tsx
```diff
@@ -17,6 +16,19 @@ import Book from '@assets/icons/book.svg';
 import LightBulb from '@assets/icons/light.svg';
 import Video from '@assets/icons/video.svg';
 import { useCurrencyStore } from '@providers';
+import { useWalletStore } from '@stores';
+import uniqBy from 'lodash/uniqBy';
+import debounce from 'lodash/debounce';
+import { LACE_APP_ID } from '@utils/constants';
+import { Skeleton } from 'antd';
+
+const updateTransactions = (
```
iccicci, line 25: This function seems to be a code repetition of an equal function in `apps/browser-extension-wallet/src/views/bitcoin-mode/features/activity/components/Activity.tsx`; does make sense to refactor in order to have a reusable function defined only once and used twice?

packages/bitcoin/src/wallet/lib/providers/MaestroBitcoinDataProvider.ts
```diff
@@ -246,25 +288,29 @@ export class MaestroBitcoinDataProvider implements BlockchainDataProvider {
...
 
-  /**
-   * Fetches details of a specific transaction by its hash.
-   */
-  private async getRpcTransactionDetails(txHash: string): Promise<any> {
     const response = await this.api.get(`/rpc/transaction/${txHash}?verbose=true`);
+
+    // Only cache transactions that are confirmed, we need to keep fetching
+    // unconfirmed transaction to get their updated block time and height when they are confirmed.
+    if (response.data.data.confirmations > 0) {
```
iccicci, line 310: Here I have two questions:
1. Could it be a good idea to permanently cache a tx when it has more than only one confirmation?
2. Regardless from the answer to previous question, in case the check is `false`, could it be a good idea to cache the result **but** with a short lifetime in order to avoid subsequent call in a timeframe where the result is most likely unchanged?

packages/bitcoin/src/wallet/lib/tx-builder/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+    tx.ins.map(async (input) => {
+      const txId = Buffer.from(input.hash).reverse().toString('hex');
+      const index = input.index;
+      try {
+        const resolvedInput = await inputResolver.resolve(txId, index);
+        return {
+          ...resolvedInput,
+          txId,
+          index
+        };
```
iccicci, line 43:
```suggestion
        return resolvedInput;
```

Aren't `txId` and `index` already present in `resolvedInput`?

apps/browser-extension-wallet/src/views/bitcoin-mode/features/activity/components/ActivityLayout.tsx
AngelCastilloB: done here https://github.com/input-output-hk/lace/pull/1809/commits/6e68df7521323a3a448a06d91d519c8afe292a26 @iccicci

packages/bitcoin/src/wallet/lib/tx-builder/utils.ts
AngelCastilloB: done here https://github.com/input-output-hk/lace/pull/1809/commits/6e68df7521323a3a448a06d91d519c8afe292a26

packages/bitcoin/src/wallet/lib/providers/MaestroBitcoinDataProvider.ts
AngelCastilloB: @iccicci we are persisting the transactions on disk, so these are somewhat permanent, but the cache has limited size and has some eviction policies, this is because before we were caching all transaction of the wallets, and big wallets had hard time reloading all transaction from disk every time. In reality users mostly look at recent transactions, so the strategy we chose here was the same as Cardano, cache persistently these transactions, but the cache has a limited size.

As for two, we dont cache these transactions, because there is some info missing (block height and timestamp since these are pending TXs), the polling is every minute and this only applies for pending transactions so I think is ok to not cache those (no big impact performance). If we cache them we would display them on the activity with timestamp 0 until we refresh the cache, which would miss-place

iccicci: Clear explanation! Thank you!

iccicci (approved):
Excellent job @AngelCastilloB !

P.S.
Due to some permission issue I can't resolve the conversation on the PR; could you please resolve them for me?
Thank you

rhyslbw (approved):

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
```diff
@@ -391,6 +392,31 @@ export const useWalletManager = (): UseWalletManager => {
...
+      combineLatest([walletManager.activeWalletId$, bitcoinWalletManager.activeWalletId$])
+    );
+
+    const walletId = activeBlockchain === 'bitcoin' ? bitcoinActiveWallet?.walletId : activeWallet?.walletId;
+    const wallets = await firstValueFrom(walletRepository.wallets$);
+    const wallet = wallets.find((w) => w.walletId === walletId);
+
+    let accountMetadata;
+    if (wallet && isAnyBip32Wallet(wallet)) {
+      const account = wallet.accounts.find((a) => a.accountIndex === wallet.metadata.lastActiveAccountIndex);
```
vetalcore, line 408: since wallet metadata is not updated in the `addAccount` flow https://github.com/input-output-hk/lace/blob/main/apps/browser-extension-wallet/src/hooks/useWalletManager.ts#L1366, `lastActiveAccountIndex` won't be accurate here

```suggestion
      const account = wallet.accounts.find((a) => a.accountIndex === cardanoWallet?.source.account?.accountIndex);
```
or please align the implementation of `addAccount` with `activateWallet` so it also updates the wallet metadata

vetalcore (changes requested):

vetalcore (approved):

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
vetalcore: resolved

PR #1827, opened 2025-04-17
feat: [lw-12632] fix spinner position on network switch for FF

vetalcore:
# Checklist

- [x] JIRA - [LW-12632](https://input-output.atlassian.net/browse/LW-12632)
- [ ] Screenshots added.

---

## Proposed solution

height specified in percentages is calculated differently in FF (should be set on parent explicitly otherwise would default to auto)

## Testing

build extension for FF, then switch the network in lace.

## Screenshots


wklos-iohk (approved):
QA OK

PR #1862, opened 2025-05-09
test: add cpu & memory monitoring for e2e tests

ljagiela:


packages/e2e-tests/src/hooks/scenarioTagRunner.ts
```diff
@@ -5,15 +5,25 @@ import consoleManager from '../utils/consoleManager';
...
+
+const monitor = PidMonitor.getInstance();
 
 // eslint-disable-next-line no-unused-vars
 Before(async () => {
   if (String(process.env.SERVICE_WORKER_LOGS) === 'true') {
     await consoleManager.startLogsCollection();
   }
+  const pidMonitorInitialized = await monitor.init();
+  if (!pidMonitorInitialized) return;
```
pczeglik-iohk, line 18: is this if needed? what will happen if pid monitor is not initialized?

packages/e2e-tests/src/hooks/scenarioTagRunner.ts
```diff
@@ -5,15 +5,25 @@ import consoleManager from '../utils/consoleManager';
...
+  monitor.start();
```
pczeglik-iohk, line 19: Maybe it's better to try / catch start()? as it may throw exception when pid is undefined, what means initialization failed?

pczeglik-iohk (approved):

pczeglik-iohk (approved):

PR #1884, opened 2025-05-28
fix: (lw-12780) do not trigger create shared wallet more than once

vetalcore:
# Checklist

- [ ] JIRA - [LW-12780](https://input-output.atlassian.net/browse/LW-12780)
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots


przemyslaw-wlodek (approved):
Approved after an offline discussion. In general this workaround works and was tested in other place in the app. Technically the major issue is with the reducers that have side-effects (they shouldn't!).

PR #1898, opened 2025-06-06
Fix/client side rate limiter for coingecko

mkazlauskas:
# Checklist

- [x] JIRA - \<[LW-13032](https://input-output.atlassian.net/browse/LW-13032)>
- [ ] Screenshots added.

---

## Proposed solution

See commit messages

## Testing

See JIRA ticket

[LW-13032]: https://input-output.atlassian.net/browse/LW-13032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser-extension-wallet/src/lib/scripts/background/services/cardanoTokenPrices.ts
+++ b/apps/browser-extension-wallet/src/lib/scripts/background/services/cardanoTokenPrices.ts
@@ -1,25 +1,33 @@
+/* eslint-disable no-magic-numbers */
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 import {
   map,
   from,
   combineLatest,
   Observable,
-  Subject,
   startWith,
-  groupBy,
   mergeMap,
-  throttleTime,
-  exhaustMap,
   scan,
   auditTime,
-  tap
+  switchMap,
+  EMPTY,
+  repeatWhen,
+  timer,
+  takeWhile,
+  shareReplay,
+  ReplaySubject,
+  take,
+  of,
+  filter
 } from 'rxjs';
 import { storage } from 'webextension-polyfill';
 import { TokenPrices, Status, MaybeTokenPrice } from '../../types';
 import { Cardano, Milliseconds } from '@cardano-sdk/core';
-import { TrackerSubject } from '@cardano-sdk/util-rxjs';
 import { config } from '@src/config';
 import Bottleneck from 'bottleneck';
+import { ActiveWallet } from '@cardano-sdk/web-extension';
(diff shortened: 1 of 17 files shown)


apps/browser-extension-wallet/src/views/browser-view/features/send-transaction/helpers.ts
```diff
@@ -256,7 +256,7 @@ export const formatRow = ({
     if (asset) {
       const ticker = asset.nftMetadata?.name ?? asset.tokenMetadata?.ticker ?? asset.tokenMetadata?.name;
       const amount = Wallet.util.calculateAssetBalance(balance, asset);
-      const tokenPriceInAda = prices?.cardano.getTokenPrice(id)?.priceInAda;
+      const tokenPriceInAda = mayBeNFT(asset) ? undefined : prices?.cardano.getTokenPrice(id)?.priceInAda;
```
mirceahasegan, line 259: Can include the `mayBeNFT` check in `getTokenPrice` to avoid calling it in multiple places?

mkazlauskas: Yes it would be a good improvement. I did it this way only because I wanted to do minimal functional changes in UI - adding the check in getTokenPrice would add a new dependency hook (to get asset info) in `useFetchCoinPrice`, which would be used as a dependency in `useMemo` and could potentially result in more re-renders/fetches. Just playing it safe.

apps/browser-extension-wallet/src/lib/scripts/background/services/cardanoTokenPrices.ts
```diff
@@ -1,119 +1,224 @@
...
+const CACHE_KEY = 'cardano-token-prices';
+const { TOKEN_PRICE_CHECK_INTERVAL } = config();
+// eslint-disable-next-line no-magic-numbers, new-cap
+const TOKEN_PRICE_EMIT_THROTTLE_FREQUENCY = Milliseconds(1000);
+
+const rateLimiterSettings: Bottleneck.ConstructorOptions = {
+  reservoir: 5,
+  reservoirIncreaseAmount: 2,
+  reservoirIncreaseInterval: 1000,
+  reservoirIncreaseMaximum: 5
```
DominikGuzei, line 61: @mkazlauskas please add the reasoning behind using these reservoir settings 🙏

DominikGuzei (approved):
Code looks great 👍 haven't tested it though … and new code is missing unit tests - but we can add those after the emergency is resolved

PR #1902, opened 2025-06-10
fix(extension): loading wallet during app init + switching [LW-13016]

przemyslaw-wlodek:
# Checklist

- [x] [JIRA](https://input-output.atlassian.net/browse/LW-13016)

---

## Proposed solution
- fix `loadWallet` implementation
- ensure we are using proper `activeWalletProps` depending on blockchain
- remove dead code
- **bonus**: add toast when wallet is switched after removal

## Testing
Please run full regression tests on switching wallets, especially bitcoin wallets and ensure that [LW-12781](https://input-output.atlassian.net/browse/LW-12781) is still working fine.

[LW-12781]: https://input-output.atlassian.net/browse/LW-12781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/browser-extension-wallet/src/hooks/useAppInit.ts
+++ b/apps/browser-extension-wallet/src/hooks/useAppInit.ts
@@ -1,12 +1,14 @@
 import { useObservable } from '@lace/common';
 import { useWalletStore } from '@stores';
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 import { isKeyHashAddress } from '@cardano-sdk/wallet';
 import { AddressesDiscoveryStatus } from '@lib/communication/addresses-discoverer';
 import { useWalletManager } from './useWalletManager';
 import { useWalletState } from './useWalletState';
 import { setBackgroundStorage } from '@lib/scripts/background/storage';
 import { useCustomSubmitApi } from '@hooks/useCustomSubmitApi';
+import { bitcoinWalletManager } from '@lib/wallet-api-ui';
+import { useCurrentBlockchain } from '@src/multichain';
 
 export const useAppInit = (): void => {
   const {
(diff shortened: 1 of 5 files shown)


vetalcore (approved):

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
```diff
@@ -688,11 +688,12 @@ export const useWalletManager = (): UseWalletManager => {
    * @returns resolves with wallet information or null when no wallet is found
    */
   const loadWallet = useCallback(
-    // eslint-disable-next-line complexity
+    // eslint-disable-next-line complexity, max-statements
```
Firefly0, line 691: hope you don't mind since I m an outsider

but:
this function seems to accumulate to much responsability - it seems that it handles a lot of things, like: wallet migration, active wallet identification, btc wallet activation, cardano wallet setup (seems like it violates SRP principle)

could  you consider splitting this up in multiple smaller functions? - would be much easier to test, read, and maintain in the future

apps/browser-extension-wallet/src/hooks/useAppInit.ts
```diff
@@ -56,7 +59,16 @@ export const useAppInit = (): void => {
   }, [environmentName, getCustomSubmitApiForNetwork]);
 
   const wallets = useObservable(walletRepository.wallets$);
-  const activeWalletProps = useObservable(walletManager.activeWalletId$);
+  const activeCardanoWalletProps = useObservable(walletManager.activeWalletId$);
+  const activeBitcoinWalletProps = useObservable(bitcoinWalletManager.activeWalletId$);
+  const activeWalletProps = useMemo(
+    () =>
+      blockchain === 'bitcoin'
```
Firefly0, line 66: would not be better to use a variable for this string? 
I see it used in multiple places arround the application. There is even a function called 'isbitcoin'
my advice: 
create an external function and import it where need it - less room for misspeling in the future, and easier to maintain

apps/browser-extension-wallet/src/hooks/useWalletManager.ts
Firefly0: Ahh
I understand 
Thank you for the feedback. I m looking forward for the v2 :D 

Till then, may you suggest me some other repo from your org where I can contribute
I m a fan of ADA and I want to help, and learn more about this ecosystem on my spare time - and I do not want to stand in the way either(FE, TS related projects) 
Thank you

PR #1940, opened 2025-07-04
Remove key from manifest for Edge Web Store publish

pczeglik-iohk:
For Edge Web Store submission we need to remove the `key` property from the manifest.json file.

### DIFF
--- a/.github/actions/build/app/action.yml
+++ b/.github/actions/build/app/action.yml
@@ -112,7 +112,9 @@ runs:
 
     - name: Prepare MS Edge
       if: ${{ inputs.BROWSER_TARGET == 'chromium' }}
-      run: 'sed -i "s/\"key\": \"[^\"]*\",//" manifest.json'
+      run: |
+        jq 'del(.key)' manifest.json | jq -c . > manifest.tmp.json
+        mv manifest.tmp.json manifest.json
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/build/app/action.yml
```diff
@@ -112,7 +112,9 @@ runs:
 
     - name: Prepare MS Edge
       if: ${{ inputs.BROWSER_TARGET == 'chromium' }}
-      run: 'sed -i "s/\"key\": \"[^\"]*\",//" manifest.json'
```
pczeglik-iohk, line 115: When I tested this it worked, because I tested it agains well formatted json file. In CI we deal with compressed to one line manifest.json file (webpack). Because of that `sed` command did not work.

`jq` should be available in Github's ubuntu runner by default, and it's better suited to this task.

.github/actions/build/app/action.yml
```diff
@@ -112,7 +112,9 @@ runs:
...
+      run: |
+        jq 'del(.key)' manifest.json | jq -c . > manifest.tmp.json
```
pczeglik-iohk, line 116: `jq` restores json formatting. That's why we need to use `-c .` to compress it back

mchappell (approved):

PR #1957, opened 2025-07-22
LW-13053 Add estimated ROS

iccicci:
# Checklist

- [x] JIRA - LW-13053
- [ ] Screenshots added.

---

## Proposed solution

Since Lace shows the ROS only when showing stake pools details, enriching the `enrichStakePool` function with the ROS estimation is enough to compute the estimated ROS when needed.

The ROS computation is based on Shelley ledger specifications (with Conway amendment); the main difference is that ledger specifications uses data from the snapshot on epoch rollovers while here live data is used.

The computation requires some data from the network, the genesis parameter and protocol parameters, so the `StakePoolService` now fetches this data from a `NetworkInfoProvider` or BF directly.

About the stake pool data, almost all data are already fetched; the only exception is the distinction between member stake and operator stake and BF doesn't offer an API to get it... strictly required to compute the `memberReward`. 😞

The first idea was to fetch the controlled stack for each pool owner, but the high number of required HTTP requests discouraged me...

By chances I decided to postpone this problem and to go on... and... in next steps, I found a mathematical trick to compute the ROS without computing the `memberReward`! 💪 🕺 
Enjoy reading the code for details. 😉 

## Testing

Manually tested.

### DIFF
--- a/packages/cardano/src/wallet/lib/stakePoolService.ts
+++ b/packages/cardano/src/wallet/lib/stakePoolService.ts
@@ -5,6 +5,7 @@
 import { BlockfrostClient } from '@cardano-sdk/cardano-services-client';
 import {
   Cardano,
+  NetworkInfoProvider,
   Paginated,
   QueryStakePoolsArgs,
   StakePoolProvider,
@@ -18,6 +19,7 @@ import Fuse from 'fuse.js';
 
 const BF_API_PAGE_SIZE = 100;
 export const CACHE_KEY = 'stake-pool-service-data';
+const SECONDS_PER_YEAR = 31_536_000; // 365 * 24 * 60 * 60
 const ONE_DAY = 86_400_000; // One day in milliseconds
 
 // The empty text placeholders used to make the stake pools with empty names or tickers to be sorted at the end of the list
@@ -64,6 +66,11 @@ interface BlockFrostPool {
 }
 
 interface StakePoolCachedData {
+  networkData: {
+    genesisParameters: Cardano.CompactGenesis;
+    network: Responses['network'];
+    protocolParameters: Cardano.ProtocolParameters;
+  };
   lastFetchTime: number;
   poolDetails: Map<Cardano.PoolId, Responses['pool']>;
   stakePools: Cardano.StakePool[];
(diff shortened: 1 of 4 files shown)


rhyslbw (commented):
_Ignore_

packages/cardano/src/wallet/lib/stakePoolService.ts
```diff
@@ -107,13 +115,92 @@ const filterByIdentifier = (identifier: IdentifierType) => (pool: Cardano.StakeP
...
+  // the reward from member perspective as if all the stake was controlled by the members.
+  // This is why the multiplication by member proportional stake is not needed.
+  // It is mandatory to keep in mind this in next steps.
+
+  const rewards = maxPool * mkApparentPerformance;
+  const c = Number(stakePool.cost);
+  const m = Cardano.FractionUtils.toNumber(stakePool.margin);
+
+  // If the rewards are less than the stakePool cost, the ROS is 0
+  if (rewards <= c) return 0;
```
mkazlauskas, line 178: I think this is incorrect for small pools that are minting <1 block per epoch, because in epochs where they do mint a block, the reward is pretty big - it's a gamble with such pools - either 0 or big reward each epoch. Since ROS is a yearly figure, expected rewards for small pools is probably still >0. Not sure how to best handle this - something to think about

packages/cardano/src/wallet/lib/stakePoolService.ts
```diff
@@ -107,13 +115,92 @@ const filterByIdentifier = (identifier: IdentifierType) => (pool: Cardano.StakeP
...
+  // Ref: https://intersectmbo.github.io/formal-ledger-specifications/cardano-ledger.pdf - Fig. 64
+
+  const blocksInEpoch = genesisParameters.epochLength * genesisParameters.activeSlotsCoefficient;
+  const poolBlocks = (blocksInEpoch * liveStake) / (totalStake * (saturation > 1 ? saturation : 1));
+
+  const mkApparentPerformance = poolBlocks / (blocksInEpoch * s);
+
+  // Simplified Fig. 47.2
+  // BF does not offer a way to distinguish between member stake and operator stake.
+  // By luck, there is no need to distinguish between member rewards and operator rewards; it is enough to compute
```
mkazlauskas, line 168: Just wondering, isn't operator stake the same as 'live pledge'?

packages/cardano/src/wallet/lib/stakePoolService.ts
```diff
@@ -107,13 +115,92 @@ const filterByIdentifier = (identifier: IdentifierType) => (pool: Cardano.StakeP
...
+interface enrichStakePoolParams {
+  details: Responses['pool'];
+  networkData: StakePoolCachedData['networkData'];
+  id: Cardano.PoolId;
+  stakePools: Cardano.StakePool[];
+}
+
+type StakePoolWithMetrics = Cardano.StakePool & { metrics: Cardano.StakePoolMetrics };
+
+const estimateROS = (stakePool: StakePoolWithMetrics, networkData: StakePoolCachedData['networkData']) => {
```
mkazlauskas, line 127: Some unit tests would be great for this

mkazlauskas (changes requested):

packages/cardano/src/wallet/lib/stakePoolService.ts
```diff
@@ -18,6 +19,8 @@ import Fuse from 'fuse.js';
 
 const BF_API_PAGE_SIZE = 100;
 export const CACHE_KEY = 'stake-pool-service-data';
+// eslint-disable-next-line no-magic-numbers
+const EPOCHS_IN_YEAR = 365 / 5; // 73 epochs in a year
```
mirceahasegan, line 23: This is valid only for mainnet genesis slotLength and epochLength: https://github.com/input-output-hk/cardano-configurations/blob/master/network/mainnet/genesis/shelley.json#L62-L65.
Maybe use genesis parameters from NetworkInfoProvider to calculate this.

packages/cardano/src/wallet/lib/stakePoolService.ts
iccicci: Figure 48 clearly mention them as two distinct values: I'd say no.

packages/cardano/src/wallet/lib/stakePoolService.ts
iccicci: e66a6354e should solve this

packages/cardano/src/wallet/lib/stakePoolService.ts
iccicci: Opened LW-13316 to handle this separately.

packages/cardano/src/wallet/lib/stakePoolService.ts
iccicci: fb107c71e should solve this

mkazlauskas: Could you please add some comment/documentation about how test data was produced?

mkazlauskas (approved):
Great work! 🚀

PR #1961, opened 2025-07-30
feat: render metadata field in transaction preview before signing

AngelCastilloB:
# Checklist

- [x] JIRA - \<[link](https://input-output.atlassian.net/browse/LW-13336)>
- [x] Screenshots added.

---

## Proposed solution

This PR adds transaction metadata rendering

## Screenshots

Screenshot from 2025-07-30 13-41-06


packages/core/src/ui/components/ActivityDetail/components/TxDetailsAuxiliaryData/TxDetailsAuxiliaryData.module.scss
```diff
@@ -0,0 +1,19 @@
...
+@import '../../../../../../../common/src/ui/styles/abstracts/typography';
+
+.auxData {
+  word-break: break-all;
+  width: 100%;
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+}
+
+.copyButton {
```
rhyslbw, line 11: What's the reason for needing to style this button here? Seems like a generic component, no?

rhyslbw (approved):
Nice work @AngelCastilloB !

packages/core/src/ui/components/ActivityDetail/components/TxDetailsAuxiliaryData/TxDetailsAuxiliaryData.module.scss
AngelCastilloB: We could import this from another .css but it seems is not centralized, all the components in the tx preview are duplicating this style, I just copied the pattern, we could extract this to a common CSS tho (we would need to update all the ther CSS too and components to import this new .css)

PR #1964, opened 2025-08-05
LW-13316 Estimated ROS for small pools

iccicci:
# Checklist

- [x] JIRA - LW-13316
- [ ] Screenshots added.

---

## Proposed solution

Current implementation makes estimated ROS for stake pools with low stake just zero as their average member rewards is lower than stake pool cost.
This change computes the estimated ROS for the pool as if it mints a number of blocks at floor and at ceil of it estimated minted block per epoch, than computes the annualized estimated ROS combining these two values with their weight.

## Testing

The first commit add some more stake pools with low stake to the ROS test, at this stage all the stake pools with low stake have a zero estimated ROS.
The second commit introduces the computation change and adjusts the estimated ROS for all the stake pools in the test.
We can see the estimated ROS change for stake pools with high stake is negligible while for stake pools with low stake the change is from zero to an actual ROS.

### DIFF
--- a/packages/cardano/src/wallet/lib/__tests__/stakePoolService.ros.test.ts
+++ b/packages/cardano/src/wallet/lib/__tests__/stakePoolService.ros.test.ts
@@ -1404,6 +1404,169 @@ const cachedData = {
       rewardAccount: '',
       status: 'active',
       vrf: ''
+    },
+    {
+      cost: {
+        __type: 'bigint',
+        value: '340000000'
+      },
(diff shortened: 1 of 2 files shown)


iccicci:
> Great work! Appreciate the test-first approach 💪

Thank you 🙇 

> Interestingly it slightly changed the estimated ros for other pools, probably due to lost precision? 🤷

The change for other pools was expected: the whole formula is changed for them as well.
It is a negligible change because the impact of the issue addressed by this PR is negligible for stake pools with high stake.

PR #2062, opened 2025-11-20
Feat/lw 13593 notification service data layer

iccicci:
# Checklist

- [ ] JIRA - \<link>
- [ ] Screenshots added.

---

## Proposed solution

## Testing

## Screenshots

### DIFF
--- a/packages/notifications/src/providers/PubNub/PubNubProvider.ts
+++ b/packages/notifications/src/providers/PubNub/PubNubProvider.ts
@@ -1,4 +1,4 @@
-import { getCurrentTimetoken, isArrayOfStrings, unused } from '../../utils';
+import { getCurrentTimetoken, getNow, isArrayOfStrings, unused } from '../../utils';
 import { Notification, NotificationsLogger, NotificationsStorage, Topic } from '../../types';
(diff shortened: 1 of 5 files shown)


packages/notifications/src/providers/PubNub/PubNubProvider.ts
```diff
@@ -330,11 +393,37 @@ export class PubNubProvider implements NotificationsProvider {
...
+   * (stored in subscribedTopics). Uses the last sync timestamp to fetch only new messages.
+   *
+   * @returns Promise that resolves when all missed messages have been fetched and processed
+   */
+  private async fetchMissedMessages(): Promise<void> {
+    const subscribedTopics = (await this.storage.getItem<Topic['id'][]>(this.storageKeys.getSubscribedTopics())) || [];
+
+    for (const topic of this.topics)
+      if (topic.isSubscribed || subscribedTopics.includes(topic.id)) {
+        const end = (await this.storage.getItem<string>(this.storageKeys.getLastSync(topic.id))) || '0';
```
mirceahasegan: If my understanding is correct, this will fetch the whole history if it was never synced.
I think it should sync from the time of subscription until present.

PR #2068, opened 2025-11-24
LW-13919 Add NotificationsClient init retry strategy

iccicci:
# Checklist

- [x] JIRA - LW-13919
- [ ] Screenshots added.

---

## Proposed solution

- Retry failed init calls.
- In case of init retry, removed stored token as it could be the source of the error.

## Testing

Unit tests added

### DIFF
--- a/packages/notifications/src/NotificationsClient.ts
+++ b/packages/notifications/src/NotificationsClient.ts
@@ -13,6 +13,18 @@ import { PendingCommands } from './PendingCommands';
 import { isArrayOfStrings, unused } from './utils';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/notifications/src/NotificationsClient.ts
```diff
@@ -134,8 +140,8 @@ export class NotificationsClient {
       if (!error) this.pendingCommands.onConnectionRestored();
     });
 
-    this.init().catch((error) => {
-      logger.error('NotificationsClient: Failed to initialize notifications client', error);
+    this.initTillSuccess().catch((error) => {
```
mirceahasegan: i think this should be exponential back-off

PR #2102, opened 2025-12-06
Fix/nc mode env for e2e

mirceahasegan:
e2e tests still had a build path where they were using the production notification center


mirceahasegan:
closed because it is being merged to `main`.
We can cherry-pick it at a later time if we'll release one more patch from 1.32.x branch

PR #2104, opened 2025-12-09
fix(extension): use defer for fetch execution and improve price update [LW-14044]

mirceahasegan:
See change comments


rhyslbw (approved):
Nice work @mirceahasegan

wklos-iohk (approved):

PR #2114, opened 2025-12-15
Sync color scheme with v2 [LW-13998]

iccicci:
# Checklist

- [x] JIRA - LW-13998
- [ ] Screenshots added.

---

## Proposed solution

Added bundle API to sync color scheme with v2.
Added bundle API to make UI aware if this is a bundle build or not.

**Note:** this should be reviewed with https://github.com/input-output-hk/lace-platform/pull/1244

## Testing

Manually tested

### DIFF
--- a/v1/apps/browser-extension-wallet/src/hooks/useLMP.ts
+++ b/v1/apps/browser-extension-wallet/src/hooks/useLMP.ts
@@ -4,6 +4,7 @@ import { Wallet } from '@lace/cardano';
 import {
   APP_MODE,
   bundleAppApiProps,
+  ColorSchemeV1Sync,
   lmpApiBaseChannel,
   lmpModeStorage,
   onboardingParamsStorage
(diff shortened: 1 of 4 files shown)


DominikGuzei (changes requested):
@iccicci there is already an existing background service for changing and persisting the theme (which makes it much easier to sync between v1/2) like in my [language PR](https://github.com/input-output-hk/lace/pull/2100)

In Lace v1 this is used (as the primary layer) via the [ThemeSwitcher](https://github.com/input-output-hk/lace/blob/ce63ec9c82d342ab0628eea2210a5e8561ae79d7/v1/apps/browser-extension-wallet/src/components/MainMenu/DropdownMenuOverlay/components/ThemeSwitcher.tsx#L50), which uses the `handleChangeTheme` from [utilityServices](https://github.com/input-output-hk/lace/blob/ce63ec9c82d342ab0628eea2210a5e8561ae79d7/v1/apps/browser-extension-wallet/src/lib/scripts/background/services/utilityServices.ts#L360) (same as it was for language) which sets it in the background storage.

I'm not sure why Lace v1 uses local storage additionally (maybe an old implementation? 🤔 ) … anyway, using the background storage would greatly simplify your design as you don't need the `ColorSchemeV1Sync` properties etc. but only provide a simple [hand-over api](https://github.com/input-output-hk/lace/pull/2100/changes#diff-e0f0336c7cafe89ea97b70a3eb06cbcb9b94c66a006ff0fb0dea0fa301e1c469R29).

Lace v1 is always considered as the "primary" source (because that's where you start when opening the wallet), the `theme$` observable would propagate the current theme to Lace v2 (which needs to be picked up in [Lace v2 LMP bundle module](https://github.com/input-output-hk/lace-platform/pull/1221/changes) - since this code is contained within the `lmp-bundle` you also don't need to store any `isBundle` prop anywhere since it only exists if the the feature flag is enabled)

PR #2205, opened 2026-03-31
ci: lace (v1+v2+lmp) bundle workflow updates [LW-14589]

iadmytro:
Change:
1)  Name: Release Lace Bundle → Release Lace Bundle (v1+v2+lmp)
2) Trigger: removed push to main, added schedule
3) test_mode default: smoke
4) Slack notification only for nightly builds
5) Local build with `make build-prod-v1-v2-lmp`


mchappell (changes requested):
This works correctly, but i'm blocking this one until we've had chance to talk about closing it as we likely do not need this.

